PR #20, opened 2023-05-05
Release/0.0.72

timrombergjakobsson:
**Describe pull-request**  
Release of @scania/tegel@0.0.72


package-lock.json
theJohnnyMe, line 1: This one can be reverted as it is top level file that we do not release.

theJohnnyMe (approved):
LGTM!

PR #33, opened 2023-05-09
fix(button): made disabled button non-clickable

mJarsater:
**Describe pull-request**  
* Disabled button is no longer clickable. It will do nothing on click.
* Scss file cleanup in button

**Should we also expose a default slot that people can use instead of the text prop? We would need to do some check if that slot has any content in connectedCallback to be able to set onlyIcon to true it the slot is empty and there is no text prop.**

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1671

**How to test**  
1. Go to storybook link below
2. Check in Components -> Button
3. Check the log on a disabled button and make sure nothing is emitted.
4. Check the overall style of the button.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

### DIFF
--- a/core/src/components/button/button.scss
+++ b/core/src/components/button/button.scss
@@ -324,7 +324,6 @@ i.sdds-btn-icon[slot='icon'] {
   ::slotted([slot='icon']) {
     width: var(--sdds-spacing-element-20);
     height: var(--sdds-spacing-element-20);
-    pointer-events: none;
   }
 
   .sdds-btn-sm {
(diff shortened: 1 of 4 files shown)


theJohnnyMe (approved):
Nice clean-up and improvement! LGTM!

theJohnnyMe:
@mJarsater  

_

> Should we also expose a default slot that people can use instead of the text prop? We would need to do some check if that slot has any content in connectedCallback to be able to set onlyIcon to true it the slot is empty and there is no text prop.

_

I think that is a good idea! The component would be more flexible and I think would work better with dynamic content. I would propose to have in in another PR thou.

PR #36, opened 2023-05-10
feat(logotype): used prefix const

mJarsater:
**Describe pull-request**  
Made the logotype variables use the prefix const.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1616

Could be tested by validating the the old scss variables doesn't exist any more. The footer uses the scania wordmark, which will not show since the variable has changed.


theJohnnyMe (approved):
LGTM

PR #38, opened 2023-05-10
fix(core): removing storybook

theJohnnyMe:
**Describe pull-request**  

- Removing the Storybook from the core package
- Setting "private" on root package so we do not publish it by accident

**Solving issue**  

-  Storybook is installed on the root level and seems like it is not needed to be in the core package too.
- We might need to tweak the Netlify command for the build, let's see...
- Setting private: true on the topmost package makes it impossible to release it, but still we can use it for installing other packages

**How to test**  
1. Run the branch on your local and see if everything works as before
2. We have to see if Netlify will fail, we might need to update the build command there

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

**Additional context**  
Please, let me know if I am missing some info regarding the Storybook in the Core package.


mJarsater:
Looks like we are loosing some of our default styling. The landing page no longer has the correct fonts, not really sure why though

package.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "tegel",
-  "version": "0.0.72",
+  "name": "tegel-monorepo",
+  "version": "0.0.1",
   "description": "Tegel Digital Design System",
+  "private": true,
```
mJarsater, line 5: Nice

mJarsater (approved):
Nice cleanup in the package-lock

package.json
```diff
@@ -1,7 +1,8 @@
 {
-  "name": "tegel",
-  "version": "0.0.72",
+  "name": "tegel-monorepo",
```
timrombergjakobsson, line 2: How come we ended up with this name?

PR #40, opened 2023-05-10
feat(mixins): prefix change

mJarsater:
**Describe pull-request**  
Renamed all mixins to use a tds prefix.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1690

**How to test**  
Not really testable right now since this is gonna break all components that are using a mixin.


core/src/mixins/mixins.scss
```diff
@@ -1,6 +1,7 @@
+@import 'box-sizing';
+@import 'disabled-style';
 @import 'flex-center';
 @import 'focus-state';
 @import 'scrollbar';
 @import 'shake-animation';
-@import 'disabled-style';
-@import 'box-sizing';
+@import 'z-index';
```
theJohnnyMe, line 7: You made baby Jesus jappy for aligning this to the alphabet! 🚀

theJohnnyMe (approved):
Even complier is trowing issues as we need other components to be updated to new prefixes, it looks good to me.

mJarsater:
> 

Yeah, do we want to actually fix this in this PR? Updating the use of all the mixins. Otherwise, once we merge this, no will be able to build locally.

mJarsater:
> Even complier is trowing issues as we need other components to be updated to new prefixes, it looks good to me.

Yeah, do we want to actually fix this in this PR? Updating the use of all the mixins. Otherwise, once we merge this, no will be able to build locally.

PR #42, opened 2023-05-10
feat(mixins): prefix change

mJarsater:
**Describe pull-request**  
Renamed all mixins to use a tds prefix.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1690

**How to test**  
1. Check storybook link below
2. Everything should still work as before.


theJohnnyMe (approved):
This one is tricky to check as all components seem broken anyway as other core packages have to be updated in every single component, but now we have no build issues.
We would have to confirm on every single component that mixin in the prefix is updated too.
Nice work man! ;)

PR #47, opened 2023-05-11
docs(prefix change): changed prefix for the mode change in storybook

mJarsater:
**Describe pull-request**  
Updated the prefix on the class controlling the theming in storybook.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1691

**How to test**  
1. Go to storybook
2. Toggle light/dark mode and check the class being toggled on the body element


theJohnnyMe (approved):
Check on my local. Works as a clock!

PR #60, opened 2023-05-12
chore(link): use singular Capital case where possible

timrombergjakobsson:
**Describe pull-request**  
Updated comments and docs to use singular capital case.

**Solving issue**  
Part of: https://tegel.atlassian.net/browse/DTS-1329

**How to test**  
Go to storybook
Check in Components -> Link
Make sure it is always referenced with singular capital case (Link)


timrombergjakobsson:
> Nice! If you want to, I guess you could remove the sdds prefix on the file names also, otherwise we'll do it in another PR 🚀

Sure! I guess I could do that in this one also.

core/src/components/link/link.scss
```diff
@@ -1,4 +1,4 @@
-.sdds-link-component {
+.tds-link-component {
```
theJohnnyMe, line 1:
```suggestion
.sdds-link-component {
```
I think we need to keep this one sdds for a bit more time until we merge prefix-change branch back to the main branch.

PR #84, opened 2023-05-22
feat(card): prefix change

theJohnnyMe:
**Describe pull-request**  
Prefix change in Card component

**Solving issue**  
Fixes: DTS-1629

**How to test**  
1. Go to the Storybook link below.
2. Open the Card component
3. It needs to be visually the same as the one in the main branch.
4. Prefix is removed from file names
5. sdds prefix is replaced with tds in the files

**Additional context**  
A small change was done to the story file in the slot area by adding an attribute slot to the icon component. Icon uses size prop instead of inline style.

### DIFF
--- a/core/src/components/card/card.tsx
+++ b/core/src/components/card/card.tsx
@@ -1,11 +1,11 @@
 import { Component, h, Prop, Event, EventEmitter, Element, State, Host } from '@stencil/core';
 
 @Component({
-  tag: 'sdds-card',
-  styleUrl: 'sdds-card.scss',
+  tag: 'tds-card',
+  styleUrl: 'card.scss',
   shadow: true,
 })
-export class SddsCard {
+export class Card {
   /** Variant of the Card based on the theme used. */
   @Prop() modeVariant: 'primary' | 'secondary' = null;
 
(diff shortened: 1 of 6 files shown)


core/src/components/card/card.tsx
```diff
@@ -108,7 +109,7 @@ export class SddsCard {
             </div>
           </div>
         )}
-        {this.bodyDivider && <div class={`sdds-divider`}></div>}
+        {this.bodyDivider && <div class={`tds-divider`}></div>}
```
mJarsater, line 112: Could this use our own divider component instead? :)

core/src/components/card/card.tsx
```diff
@@ -1,11 +1,11 @@
...
 
 @Component({
-  tag: 'sdds-card',
-  styleUrl: 'sdds-card.scss',
+  tag: 'tds-card',
+  styleUrl: 'card.scss',
   shadow: true,
 })
-export class SddsCard {
+export class Card {
```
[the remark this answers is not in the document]
theJohnnyMe: Damn! Thanks for checking!

mJarsater (approved):
Nice, nice, nice! One small question on the divider used in the card. 🚀

core/src/components/card/card.tsx
theJohnnyMe: Not only that we could use it, but probably we need to use it as classes are no longer available. Great catch! I will inject the component, but most probably it won't be visible until the divider component is updated with a new prefix.

PR #93, opened 2023-05-23
fix(toggle/toast/textarea/message/dropdown-v2): added root selector to get light mode by default

mJarsater:
**Describe pull-request**  
Added root selector to get light mode by default. We were missing this on some components, which lead to variables being undefined if the user had not put sdds-mode-light as class anywhere.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1757

**How to test**  
1. Check out branch
2. Remove the wrapping sdds-mode-light class
3. Check that the toggle is using light mode by default.

**Suggested test steps**


theJohnnyMe (approved):
Really nice that you did this improvement! Great work! 
Checked by comparing the main version of the storybook and this one, works as intended!

PR #123, opened 2023-05-29
chore(readme): restructure readmes

alexandereneroth:
**Describe pull-request**  
Restructure readmes with inspiration from https://github.com/ionic-team/ionic-framework.

Tried to put only what makes sense in the github readme in the root folder, and only what makes sense on the npm package page in the ./core folder.

Some sections are duplicated since I thought they were needed on both npm and github.
Information about events are removed from the readme, since it is documented in Storybook.

**Solving issue**  
Fixes: #DTS-1667

**How to test**  
Open the branch on github and read the readmes, imagining you are a user.
While reading the readme in the core folder, imagine you are a user reading it on the npm package page https://www.npmjs.com/package/@scania/tegel

PR #125, opened 2023-05-30
docs(table): fix last uses of table lowercase

alexandereneroth:
**Describe pull-request**  
some places in docs still used table instead of Table.

**Solving issue**  
_In case of GitHub issue add # plus the number of the issue (for example #123) OR if it is Azure then AB# and number of ticket_
Fixes: #DTS-1733


mJarsater (approved):
Awesome sauce 🥫

PR #143, opened 2023-06-07
build: update to node 18

theJohnnyMe:
**Describe pull-request**  
Updating code to use version 18

**Solving issue**  
Fixes: DTS-1706

**How to test**  
1. Open this branch on your local machine
2. Try to run it on your current version of node - anything below node version 18.16 should result in error
3. Install node 18
4. Run the code again (`npm run build` and `npm run start`)
5. If everything runs OK, we are good to go
6. Check below in case you are Mac or Windows user. It would be nice to have confirmation it works on both OS.

Important to check:

- [x] Code was tested on Mac
- [x] Code was tested on Windows

**Additional context**  
I have put `postcss: false` as that feature was creating an issue for us on node 18. There is a possibility for a workaround, but I am not sure we need it as we load a CSS file (tegel.css) which is already compiled by Stencil. In version 7 of Storybook, this feature is anyway deprecated. If someone thinks this should be enabled by workarounds, please write in the comments.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -75,9 +75,5 @@
     "storybook-addon-designs": "^6.3.1",
     "storybook-dark-mode": "^3.0.0"
   },
-  "license": "MIT",
-  "engines": {
-    "npm": ">=8.0.0 <9.0.0",
-    "node": ">=16.10.0 <17.0.0"
-  }
+  "license": "MIT"
 }
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,5 @@
 # enforce node version in package.json
 engine-strict=true
+
+# Fix for storybook v6 and node 18
+node-options="--openssl-legacy-provider"
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The code conventions used in (and enforced by) Tegel is documented [here](https:
 
 ### Setting up the development environment
 
-1. Make sure you are using the required node.js version specified in `tegel/package.json` (node 16 at the time of writing).
+1. Make sure you are using the required node.js version specified in `tegel/package.json` (node 18 at the time of writing).
 2. Run `npm install` in the root directory.
 3. Run `npm install` in the core directory.
 3. Create a `.env` file in the /tegel directory with the following contents:
--- a/core/.storybook/main.js
+++ b/core/.storybook/main.js
@@ -13,6 +13,9 @@ if (process.env.STORYBOOK_ENV === 'development') {
 
 module.exports = {
   addons: addons,
+  features: {
+    postcss: false,
+  },
   stories: ['../src/**/*.stories.@(js|jsx|ts|tsx|mdx)'],
   framework: '@storybook/html',
   staticDirs: ['../public'],
--- a/core/README.md
+++ b/core/README.md
@@ -2,6 +2,9 @@
 [![Storybook](https://img.shields.io/badge/docs-storybook-ff69b4)](https://tegel-storybook.netlify.app/)
 ![](https://img.shields.io/github/license/scania-digital-design-system/tegel)
 ![Status: Beta](https://img.shields.io/badge/status-beta-red)
+![node-current](https://img.shields.io/node/v/%40scania%2Ftegel)
+
+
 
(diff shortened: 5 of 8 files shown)


core/package.json
```diff
@@ -76,7 +76,7 @@
   },
   "license": "MIT",
   "engines": {
-    "npm": ">=8.0.0 <9.0.0",
```
alexandereneroth, line 79: I think we could remove the maximum npm version and just have the minimum, since all versions beginning with 9 should not be breaking changes with other 9 versions.
Alternatively we could set <10 instead

alexandereneroth:
Hmm does not run for me.. Im getting
```
Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:71:19)
    at Object.createHash (node:crypto:133:10)
    at module.exports (/Users/alexander.eneroth/projects/scania/tegel/core/node_modules/webpack/lib/util/createHash.js:135:53)
```

theJohnnyMe:
> Looks like this is the solution: https://stackoverflow.com/a/73027407
> 
> TLDR: Its because of a SSL vulnerability in one of our dependencies. `npm audit fix --force` might fix it though.

I was getting the same error as you. Going with `npm audit fix --force` will also bump the Storybook and other dependencies after which some scripts did not work any more - probably worth checking do we want to do this migration now.

theJohnnyMe:
Upper issues are fixed in the last commit.

mJarsater:
Nice, tried it out with the latest commit and it works fine. Could there be a way of fixing it without setting a flag on the terminal command? 

Also, should we keep support for 16 to 20? Might be a big thing for some projects to go from 16 -> 18 if we force them

theJohnnyMe:
> Nice, tried it out with the latest commit and it works fine. Could there be a way of fixing it without setting a flag on the terminal command?
> 
> Also, should we keep support for 16 to 20? Might be a big thing for some projects to go from 16 -> 18 if we force them

Alex and I took a look at the issue with the flag. It looks like Storybook is requesting it in order to work properly with version 6 that we use. So we agreed to leave it there until/when we migrate to the newer version of Storybook.

Regarding the larger span for the node and npm versions, I can revert it so it accepts anything from version 16 too.

mJarsater:
> > Nice, tried it out with the latest commit and it works fine. Could there be a way of fixing it without setting a flag on the terminal command?
> > Also, should we keep support for 16 to 20? Might be a big thing for some projects to go from 16 -> 18 if we force them
> 
> Alex and I took a look at the issue with the flag. It looks like Storybook is requesting it in order to work properly with version 6 that we use. So we agreed to leave it there until/when we migrate to the newer version of Storybook.
> 
> Regarding the larger span for the node and npm versions, I can revert it so it accepts anything from version 16 too.

Nice! Sounds good to me 😄

mJarsater (dismissed):
Works fine for me! Maybe wait for a Windows user to test it also?

alexandereneroth (dismissed):
Works for me too!
⏳🪟🪟

theJohnnyMe:
Engines restriction is left only on a root level.

alexandereneroth (dismissed):
nice!

mJarsater (dismissed):

alexandereneroth (approved):
Works on my machine ™️
👏

nathalielindqvist (commented):
Runs well on Node version 18 and throws an error when using version 16, but it still runs on version 17 on a Windows machine. Is this expected behavior?

alexandereneroth:
> Runs well on Node version 18 and throws an error when using version 16, but it still runs on version 17 on a Windows machine. Is this expected behavior?

Surprising that it runs on 17, are you running it from the root folder and not the core folder?

nathalielindqvist:
> > Runs well on Node version 18 and throws an error when using version 16, but it still runs on version 17 on a Windows machine. Is this expected behavior?
> 
> Surprising that it runs on 17, are you running it from the root folder and not the core folder?

Yes running on root level 👍

nathalielindqvist (approved):
Great work!

PR #144, opened 2023-06-07
fix(card/banner/toast): align components

mJarsater:
Aligns the slots in the card, banner and toast components.
All components now have slots **and** prop for header and subheader.

**Solving issue**  
Fixes: #

**How to test**  
1. Go to storybook
2. Check in Components -> Card/Toast/Banner
3. Check the API and the subheader slot in particular.


core/src/components/card/card.tsx
```diff
@@ -75,37 +73,47 @@ export class TdsCard {
...
       )}
       <div class={`card-body`}>
         {this.bodyImg && <img class={`card-body-img`} src={this.bodyImg} alt={this.bodyImgAlt} />}
-        {this.headerPlacement === 'below' && (this.headerImg || this.header || this.subheader) && (
+        {this.headerPlacement === 'below' && (this.header || this.hasSubheaderSlot) && (
           <div class={`card-top ${this.headerPlacement}`}>
-            {this.headerImg && (
-              <img class={`card-top-image`} src={this.headerImg} alt={this.headerImgAlt} />
+            {this.hasCardThumbnailSlot && (
+              <div class={this.hasCardThumbnailSlot ? 'card-thumbnail' : 'no-header-img'}>
```
alexandereneroth, line 102:
```suggestion
              <div class='card-thumbnail'>
```
Because it will always be this value anyway because of the conditional rendering.

mJarsater: 🤦 I'll make sure we dont need the CSS first

mJarsater:
> Why do we need to conditionally render the slots? Might be a good idea, might not

Because the padding/margins and such differ if the slots are present or not. And we don't conditionally render them we will get elements in the DOM that are only there for the slots.

mJarsater:
> Instead of removing the subheader prop, could we keep both prop and slot to chose from? Prop for convenience, slot for flexibility?
> 
> Seems more user friendly

Agree, but then I'll do the same on the other component (toast/banner)

alexandereneroth:
Also, for the storybook controls - I think that, for example, the "subheader text" should just contain the content, not the slot itself. 
I guess if the text is null or an empty string we could conditionally skip adding the slot.

theJohnnyMe (changes requested):
I gave this a second thought and I am a bit worried we might make confusion for us and users. I am a bit more for either/or slot approach, but not both + making standard when we use a prop and when we use a slot. For example, should a slot be there for any kind of content injection, where the prop is for changing the type/feature/size/variation/state of a component? Let's talk about it on Monday, I will set up something.

The second thing is to prefer PR per component - just to make it easier for review.

mJarsater:
> I gave this a second thought and I am a bit worried we might make confusion for us and users. I am a bit more for either/or slot approach, but not both + making standard when we use a prop and when we use a slot. For example, should a slot be there for any kind of content injection, where the prop is for changing the type/feature/size/variation/state of a component? Let's talk about it on Monday, I will set up something.
> 
> The second thing is to prefer PR per component - just to make it easier for review.

Sure thing, let's discuss it. I just want to emphasise that I think there is great value in offering both a slot and a prop. For the subheader for instance a prop can ONLY be text. If we also open up the flexibility in user adding text + icon in the subheader. I feel like this is important, since we can't calculate for every use case and with this, atleast we leave it up to the users that need it to actually achieve it.

core/src/components/banner/banner.tsx
```diff
@@ -115,10 +129,25 @@ export class TdsBanner {
...
         <div class={`banner-content ${this.type} ${!this.icon ? 'no-icon' : ''}`}>
-          {this.header && <span class={`banner-header`}>{this.header}</span>}
-          {this.hasSubheader && <slot name="banner-subheader"></slot>}
+          {this.header ||
+            (this.hasHeaderSlot && (
+              <span class={`banner-header`}>
+                {this.header}
+                <slot name="banner-header"></slot>
+              </span>
+            ))}
```
theJohnnyMe, line 138:
```suggestion
          {(this.header || this.hasHeaderSlot) && (
              <span class={`banner-header`}>
                {this.header}
                <slot name="banner-header"></slot>
              </span>
            ))}
```

So it behaves the same way as the subheader below, meaning both prop and slot appear.

core/src/components/banner/banner.tsx
```diff
@@ -115,10 +129,25 @@ export class TdsBanner {
...
+          {(this.hasSubheaderSlot || this.subheader) && (
+            <div class="banner-subheader">
+              {this.subheader}
+              <slot name="banner-subheader"></slot>
+            </div>
+          )}
```
theJohnnyMe, line 144:
```suggestion
          {this.hasSubheaderSlot || (this.subheader && (
            <div class="banner-subheader">
              {this.subheader}
              <slot name="banner-subheader"></slot>
            </div>
          ))}
```

Or tweak this behavior instead?

mJarsater:
Closing this PR and opening one PR per component instead.

PR #146, opened 2023-06-07
feat(tooltip): update method for tooltip

mJarsater:
**Describe pull-request**  
Update method for tooltop.

**Solving issue**  
Fixes: -

**How to test**  
1. Checkout the branch.
2. Try moving a tooltip target in runtime.
3. Call the update method
4. Make sure the tooltips position is updated.


alexandereneroth (changes requested):
Since the component is already called tooltip, I'd skip that in the method name. Maybe something like "updatePosition" or just "update" would be sufficient.

Another small thing; the comment at the end of the file that says something about a slot on line 118 should be changed.

alexandereneroth:
> > Since the component is already called tooltip, I'd skip that in the method name. Maybe something like "updatePosition" or just "update" would be sufficient.
> > Another small thing; the comment at the end of the file that says something about a slot on line 118 should be changed.
> 
> Not sure I agree. We have hideToast, showBanner, hideBanner, etc already. I would like to keep it as consistent possible.
> 
> I'll have a look at comment 👀

Ok didn't know we have that already. Then it's fine I guess.

PR #148, opened 2023-06-12
fix(button): locked in disabled state

theJohnnyMe:
**Describe pull-request**  
Possible fix for the forever-lasting disabled state of the button.

**Solving issue**  
When the button is set statically to a disabled state on the initial load, it seems to remain in it.
connectedCallback is run only once, on the initial load so the component is not aware of the change of "disabled" prop change. We add click listener on it, but do not remove it which makes the button remain in the disabled state even when disabled is false.
Martin and I took a look and proposed a solution where we remove logic that sets it disabled via stoppingPropagation and preventingDefault. Now button uses only CSS rule of pointer-events:none and we leave it up to user to stopPorpagation or preventDefault. 

**How to test**  
Not sure if is this testable without a release, suggestion is to try via linking or packing.

### DIFF
--- a/core/src/components/button/button.tsx
+++ b/core/src/components/button/button.tsx
@@ -1,5 +1,5 @@
-import { Component, Element, h, Host, Prop, State } from '@stencil/core';
-import { disableClickOnElement } from '../../utils/utils';
+import { Component, Element, h, Host, Prop, State, Watch } from '@stencil/core';
+import { disableClickOnElement, enableClickOnElement } from '../../utils/utils';
 
 @Component({
   tag: 'tds-button',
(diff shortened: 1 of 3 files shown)


core/src/components/button/button.tsx
```diff
@@ -45,6 +42,27 @@ export class TdsButton {
...
+    }
+  }
+
+  @Watch('disabled')
+  updateValue(newValue: boolean) {
+    if (newValue === true) {
+      this.enableComponent();
+    } else this.disableComponent();
+  }
+
```
mJarsater, line 65: Im not sure this is needed.

core/src/components/button/button.tsx
```diff
@@ -33,10 +33,7 @@ export class TdsButton {
   connectedCallback() {
     this.slotElement = this.host.children ? this.host.children[0] : null;
     if (this.disabled) {
```
mJarsater, line 35: What happens if we move this one to componentWillRender? That way it will run on every render, and we can check if we disabled === true and do what we need.

mJarsater:
@theJohnnyMe do you want to update the description of this one?

theJohnnyMe:
> @theJohnnyMe do you want to update the description of this one?

Updated! ;)

mJarsater (approved):

PR #174, opened 2023-06-28
fix(breadcrumbs): last element and disabled corrections

mJarsater:
**Describe pull-request**  
Removed the logic for checking and added a last class to the last breadcumb and instead implemented CSS to handle this. Also removed the disabled prop since it's only used to show "current" in design.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-2042, https://tegel.atlassian.net/browse/DTS-2043

**How to test**  
1. Go to Breadbrumb
2. Check that the styling is correct for all items.

### DIFF
--- a/core/src/components/breadcrumbs/breadcrumb/breadcrumb.scss
+++ b/core/src/components/breadcrumbs/breadcrumb/breadcrumb.scss
@@ -17,13 +17,12 @@
     @include tds-focus-state;
   }
 
-  &.disabled,
   &.current,
   [aria-current='page'] {
     ::slotted(*) {
       pointer-events: none;
       cursor: default;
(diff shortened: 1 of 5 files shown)


core/src/components/breadcrumbs/breadcrumb/breadcrumb.scss
```diff
@@ -45,7 +44,7 @@
   }
 }
 
-:host(.last) {
+:host(:last-of-type) {
```
theJohnnyMe, line 47: Cool!

theJohnnyMe (approved):
LGTM

alexandereneroth (approved):
🥳

PR #176, opened 2023-06-28
fix(tabs): no event emitted on disabled tabs

mJarsater:
**Describe pull-request**  
Adds an if check for disabled state in the event handler for click. This makes sure that the tdsChange is only emitted if the tab in not disabled. This also enables a tab to disabled/enabled in runtime and the functionality will still work. 

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-1835

**How to test**  
1. Go to Folder/Inline/Navigation Tabs
2. Try clicking the disabled tab.
3. Make sure it doesn't change the content.


alexandereneroth (approved):

nathalielindqvist (approved):
🌷

PR #179, opened 2023-06-29
fix(side-menu): slots

theJohnnyMe:
**Describe pull-request**  
Aligning slots naming per our convention

**Solving issue**  
Fixes: [DTS-1957](https://tegel.atlassian.net/browse/DTS-1957)

**What was done:**
 - document all slot and their names in readme files
 - use `<default>` to indicate those are not named for purpose, can be used for side-menu sub-components ect...
 - show the user only Notes they really need to know about (show one used in stories, hide internal ones)

**How to test**  
1. Go to Netlify PR preview link
4. Check that the Side-menu component and pattern stories still look and function OK.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

[DTS-1957]: https://tegel.atlassian.net/browse/DTS-1957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/components/side-menu/readme.md
+++ b/core/src/components/side-menu/readme.md
@@ -23,13 +23,13 @@
 
 ## Slots
 
-| Slot             | Description                                                                            |
-| ---------------- | -------------------------------------------------------------------------------------- |
-| `"UNNAMED-SLOT"` | Used for nesting upper/top content in Side menu                                        |
-| `"close-button"` | Used for injection of tds-side-menu-close-button that is show when in mobile view      |
-| `"end"`          | Used for items that are presented at the bottom of the side menu, eg. profile settings |
-| `"overlay"`      | Used of injection of tds-side-menu-overlay                                             |
-| `"sticky-end"`   | Used for tds-side-menu-collapse-button component                                       |
+| Slot             | Description                                                                                                                                                  |
+| ---------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `"<default>"`    | <b>Unnamed slot.</b> Used for nesting main content of Side Menu, e.g. <code><tds-side-menu-item></code> and <code><tds-side-menu-dropdown></code> components |
+| `"close-button"` | Used for injection of tds-side-menu-close-button that is show when in mobile view                                                                            |
+| `"end"`          | Used for items that are presented at the bottom of the Side Menu, e.g. profile settings                                                                      |
+| `"overlay"`      | Used of injection of tds-side-menu-overlay                                                                                                                   |
+| `"sticky-end"`   | Used for tds-side-menu-collapse-button component                                                                                                             |
 
 
 ----------------------------------------------
--- a/core/src/components/side-menu/readme.md
+++ b/core/src/components/side-menu/readme.md
@@ -1,22 +1,5 @@
 # tds-side-menu
 
-### Side Menu is a component that consists of these sub-components:
-
- - side-menu-user-label
- - side-menu-user-image
- - side-menu-user
- - side-menu-overlay
- - side-menu-item
- - side-menu-dropdown
- - side-menu-dropdown-list
- - side-menu-dropdown-list-item
- - side-menu-collapse-button
- - side-menu-close-button
-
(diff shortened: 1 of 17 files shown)


core/src/components/side-menu/readme.md
```diff
@@ -21,6 +21,17 @@
 | `tdsCollapse` | Event that is emitted when the Side Menu is collapsed. | `CustomEvent<{ collapsed: boolean; }>` |
 
 
+## Slots
+
+| Slot             | Description                                                                            |
+| ---------------- | -------------------------------------------------------------------------------------- |
+| `"UNNAMED-SLOT"` | Used for nesting upper/top content in Side menu                                        |
```
mJarsater, line 28: I think this is a great question. But my feeling is that these should not be named at all. The should remain as unnamed slots, and in the readme we either don't include it for now. Or we document it as unnamed/default slot.

mJarsater: I think the reason for not including it in the readme would be that just adding an item, without a slot name is the expected behaviour. So to follow convention I guess this shouldn't be documented

mJarsater: This is how liquid handles it, so the unnamed slot it just blank. 

Screenshot 2023-07-03 at 10 23 14

nathalielindqvist (commented):
The Side Menu does not close when clicking the collapse button on the "Many Navigation Items" pattern example, but it does close in our current Storybook link. Should this be addressed?

https://github.com/scania-digital-design-system/tegel/assets/48509966/2934cd62-2ca6-457b-ba24-8f831c9156b2

mJarsater:
> The Side Menu does not close when clicking the collapse button on the "Many Navigation Items" pattern example, but it does close in our current Storybook link. Should this be addressed?
> 
>  Side.Menu.not.closing.Many.Nav.Items.example.mp4

Fixed in another issue, just needed a rebase.

core/src/components/side-menu/side-menu.tsx
```diff
@@ -192,6 +201,7 @@ export class TdsSideMenu {
             <slot name="close-button"></slot>
             <div class="tds-side-menu-wrapper">
               <ul class={`tds-side-menu-list tds-side-menu-list-upper`}>
+                {/* FIXME: Change to start-end, top-bottom and name the slot */}
```
alexandereneroth, line 204: I think we decided not to do this right?
```suggestion
```

core/src/components/side-menu/readme.md
```diff
@@ -1,6 +1,21 @@
...
+ - side-menu-overlay
+ - side-menu-item
+ - side-menu-dropdown
+ - side-menu-dropdown-list
+ - side-menu-dropdown-list-item
+ - side-menu-collapse-button
+ - side-menu-close-button
+
+Some of the upper mentioned subcomponents are not listed under <i>Notes</i> tab
+because either they are used only internally or they have no props or events but serve as wrappers.
```
alexandereneroth, line 17: I think it's nice if subcomponents that are only wrappers still are listed under Notes, so it's clear that they exist and what props they have, even if they don't have any.

Also, if a component is only used internally I think it's better to not mention it in documentation at all, so it remains an implementation detail that we can change and/or remove easily and without updating the docs.

alexandereneroth (changes requested):

core/src/components/side-menu/readme.md
theJohnnyMe: Good comment, I will double check then which one we use as wrappers as expose then and keep hidden ones that are used internally ;)

theJohnnyMe: PR updated! ;)

alexandereneroth (approved):
Nice work! LGTM

mJarsater (approved):
LGTM

nathalielindqvist (approved):
👌

PR #185, opened 2023-07-03
fix(side-menu): z-index

mJarsater:
**Describe pull-request**  
Updated z-index to only take affect on smaller screens. This fixes the issue of a user on small screen opening the side-menu -> making the window bigger -> side-menu now overlaps header.

**Solving issue**  
Fixes: 

**Screenshots**  
Screenshot 2023-07-03 at 12 25 12

### DIFF
--- a/core/src/components/side-menu/side-menu.scss
+++ b/core/src/components/side-menu/side-menu.scss
@@ -64,6 +64,13 @@
   }
 }
 
+@media (max-width: $grid-lg) {
+  :host([open]) {
+    pointer-events: auto;
+    z-index: tds-z-index(header) + 1;
+  }
+}
+
 @media (min-width: $grid-lg) {
   :host([persistent]) {
     pointer-events: auto;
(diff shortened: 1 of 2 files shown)


nathalielindqvist (commented):
Looks like the close button on the mobile Side Menu is hidden behind the header, could it have happened in this PR?

image

core/src/components/side-menu/side-menu.scss
```diff
@@ -70,6 +65,11 @@
 }
 
 @media (min-width: $grid-lg) {
+  :host([open]) {
+    pointer-events: auto;
+    z-index: tds-z-index(header) + 1;
+  }
+
```
mJarsater, line 72:
```suggestion
```

mJarsater:
> Looks like the close button on the mobile Side Menu is hidden behind the header, could it have happened in this PR?
> 
> image

Faulty media query, should be updated now.

nathalielindqvist (dismissed):
🙌

alexandereneroth (dismissed):
LGTM, however the collapse button is not working. 
I think that broke in an earlier PR though.

alexandereneroth (approved):
That issue above seems unlikely to be related to this PR

nathalielindqvist (approved):

PR #187, opened 2023-07-03
fix(tabs): set transition on hover

mJarsater:
**Describe pull-request**  
Sets the width transition on hover to not the the flashing on page load.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-2006

**How to test**  
1. Go to Inline/Navigation Tabs
2. Check the animation on the selected tab on load.
3. It should not transition.

### DIFF
--- a/core/src/components/tabs/inline-tabs/inline-tab/inline-tab.scss
+++ b/core/src/components/tabs/inline-tabs/inline-tab/inline-tab.scss
@@ -28,6 +28,11 @@
     @include tds-focus-state;
   }
 
+  .inline-tab-item:not(.selected)::after {
+    width: 0%;
+    transition: width 0.15s ease-in-out 0s;
+  }
+
   .inline-tab-item {
     position: relative;
     margin-right: 32px;
@@ -53,10 +58,6 @@
     height: 2px;
     background-color: var(--tds-inline-tabs-tab-indicator-background-hover);
     z-index: 1;
-  }
-
-  .inline-tab-item:hover::after {
-    transition: width 0.15s ease-in-out 0s;
   }
 
   .selected {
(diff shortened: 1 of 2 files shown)


nathalielindqvist (dismissed):
LGTM!

alexandereneroth (changes requested):
Putting the transition only on :hover will remove the "exit" animation. Since when it is no longer hovered a transition will no longer be defined on it.
Reference: https://stackoverflow.com/questions/48866330/transition-only-on-hover-in 

If you look on the deployed version you can see that there is a "hover out" animation: https://tegel-storybook.netlify.app/?path=/story/components--inline-tabs

I am not sure that the issue that this PR attempts to address can be easily fixed another way though.

Although my understanding is that It shouldn't occur if CSS is loaded in a proper way.

There is apparently a Chrome bug that causes this though https://stackoverflow.com/a/42969608.

Some are suggesting adding an empty script tag at the end of body fixes it: https://stackoverflow.com/a/40095851

Otherwise a workaround like this would work, but are perhaps not ideal to include in a component library: https://css-tricks.com/transitions-only-after-page-load/

mJarsater:
> Putting the transition only on :hover will remove the "exit" animation - since when it is no longer hovered a transition will no longer be defined on it. Reference: https://stackoverflow.com/questions/48866330/transition-only-on-hover-in
> 
> If you look on the deployed version you can see that there is a "hover out" animation: https://tegel-storybook.netlify.app/?path=/story/components--inline-tabs
> 
> I am not sure that the issue that this PR attempts to address can be easily fixed another way though.
> 
> Although my understanding is that It shouldn't occur if CSS is loaded in a proper way.
> 
> There is apparently a Chrome bug that causes this though https://stackoverflow.com/a/42969608.
> 
> Some are suggesting adding an empty script tag at the end of body fixes it: https://stackoverflow.com/a/40095851
> 
> Otherwise a workaround like this would work, but are perhaps not ideal to include in a component library: https://css-tricks.com/transitions-only-after-page-load/

Updated the implementation now, please have a look whenever 😄

alexandereneroth (approved):
True, the selected tab doesn't need to animate at all, nice!

nathalielindqvist (approved):

PR #191, opened 2023-07-03
docs(dropdown): standardise order of controls

mJarsater:
**Describe pull-request**  
Standardise the order of the controls. I chose to put "advanced" controls deafultOption and optionType at the bottom since most users wont bother with those.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-2037

**How to test**  
1. Go to Dropdown
2. Check the order of the controls.


nathalielindqvist (commented):
Nicely done! If we want to be consistent with other components the 'disabled' prop is usually at the very bottom in the list. Also, do you see any benefit in putting the argTypes in the corresponding order as well? I've done that with the other components, but that might just be my preference :)

nathalielindqvist (approved):
Good point! Looks great 😄

theJohnnyMe (approved):
LGTM

PR #196, opened 2023-07-05
fix(table): removed dummy-data prop

mJarsater:
BREAKING CHANGE: Removed enable-dummy-data prop

**Describe pull-request**  
Removed dummy-data prop and updated stories to not use the prop.

**Solving issue**  
Fixes: https://tegel.atlassian.net/browse/DTS-2056

**How to test**  
1. Go to storybook
2. Check in Table -> All stories
3. Make sure they all still look as intended.

### DIFF
--- a/core/src/components/table/table-component-batch-actions.stories.tsx
+++ b/core/src/components/table/table-component-batch-actions.stories.tsx
@@ -9,6 +9,7 @@ import tdsBodyRowExpandable from './table-body-row-expandable/readme.md';
 import tdsBodyCell from './table-body-cell/readme.md';
 import tdsTableFooter from './table-footer/readme.md';
 import { ComponentsFolder } from '../../utils/constants';
+import dummyData from './table-body/dummy-data.json';
 
 export default {
   title: `${ComponentsFolder}/Table`,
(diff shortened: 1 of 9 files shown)


core/src/components/table/table-component-batch-actions.stories.tsx
```diff
@@ -179,7 +180,7 @@ const BatchActionTemplate = ({
                 column4Width ? `style="width: ${column4Width};"` : ''
               } column-key='mileage' column-title='Mileage' text-align='right'></tds-header-cell>
           </tds-table-header>
-          <tds-table-body enable-dummy-data>
```
alexandereneroth, line 182: =/ so we do support setting data as a string. 
Should we create a ticket to change this?

theJohnnyMe (dismissed):
🤘

core/src/components/table/table-component-batch-actions.stories.tsx
theJohnnyMe: Yes, I think it can be done the same way it was done for dropdown options instead...

alexandereneroth (approved):

core/src/components/table/table-component-batch-actions.stories.tsx
mJarsater: Added an update in the last commit. The only downside that this doesn't work in the docs tab, since storybook can't run js in there.

alexandereneroth (dismissed):

theJohnnyMe (approved):
👏 Great work!

alexandereneroth (approved):

theJohnnyMe (approved):

alexandereneroth (approved):
💯

PR #203, opened 2023-07-06
fix(table:sorting): fixed width on icon to prevent content jump

theJohnnyMe:
**Describe pull-request**  
Content jumping on sorting in react-demo-apge

**Solving issue**  
Fixes: [DTS-2005](https://tegel.atlassian.net/browse/DTS-2005)

**How to test**  
1. Go to react demo example
2. Open table component
3. Try to sort it - content should move left/right a bit
4. Now inspect the element and on `tds-table__header-button-icon` add width: 16px.
5. Jumping left and right should stop now.
6. I have added that rule in CSS in this PR...

**Note**
The issue seems not to be visible in Storybook, therefore testing in React.

[DTS-2005]: https://tegel.atlassian.net/browse/DTS-2005?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/components/table/table-header-cell/table-header-cell.scss
+++ b/core/src/components/table/table-header-cell/table-header-cell.scss
@@ -59,6 +59,7 @@
       /* not to shrink on lot of text */
       flex: 0 0 16px;
       height: 16px;
+      width: 16px;
       opacity: 0;
       transform-origin: center;
       transition: opacity 200ms ease-in, transform 200ms ease;


alexandereneroth (approved):
nice

nathalielindqvist (approved):
Works great, good catch! This is not part of this PR, but I don't think the sortable Table actually sorts the content on the React demo page? Is the event defaultPrevented here as well?

https://github.com/scania-digital-design-system/tegel/assets/48509966/3d879fc5-9682-411e-90fd-84486d4294c1

theJohnnyMe:
> Works great, good catch! This is not part of this PR, but I don't think the sortable Table actually sorts the content on the React demo page? Is the event defaultPrevented here as well?
> 
>  sortable.Table.not.sorting.content.React.Demo.mp4

I could not reproduce this on tegel-react-demo page. Can you please double-check it and give me a call if you can reporduce it on your side? The upper example there works and one on bottom is prevented to show examples on events. However, makes sense to update IDs of table or columnsKey as other table sort too on trigger on upper one.

PR #207, opened 2023-07-10
feat: use custom uid utility fn instead of crypto fn

alexandereneroth:
**Describe pull-request**  
Unique identifier strings are now generated with a custom local utility function instead of using the browsers built-in crypto UUID function - since that function seemed to not always be supported across different environments.

**Solving issue**  
Fixes: #DTS-2053

**How to test**  
Checkout repo locally and check that the crypto function is not used anymore.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


nathalielindqvist (approved):
LGTM! ⭐

theJohnnyMe (approved):
Checked and works nice!

PR #210, opened 2023-07-11
refactor(popover, tooltip): share code instead of duplicating

alexandereneroth:
Moves shared logic of popover-menu, popover-canvas, and tooltip into a core component which they are all composed out of. 
This makes it so we wont have to fix bugs related to popovers in three separate places.

**How to test**  
1. Go to storybook
2. Check in the popover-canvas, popover-menu, and tooltip and verify everything works as expected.
3. Come up with your own method to verify if something does not work as expected.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


theJohnnyMe (approved):
Really nice approach in order to respect DRY principle! Great work Alex!

core/src/components/tooltip/tooltip.tsx
```diff
@@ -26,112 +26,57 @@ export class TdsTooltip {
...
+        if (state.placement === 'bottom-start' || state.placement === 'right-start') {
+          this.border = 'top-left';
+        } else if (state.placement === 'bottom-end' || state.placement === 'left-start') {
+          this.border = 'top-right';
+        } else if (state.placement === 'top-end' || state.placement === 'left-end') {
+          this.border = 'bottom-right';
+        } else if (state.placement === 'top-start' || state.placement === 'right-end') {
+          this.border = 'bottom-left';
+        } else if (state.placement === 'bottom' || state.placement === 'top') {
+          this.border = 'default';
```
nathalielindqvist, line 50: Could this maybe be a switch case? Still learning when to use and not use them so might be wrong :)

nathalielindqvist (commented):
Looks like the content have switched place a bit in the Popover Canvas, is that intentional? Screenshot from this PR:
image

Compared to current Storybook:
image

core/src/components/tooltip/tooltip.tsx
nathalielindqvist: Doesn't need to be part of this PR, mostly just curious ☺️

alexandereneroth: Yes it could be!
It comes down to readability in this case.
The equivalent switch statement is more lines but maybe more readable?

```js
switch(state.placement) {
  case 'bottom-start':
  case 'right-start':
    this.border = 'top-left';
    break;
  case 'bottom-end':
  case 'left-start':
    this.border = 'top-right';
    break;
  case 'top-end':
  case 'left-end':
    this.border = 'bottom-right';
    break;
  case 'top-start':
  case 'right-end':
    this.border = 'bottom-left';
    break;
  case 'bottom':
  case 'top':
  default:
    this.border = 'default';
}
```

alexandereneroth:
```js
[
  ['top-left', ['bottom-start', 'right-start']],
  ['top-right', ['bottom-end', 'left-start']],
  ['bottom-right', ['top-end', 'left-end']],
  ['bottom-left', ['top-start', 'right-end']],
  ['default', ['bottom', 'top']]
].forEach(([border, placements]) => {
  if(placements.includes(state.placement)) {
    this.border = border;
  } else {
    this.border = 'default';
  }
})
```
Another alternative perhaps

nathalielindqvist (approved):
Works well! And kudos for avoiding template literals 🤩

theJohnnyMe (approved):

PR #211, opened 2023-07-11
fix(slider): disabled state update in runtime

nathalielindqvist:
**Describe pull-request**  
Refactoring the way class name for disabled state is set. The use of template literals stops the component updating in runtime.

**Solving issue**  
Fixes: https://tegel.atlassian.net/jira/software/projects/DTS/boards/1?selectedIssue=DTS-2064

**How to test**  
1. Check out branch
2. CD into core
3. Run npm pack
4. Copy generated file into tegel-react-demo local repo on root level
5. In package.json, update path to @scania/tegel to the generated file name (example: "@scania/tegel": "file:scania-tegel-0.0.9.tgz",)
6. Run npm run build (this might cause some errors since we've made changes in Tegel that haven't been released yet. Remove what you don't need for testing the Slider.
7. Run npm run start.
8. Check that the Slider disabled state can be update in runtime.

**Screenshots**  
image

**Additional context**  
Using template literals, like the implementation seen in the screenshot above, restricts the components from updating class names in runtime. This might be useful to know moving forward.

### DIFF
--- a/core/src/components/slider/slider-vars.scss
+++ b/core/src/components/slider/slider-vars.scss
@@ -4,7 +4,7 @@
   --tds-slider-thumb-color: var(--tds-blue-800);
   --tds-slider-divider-color: var(--tds-grey-400);
   --tds-slider-track-color: var(--tds-grey-400);
-  --tds-slider-track-fill-color: var(--tds-blue-800);
+  --tds-slider-track-fill-color: red;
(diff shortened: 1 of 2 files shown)


alexandereneroth (approved):
Learned something new!
I think the updated code is more readable too which is a plus 🎉

theJohnnyMe (approved):
Awesome!

PR #216, opened 2023-07-14
fix: prefix change banner slot name

alexandereneroth:
**Describe pull-request**  
Corrects a slot name


theJohnnyMe (approved):
Great catch!

PR #240, opened 2023-08-07
docs(css reset): slimmed down css reset

mJarsater:
**Describe pull-request**  
Slimmed down our CSS reset to not impact out components.

**Solving issue**  
Fixes: [DTS-2055](https://tegel.atlassian.net/browse/DTS-2055)

**How to test**  
1. Go to storybook
2. Check in CSS Reset

[DTS-2055]: https://tegel.atlassian.net/browse/DTS-2055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/stories/css-reset.stories.mdx
+++ b/core/src/stories/css-reset.stories.mdx
@@ -15,25 +15,13 @@ For an in-depth explanation of each part we recommend reading the article on his
 
 
 ```css
-
 /* BEGIN - CSS Reset */
 
 /*
-  Use a more intuitive box-sizing model.
-*/
-*, *::before, *::after {
-  box-sizing: border-box;
-}
-/*
-  Remove default margin.
-*/
-* {
-  margin: 0;
-}
-/*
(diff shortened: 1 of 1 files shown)


theJohnnyMe (approved):
LGTM 🤘

alexandereneroth:
We should revisit this, if our components dont work with the rules you removed they should probably be fixed. These are very common css rules that many of our users likely have.

alexandereneroth:
> > We should revisit this, if our components dont work with the rules you removed they should probably be fixed. These are very common css rules that many of our users likely have.
> 
> Add it to the backlog. The problem is that we would need to build our component based on the reset. Which is doable, but a lot of work and also requires everyone to use the reset.

Ideally shadow dom would help us avoid conflicting styles, but yea we haven't been able to use it consistently.

PR #243, opened 2023-08-09
fix(side-menu): pointer-events, css classes instead of attributes, z-index

theJohnnyMe:
**Describe pull-request**  

- Using CSS classes instead of attribute selectors in CSS to prevent applying rules on side-menu even when menu is not opened/persistent/collapssed. For example, setting them in frameworks as open="false" makes open appear as attribute event it is not used, but CSS rule is applied.
- setting pointer-events:none on "div" of overlay so no clicks behind are possible while still leaving option to listen to a click on host tds-side-menu-overlay itself.
- setting z-index to side-menu just after the header index in z-index array and assigning it to the tds-side-menu component to prevent particles of components like input elements from radio-button, checkboxes or buttons to be over.

**Solving issue**  
Fixes: [DTS-2173](https://tegel.atlassian.net/browse/DTS-2173)
Fixes: [DTS-2156](https://tegel.atlassian.net/browse/DTS-2156)
Fixes: [DTS-2175](https://tegel.atlassian.net/browse/DTS-2175)

**How to test**  
1. Some of things might be able to test in Storybook, as pointer-events, for other I would recommend building pack and testing it out it React.

**Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

**Additional context**  
_Add any other context about the pull-request here._

[DTS-2173]: https://tegel.atlassian.net/browse/DTS-2173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DTS-2156]: https://tegel.atlassian.net/browse/DTS-2156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DTS-2175]: https://tegel.atlassian.net/browse/DTS-2175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/src/components/side-menu/side-menu.scss
```diff
@@ -65,14 +66,14 @@
 }
 
 @media (max-width: $grid-lg) {
-  :host([open]) {
+  :host(.menu-opened) {
```
mJarsater, line 69: Does [open="true"] not work as we want it here?

theJohnnyMe: I was expecting it to work the same way as disabled, so to be there but have a value of true or false, but it did not work. Then I tried setting it as hard-coded attribute on <Host, did not work either. Also I have played with removing reflect and putting it back for open prop, did not do much of change. 
Using class in CSS was only thing I could think of as then devs can have "open" prop all the time in the code and just change toggle the value from true to false while CSS will act differently if class is there or not.
Do you have some other idea how we could do it?

mJarsater: No other ideas, was more wondering why you went with classes intsead of attr. The only small thing for me would be to change from ".menu-opened" to just ".open"?  And the same for ".menu-persistent" -> ".persistent".

To me it's just to keep the names as short as possible and the prefix of menu- doesn't really add anything since we are already withing the side-menu. What do you think?

core/src/components/side-menu/side-menu-overlay/side-menu-overlay.scss
```diff
@@ -1,8 +1,15 @@
...
   background-color: black;
   transition: opacity 0.4s linear;
   opacity: 0;
 }
+
+div {
+  position: fixed;
+  width: 100%;
+  height: 100%;
+  pointer-events: none;
```
mJarsater, line 14: Did you test this with adding a element behind and try pressing it? And also with the combination of closing the side-menu but not triggering a click on the element behind?

core/src/components/side-menu/side-menu.scss
mJarsater, line 1: By the way, should we remove the animation of the side menu?

mJarsater: It is causing some performance issues when you click open and close quickly.

core/src/components/side-menu/side-menu.scss
theJohnnyMe: I will try to shorten it, it was bit off in the beginning as we have props with same names so my IDE was bit confused, but I will double check it. ;)

core/src/components/side-menu/side-menu-overlay/side-menu-overlay.scss
theJohnnyMe: I made small change in this commit: https://github.com/scania-digital-design-system/tegel/pull/243/commits/8582d6e11ac836f394c11dc0bda3e156f5875455 
in order to test it, it works as intended. However, a proper test would be nice to conduct by packing and testing in react demo or angular one.

core/src/components/side-menu/side-menu.scss
theJohnnyMe: Post-discussion conclusion: we leave it as it for now.

core/src/components/side-menu/side-menu-overlay/side-menu-overlay.scss
theJohnnyMe: Post-discussion comment: Functionality work as intended. Will be tested with npm pack in tegel-react-demo page.

mJarsater (approved):
LGTM

PR #246, opened 2023-08-10
Feat/header filter

HasanVatansever:
**Describe pull-request**  
Adds functionality in header for filtering columns

**How to test**  
1. Go to  storybook
2. Check in Table > Header Filter

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

**Screenshots**  
image


mJarsater:
I think this is a great first draft but I'm not sure about the current implementation. This to me feels like a feature request that first of all would need to go through design investigation to figure out the use case and see what the solution might be. I also feel like it should be a slot rather than a prop. Seeing as something else might be slotted into this position.

mJarsater:
Closing this issue for now. We have noted this as a feature request.

PR #249, opened 2023-08-16
fix(button): fixed usage of label slot

mJarsater:
**Describe pull-request**  
In the check for if it in an only-icon button we also need to check if they are using the slot rather than the text prop.
Before this fix the button would be treated as a only-icon button if they used the label slot.

Note: I have not added any slot/prop toggle in storybook. When testing you need to checkout the branch and implement the slot usage on your own.

**Solving issue**  
Fixes: [CDEP-2196](https://tegel.atlassian.net/browse/CDEP-2196)

**How to test**  
1. Check out branch
2. In story file, use the slot="label" rather than the text prop.
3. See that the button still works as intended.

[CDEP-2196]: https://tegel.atlassian.net/browse/CDEP-2196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/components/button/button.scss
+++ b/core/src/components/button/button.scss
@@ -215,31 +215,33 @@ button {
   justify-content: center;
 }
 
-tds-button:not([only-icon]) {
-  display: inline-flex;
-  align-items: center;
+tds-button {
+  :not(.only-icon) {
+    display: inline-flex;
+    align-items: center;
 
-  .sm {
-    ::slotted([slot='icon']) {
-      margin-left: var(--tds-spacing-element-12);
-      width: var(--tds-spacing-element-16);
-      height: var(--tds-spacing-element-16);
+    &.sm {
+      ::slotted([slot='icon']) {
+        margin-left: var(--tds-spacing-element-12);
+        width: var(--tds-spacing-element-16);
+        height: var(--tds-spacing-element-16);
+      }
     }
-  }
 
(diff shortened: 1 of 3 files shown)


nathalielindqvist (dismissed):
Looks good!

theJohnnyMe (dismissed):
LGTM
Side note: I would like that we change it so we do not set attribute only-icon, but rather use CSS class. The button component is in need for refactoring and simplification.

mJarsater:
@theJohnnyMe added a commit to remove the setting of attribute, we can leave it out of this PR if you want to:
https://github.com/scania-digital-design-system/tegel/pull/249/commits/dcb064cae6853780a2529e71f20cbf3d70b04db9

theJohnnyMe:
> @theJohnnyMe added a commit to remove the setting of attribute, we can leave it out of this PR if you want to: [dcb064c](https://github.com/scania-digital-design-system/tegel/commit/dcb064cae6853780a2529e71f20cbf3d70b04db9)

Great work buddy, thanks for sorting it out quickly! ;)

theJohnnyMe (approved):

nathalielindqvist (approved):
👍

PR #269, opened 2023-08-31
fix(modal): slot alignment

mJarsater:
**Describe pull-request**  
Renamed action slot to conform with other components.

**Solving issue**  
Fixes: [CDEP-2172](https://tegel.atlassian.net/browse/CDEP-2172)

**How to test**  
1. Go to Modal
2. Check notes and the slot names.
3. Verify that the Modal still works as expected.

[CDEP-2172]: https://tegel.atlassian.net/browse/CDEP-2172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/components/modal/modal.scss
+++ b/core/src/components/modal/modal.scss
@@ -180,23 +180,19 @@ $modals: (
       overflow-y: auto;
     }
 
-    slot[name='actions'] {
+    slot[name='bottom'] {
       position: absolute;
       bottom: -1px;
       left: 0;
       right: 0;
       background-color: var(--tds-modal-background);
       padding: var(--tds-spacing-element-16);
-      display: flex;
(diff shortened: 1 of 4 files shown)


core/src/components/modal/modal.scss
```diff
@@ -180,23 +180,19 @@ $modals: (
       overflow-y: auto;
     }
 
-    slot[name='actions'] {
+    slot[name='bottom'] {
```
nathalielindqvist, line 183: If the Modal body text is long it cuts off at the bottom, hiding the action buttons. Is this something we should address or is it up to the user to adjust padding as needed?
image

PR #272, opened 2023-09-01
fix(tabs): watcher for selectedIndex & defaultSelectedIndex

mJarsater:
**Describe pull-request**  
Added a watcher for the default-selected-index & selected-index in order to keep the state aligned with the attribute.

**Solving issue**  
Fixes: - 

**How to test**  
1. Go the the react test environment: https://github.com/scania-digital-design-system/tegel-react-demo/pull/130
2. Check under Tabs -> Link
3. Try switching tab and see that all tabs change the selected tab

### DIFF
--- a/core/src/components/tabs/folder-tabs/folder-tabs.tsx
+++ b/core/src/components/tabs/folder-tabs/folder-tabs.tsx
@@ -77,6 +77,15 @@ export class TdsFolderTabs {
     };
   }
 
+  @Watch('selectedIndex')
+  handleSelectedIndexUpdate() {
+    this.children = Array.from(this.host.children).map((tabElement: HTMLTdsFolderTabElement) => {
+      tabElement.setSelected(false);
+      return tabElement;
+    });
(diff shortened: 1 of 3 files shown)


nathalielindqvist (approved):
Sweet! 🍬

alexandereneroth (approved):
I have a hunch that we might not want to set this on defaultSelectedIndex.

But nothing more than a hunch - so lets go with this for now.

alexandereneroth (changes requested):
Never-mind, I think this is a code smell. 

I think that listening for changes to the defaultIndex could lead to users using that instead of setting index directly, and could lead to unpredictable behaviour. 

I would suggest that rather than reacting to changes to defaultIndex, we leave it out and evaluate the kind of user issues/requests we see. 

I think that probably all instances where a user want to change defaultIndex post initialisation should have been done with selectedIndex instead.

alexandereneroth (approved):

timrombergjakobsson (approved):

PR #279, opened 2023-09-05
feat(popover menu): design aligement

mJarsater:
**Describe pull-request**  
Added fluid prop to allow for wider popover-menu. Aligned padding with design.

**Solving issue**  
Fixes: [CDEP-2442](https://tegel.atlassian.net/browse/CDEP-2442)

**How to test**  
1. Go to Popover Menu
2. Check that the component is aligned with design
3. Check the padding on the divider
4. Check the general padding
5. Test the fluid control
6. Make sure the fluid version is aligned with design

[CDEP-2442]: https://tegel.atlassian.net/browse/CDEP-2442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alexandereneroth (commented):
I think "fluid-width" would be easier to understand for someone reading code for the first time in a while

mJarsater:
> I think "fluid-width" would be easier to understand for someone reading code for the first time in a while

Agree, renamed.

alexandereneroth:
Screenshot 2023-09-05 at 14 54 52

I find it jarring that i can hover so close to the text without the menu item being clickable.

Would be nice if there was no gap between items unless there was a divider between them.

This could be achieved by putting the padding on the link rather than the <li> - or since you went with :before you could remove the :hover from your selector, so it doesn't only appear when hovering the link

alexandereneroth (changes requested):
see above comment

alexandereneroth:
Not a blocker, but something id like on the backlog if not fixed

mJarsater:
> Screenshot 2023-09-05 at 14 54 52
> 
> I find it jarring that i can hover so close to the text without the menu item being clickable.
> 
> Would be nice if there was no unclickable gap between items unless there was a divider between them.
> 
> This could be achieved by putting the padding on the link rather than the <li> or, since you went with :before pseudo element solution, you could remove the :hover from your pseudo element selector, so it exists even when not hovering the link

Agree. The popover-core styling is causing some issue so had to add an !important.

Please do also have a look at the web-component version of the popover-menu-item.

alexandereneroth (changes requested):
The hover highlight dont go all the way to the edges now. Also you can hover the lines:

Screenshot 2023-09-05 at 16 25 52

alexandereneroth:
I'd suggest removing the horizontal padding from the popover element and just adding it on the divider elements instead

alexandereneroth (approved):
Works great now

theJohnnyMe (approved):

TatyanaBumazhkina:
Tested and passed in Chrome, Safari, Firefox, Edge browsers, light and dark modes.

Checked:

font style

divider style

paddings

background, hover color

mJarsater:
Will instead implement web component: https://github.com/scania-digital-design-system/tegel/pull/280

PR #287, opened 2023-09-08
fix(utility classes): update title for gap story to display in Storybook

nathalielindqvist:
**Describe pull-request**  
Update title for Gap utility classes story to display it in Storybook.

**Solving issue**  
-

**How to test**  
1. Go to Storybook link below
2. Check in Utility Classes
3. Check that there is a story for "Gap" and that the content in "Flex" story matches what's in the code base


theJohnnyMe (approved):
Good catch and thanks for fixing it! ;)

mJarsater (approved):
Sweet!

We would need to merge develop into main after this one is merged to get these changes "published", right @theJohnnyMe ?

PR #288, opened 2023-09-08
docs(side-menu): slots comment structure update

theJohnnyMe:
**Describe pull-request**  
Update on slot comment 

**Solving issue**  
Content of title and description of the slot were previously under "Title" column. [Issue discovered in documentation PR.
](https://github.com/scania-digital-design-system/sdds-website/pull/439#issuecomment-1710182322)

**How to test**  
1. Check under the files here if the content is now separated between the Title and Description columns.

### DIFF
--- a/core/src/components/side-menu/side-menu-user-image/side-menu-user-image.tsx
+++ b/core/src/components/side-menu/side-menu-user-image/side-menu-user-image.tsx
@@ -5,9 +5,8 @@ import { Component, h, Host, Prop } from '@stencil/core';
 // FIXME: Can the logic for it be directly integrated in side-menu-user instead?
 
 /**
- * @slot <default> -
- * <b>Unnamed slot.</b> Used as alternative to props to inject <code><img...</code> element directly into the DOM.
- * */
+ * @slot <default> - <b>Unnamed slot.</b> Used as alternative to props to inject <code><img...</code> element directly into the DOM.
+ */
 
 @Component({
   tag: 'tds-side-menu-user-image',
--- a/core/src/components/side-menu/side-menu-user-image/readme.md
+++ b/core/src/components/side-menu/side-menu-user-image/readme.md
@@ -15,9 +15,9 @@
 
 ## Slots
 
-| Slot                                                                                                                            | Description |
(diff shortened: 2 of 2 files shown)


mJarsater (approved):

mJarsater (approved):

core/src/components/side-menu/side-menu-user-image/side-menu-user-image.tsx
```diff
@@ -5,9 +5,8 @@ import { Component, h, Host, Prop } from '@stencil/core';
 // FIXME: Can the logic for it be directly integrated in side-menu-user instead?
 
 /**
- * @slot <default> -
- * <b>Unnamed slot.</b> Used as alternative to props to inject <code><img...</code> element directly into the DOM.
- * */
+ * @slot <default> - <b>Unnamed slot.</b> Used as alternative to props to inject <code><img...</code> element directly into the DOM.
```
nathalielindqvist, line 8: Side Menu User Image is not present as a tab under the Notes tab in Storybook, should it be added?

alexandereneroth: I don't think they need to be. They are (internal) components of the <tds-side-menu-user>. I think its more bad than good if we document them tbh, since then we cant change them.
Perhaps they should be renamed with `core-` or something in their names? Ill add a check in topic.

alexandereneroth (approved):

core/src/components/side-menu/side-menu-user-image/side-menu-user-image.tsx
theJohnnyMe: To on what Alex wrote, as those are internal components for other component, I think it is OK not to display them. For those are curiouse, there is always part od readme which states what is component dependent on: https://github.com/scania-digital-design-system/tegel/tree/develop/packages/core/src/components/side-menu/side-menu-user

It is good idea to have different prefix so we know it is internaly used component.

theJohnnyMe:
I could not find this issue in newly cloned "core" folder under packages so I am closing it.

PR #298, opened 2023-09-13
chore: update deps minor version

alexandereneroth:
Just updating minor versions of dependencies.

Tried updating the dependencies listed in the colors package as well as prettier to 3.0 but both of those broke the build in different ways so skipped that for now.

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -54,47 +54,48 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.22.5",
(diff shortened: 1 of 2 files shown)


mJarsater:
Why isn't this showing in the package.json?

alexandereneroth:
I ran `npm update` which updates the dependencies within the bounds set in package.json. so as not to pull in breaking changes etc.

mJarsater:
Maybe try this after rebasing?

PR #315, opened 2023-09-20
fix(utils): seperate utils 

mJarsater:
**Describe pull-request**  
This PR separates all utils functions into their own files. This decreases the bundle size and should have an impact on performance. All files are located under `src/utils` with `formatHtmlPreview` being the one exception. This is instead placed under `src/stories` since it is only ever used in the `.story.tsx` files.

**Solving issue**  
Fixes: [CDEP-2643](https://tegel.atlassian.net/browse/CDEP-2643)

**How to test**  
1. Go through all components.
2. Validate that their behaviour have not changed.

[CDEP-2643]: https://tegel.atlassian.net/browse/CDEP-2643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alexandereneroth (dismissed):
Seems to work

packages/core/src/utils/appendChildElement.ts
theJohnnyMe, line 1: Seems like it is not used by any of the components. Do you see a value in keeping it?

packages/core/src/utils/getNextNestedChildOfSiblingsMatching.ts
theJohnnyMe, line 1: This one seems not to be used, but its counterpart function "getPrevNested..." seems to be in use so maybe it is good to keep it. What do you think?

packages/core/src/utils/getNextSibling.ts
theJohnnyMe, line 1: I saw this one is used only by getNextNestedChildOfSiblingsMatching function. 
Is there any value to keep directly inside the dependent one?

packages/core/src/utils/getPreviousSibling.ts
theJohnnyMe, line 1: I saw this one is used only by getPreviousNestedChildOfSiblingsMatching function. 
Is there any value to keep directly inside the dependent one?

packages/core/src/utils/isFocusable.ts
theJohnnyMe, line 1: Looks like this one is not used. Is there a value to keep it?

theJohnnyMe:
Really nice improvements. I have checked all components in Storybook and both demo pages. There are no issues.
I dropped couple of comments regarding functions that look like they are not used.

packages/core/src/utils/isFocusable.ts
mJarsater: It it's not used by a component or another utils function I'd say we can remove it

packages/core/src/utils/getPreviousSibling.ts
mJarsater: Since the treeshaking isn't great maybe we should move it into the file where it's being used instead.

packages/core/src/utils/appendChildElement.ts
mJarsater: It it's not used by a component or another utils function I'd say we can remove it

packages/core/src/utils/getNextSibling.ts
mJarsater: I think there could be a future usecase for it, so maybe keep it for consistency?

packages/core/src/utils/getPreviousSibling.ts
mJarsater: Scratch that: I think there could be a future usecase for it, so maybe keep it for consistency?

packages/core/src/utils/getPreviousNestedChildOfSiblingsMatching.ts
```diff
@@ -0,0 +1,37 @@
...
+ * //   </ul>
+ * // </div>
+ * // <div>
+ * //   <h2>Heading 2</h2>
+ * //   <ul role="list">
+ * //     <li>Item 2</li>
+ * //   </ul>
+ * // </div>
+ *
+ * const searchPredicate = (el) => el.tagName.toLowerCase() === 'h2';
```
timrombergjakobsson, line 27: Should we keep the commented out parts?

mJarsater: Im pretty sure they are part of the @example

timrombergjakobsson: Ok, yes, was not sure. But then nothing :).

packages/core/src/utils/getNextSibling.ts
theJohnnyMe: Agreed!

packages/core/src/utils/getPreviousSibling.ts
theJohnnyMe: Yep, it is OK to leave it as is and for the consistency. If change is ever needed it is no brainer to do it.

theJohnnyMe (approved):
Checked all components, Angular and React demo.
LGTM! Kudos to the work done, really great improvement!

alexandereneroth (approved):

PR #316, opened 2023-09-21
fix(dropdown): setValue method

mJarsater:
**Describe pull-request**  
The setValue method updated the `selection` of the dropdown without updating the selected state of its children. 

This PR updates the `setValue` method to enable for both value and label pair, object and array of object as it parameters. It also fixes a previous issue with double events firing on selection on multiselect. 

**Solving issue**  
Fixes: [CDEP-2656](https://tegel.atlassian.net/browse/CDEP-2656)

**How to test**  
1. Go to Dropdown
2. Update the selection using the setValue method.

[CDEP-2656]: https://tegel.atlassian.net/browse/CDEP-2656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -268,7 +268,7 @@ export class TdsDropdown {
   }
 
   private internalReset() {
-    this.children = this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
+    this.getChildren().forEach((element: HTMLTdsDropdownOptionElement) => {
       element.setSelected(false);
       return element;
     });
@@ -302,7 +302,7 @@ export class TdsDropdown {
   };
 
   selectChildrenAsSelectedBasedOnSelectionProp() {
-    this.children = this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
+    this.getChildren().forEach((element: HTMLTdsDropdownOptionElement) => {
       this.selection.forEach((selection) => {
         if (element.value !== selection.value) {
           // If not multiselect, we need to unselect all other options.
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -104,8 +104,6 @@ export class TdsDropdown {
     else nextValue = value;
 
     if (!this.multiselect && nextValue.length > 1) {
-      console.log(nextValue);
-      console.log(nextValue.length);
       console.warn('Tried to select multiple items, but multiselect is not enabled.');
       nextValue = [nextValue[0]];
     }
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -81,28 +81,56 @@ export class TdsDropdown {
     this.handleChange();
   }
 
-  /** Method for setting the value of the Dropdown. */
+  /** Method for setting the value of the Dropdown.
+   *
+   * Single selection example:
+   *
+   * <code>
+   *  dropdown.setValue('option-1', 'Option 1');
+   * </code>
+   *
+   * Multiselect example:
+   *
+   * <code>
+   *  dropdown.setValue([<br />
+   *  &nbsp;{ value: 'option-4', label: 'Option 4' },<br />
+   *  &nbsp;{ value: 'option-1', label: 'Option 1' }<br />
+   *  ]);
+   * </code>
+   */
   @Method()
-  async setValue(newValue: string, newValueLabel: string) {
-    const optionExist = this.getChildren().some(
-      (element: HTMLTdsDropdownOptionElement) => element.value === newValue,
-    );
-    // Check if any of the dropdown options has the value that is passed to the method.
-    if (optionExist) {
-      if (this.multiselect) {
-        this.selection = this.selection
-          ? [...this.selection, { value: newValue, label: newValueLabel }]
-          : [{ value: newValue, label: newValueLabel }];
-      } else {
-        this.selection = [{ value: newValue, label: newValueLabel }];
-      }
-      this.host.setAttribute(
-        'value',
-        this.selection.map((selection) => selection.value).toString(),
-      );
-      this.selectChildrenAsSelectedBasedOnSelectionProp();
-      this.handleChange();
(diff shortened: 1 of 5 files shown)


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -254,6 +257,22 @@ export class TdsDropdown {
       });
   };
 
+  selectChildrenAsSelectedBasedOnSelectionProp() {
+    this.children = this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
+      this.selection.forEach((selection) => {
+        if (element.value !== selection.value) {
+          // If it not multiselect we need to unselect all other options.
```
timrombergjakobsson, line 264:
```suggestion
          // If not multiselect, we need to unselect all other options.
```

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -88,21 +88,24 @@ export class TdsDropdown {
...
-    } else {
-      this.selection = [{ value: newValue, label: newValueLabel }];
-      this.children = this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
-        if (element.value !== newValue) {
-          element.setSelected(false);
-        }
-        return element;
-      });
+    // Check if any of the dropdown options has the value that is passed to the method.
+    if (
```
timrombergjakobsson, line 92: Maybe rewrite like this a little, for more readability?
```suggestion
    const optionExists = this.getChildren().some(
    (element: HTMLTdsDropdownOptionElement) => element.value === newValue
  );

  if (optionExists) {
    // Create a new selection based on the multiselect flag.
    const newSelection = this.multiselect
      ? [...(this.selection || []), { value: newValue, label: newValueLabel }]
      : [{ value: newValue, label: newValueLabel }];

    // Update the selection and the 'value' attribute of the host element.
    this.selection = newSelection;
    this.host.setAttribute('value', newSelection.map(selection => selection.value).toString());

    // Update the selected state of child options and trigger the change event.
    this.selectChildrenAsSelectedBasedOnSelectionProp();
    this.handleChange();
  }

  return this.selection;
```

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -254,6 +258,22 @@ export class TdsDropdown {
       });
   };
 
+  selectChildrenAsSelectedBasedOnSelectionProp() {
```
timrombergjakobsson, line 261: Don't know if better performance but maybe an enhancement in readability?
```suggestion
  selectChildrenAsSelectedBasedOnSelectionProp() {
  this.children.forEach((element: HTMLTdsDropdownOptionElement) => {
    const isSelected = this.selection.some((selection) => element.value === selection.value);
    element.setSelected(isSelected && (this.multiselect || this.selection.length === 1));
  });
}
```

mJarsater: We need to get the children first, because the children might have changed in runtime

timrombergjakobsson (approved):
💯

alexandereneroth (dismissed):
👍

timrombergjakobsson (dismissed):

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -83,27 +83,67 @@ export class TdsDropdown {
...
+    else if (!Array.isArray(value)) nextValue = [value];
+    else nextValue = value;
+
+    if (!this.multiselect && nextValue.length > 1) {
+      console.warn('Tried to select multiple items, but multiselect is not enabled.');
+      nextValue = [nextValue[0]];
     }
-    this.handleChange();
+
+    this.reset();
```
mJarsater, line 118: Since this will emit a `tdsChange` event, should we do an "inline"-reset here istead? So reset without calling the function? @alexandereneroth

alexandereneroth: Yea, I started on somehting like this, what do you think?
I think it makes sense to reuse code like this, but naming could perhaps improve.

```

  private internalReset() {
    this.children = this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
      element.setSelected(false);
      return element;
    });
    this.selection = null;
    this.host.setAttribute('value', null);
  }

  /** Method that resets the Dropdown, marks all children as non-selected and resets the value to null. */
  @Method()
  async reset() {
    this.internalReset()
    this.handleChange();
  }
  ```

Comment (alexandereneroth) [reply packages/core/src/components/dropdown/dropdown.tsx:118]: (and then use internalReset() instead of reset())

Comment (mJarsater) [reply packages/core/src/components/dropdown/dropdown.tsx:118]: Love it! Ping me when you've committed it 🚀

File: packages/core/src/components/dropdown/dropdown.tsx [hunk 5]
```diff
@@ -111,11 +106,13 @@ export class TdsDropdown {
     else nextValue = value;
 
     if (!this.multiselect && nextValue.length > 1) {
+      console.log(nextValue);
```
timrombergjakobsson, line 109: Was this unintentionally?

alexandereneroth:
Appears to work great now, but I think QA on this would be extra good considering the underlying logic is not trivial

mJarsater:
> Appears to work great now, but I think QA on this would be extra good considering the underlying logic is not trivial

Since we are interacting with a method rather that the UI component I think if we want QA on this we would have to help out with buttons that performs X,Y and Z. Personally I do not think QA is needed outside of regression test.

alexandereneroth:
> > Appears to work great now, but I think QA on this would be extra good considering the underlying logic is not trivial
> 
> Since we are interacting with a method rather that the UI component I think if we want QA on this we would have to help out with buttons that performs X,Y and Z. Personally I do not think QA is needed outside of regression test.

Yea I mostly meant regression test, but true that its hard to test the actual bug that this PR is trying to solve

alexandereneroth:
@TatyanaBumazhkina If testing this, only a regression test is needed

alexandereneroth (dismissed):

alexandereneroth (approved):

PR #326, opened 2023-09-22
fix(dropdown): use getChildren over fetching children in connectedCallback

mJarsater:
**Describe pull-request**  
The filter and tabbing functionality of the dropdown did not work until a selection had been made in the dropdown. This was due to us fetching children in the connectedCallback. Since all children are rendered after the dropdown we need to get the children before performing a filter or tabbing. This using the `getChildren()` function. 

See video:
https://github.com/scania-digital-design-system/tegel/assets/62651103/56efca81-968a-4643-b058-9d71874d970a

**Solving issue**  
Fixes: [CDEP-2691](https://tegel.atlassian.net/browse/CDEP-2691)

**How to test**  
1. Go to Dropdown
2. Try filtering before any selection have been made.
3. Try filtering after a selection have been made. 
4. Do the same for tabbing.

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

[CDEP-2691]: https://tegel.atlassian.net/browse/CDEP-2691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -72,7 +72,7 @@ export class TdsDropdown {
   /** Method that resets the Dropdown, marks all children as non-selected and resets the value to null. */
   @Method()
   async reset() {
-    this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
+    this.getChildren().forEach((element: HTMLTdsDropdownOptionElement) => {
       element.setSelected(false);
       return element;
     });
@@ -90,7 +90,7 @@ export class TdsDropdown {
         : [{ value: newValue, label: newValueLabel }];
     } else {
       this.selection = [{ value: newValue, label: newValueLabel }];
-      this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
+      this.getChildren().forEach((element: HTMLTdsDropdownOptionElement) => {
         if (element.value !== newValue) {
           element.setSelected(false);
         }
@@ -106,7 +106,7 @@ export class TdsDropdown {
   @Method()
   async removeValue(oldValue: string) {
     if (this.multiselect) {
-      this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
+      this.getChildren().forEach((element: HTMLTdsDropdownOptionElement) => {
         if (element.value === oldValue) {
           this.selection = this.selection.filter((item) => item.value !== element.value);
           element.setSelected(false);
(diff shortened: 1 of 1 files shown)


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -191,20 +191,22 @@ export class TdsDropdown {
       nextElementSibling return the index for the first child in our Dropdown.  */
 
       const startingIndex = activeElement.nextElementSibling
-        ? this.children.findIndex((element) => element === activeElement.nextElementSibling)
```
alexandereneroth, line 194: Do you know where the getChildren method is documented? 
Couldn't find any documentation on it.

Thinking it's possible it is expensive, so it might be better to run it once in the method, put the result in a variable, and use that.

mJarsater: line 255 in dropdown.tsx:
  /* Returns a list of all children that are are tds-dropdown-option elements */
  private getChildren = () =>
    Array.from(this.host.children).filter((element) => element.tagName === 'TDS-DROPDOWN-OPTION');

mJarsater: The problem with running it once it that we can't be sure all options are in the first render and so on

alexandereneroth: Aha, I thought it was a built in Stencil method.

Once per method I mean.

Ie.

```js
 async onKeyDown(event: KeyboardEvent) {
    
    // ( ... )
    
    const children = this.getChildren(); // <---- Run once in onKeyDown
     
    if (event.key === 'ArrowDown') {
      
      // ( ... )

      const startingIndex = activeElement.nextElementSibling
        ? children.findIndex((element) => element === activeElement.nextElementSibling)
        : 0;

      const elementIndex = findNextFocusableElement(children, startingIndex);
      children[elementIndex].focus();
    } else if (event.key === 'ArrowUp') {
    
      // ( ... )
      
      const startingIndex = activeElement.nextElementSibling
        ? children.findIndex(
            (element) => element === activeElement.previousElementSibling,
          )
        : 0;

      const elementIndex = findPreviousFocusableElement(children, startingIndex);
      children[elementIndex].focus();
    } else if (event.key === 'Escape') {
      this.open = false;
    }
  }
  ```

mJarsater: Makes sense! I'll update

alexandereneroth (dismissed):
Works great

timrombergjakobsson (approved):

alexandereneroth (approved):

TatyanaBumazhkina:
Tested and passed in Chrome, Safari, Edge, Firefox.
Tested with light/dark modes, large/medium/small size.

User is able to filter item by clicking on dropdown field or tabbing.

PR #331, opened 2023-09-26
refactor(dropdown): cleanup and enhancement setValue & removeValue

mJarsater:
**Describe pull-request**  
This PR builds further on the `setValue()` and `removeValue()` methods. This removes the dependency on `label` as an attribute to the methods. 

The user should still be able to interact with the `setValue` method by passing in two attributes (value, label). However the label passed in is not used, this is only there to not break the API. The user can now also pass in a string array that should have one entry per value they want selected.

**Solving issue**  
Fixes: -

**How to test:**
1. Checkout branch.
2. Try updating the dropdown using the `setValue` method.
3. Test it for default, filter and multiselect dropdown
4. Verify that the dropdown works as before

### DIFF
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -312,7 +312,7 @@ export class TdsDropdown {
   };
 
   selectChildrenAsSelectedBasedOnSelectionProp() {
-    this.getChildren().map((element: HTMLTdsDropdownOptionElement) => {
+    this.getChildren().forEach((element: HTMLTdsDropdownOptionElement) => {
       this.value.forEach((selection) => {
         if (element.value !== selection) {
           // If not multiselect, we need to unselect all other options.
@@ -323,7 +323,6 @@ export class TdsDropdown {
           element.setSelected(true);
         }
       });
-      return element;
     });
   }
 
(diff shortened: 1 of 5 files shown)


mJarsater:
@TatyanaBumazhkina this only need regression testing of the Dropdown

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -156,7 +160,9 @@ export class TdsDropdown {
       this.reset();
     }
     this.handleChange();
-    return this.selection;
+    /* This returns an array of object with a value and label pair. This is ONLY to not break the API. Should be removed for 2.0. */
+    // TODO - Clean up and just return this.value for 2.0
```
timrombergjakobsson, line 164: Should we have this in a ticket also? Maybe good rule to only introduce a TODO together with a Jira ticket? Or use some kind of linting for it, https://github.com/maxprilutskiy/eslint-plugin-no-expired-todo-comments, this one checks for expiry date of TODO?

mJarsater: I can add it as a ticket also maybe? Feel like it could be good to have it here also to not forget it, feels like things gets lost in the backlog

nathalielindqvist (approved):
New setValue() method works as expected for default, multiselect and filter, great work! Non-related bug in filter version reported in Jira.

timrombergjakobsson (approved):

PR #341, opened 2023-09-28
fix(css reset/components): added important for slotted(*) targeted elements font styles

mJarsater:
**Describe pull-request**  
Our CSS-reset has a rule:
```css
input,
button,
textarea,
select {
  font: inherit;
}
```
Which was causing issues for the Tabs. There selector (slotted(*)) for targeting slotted buttons was not strong enough and was getting overwritten by the rule above. This resulted in this look:
Screenshot 2023-09-28 at 13 55 39

When it should look like this:
Screenshot 2023-09-28 at 13 56 50

To combat this. I have used `!important` for the `font` and `letter-spacing` in our CSS whenever we are targeting `::slotted(*)`

**Solving issue**  
Fixes: [CDEP-2708](https://tegel.atlassian.net/browse/CDEP-2708)

**How to test**  

To see the error: 
1. Checkout the `develop` branch.
2. In `preview.css` add our CSS-reset
3. Check in Tabs and verify that the text have the wrong styling.

1. Checkout this branch.
3. In `preview.css` add our CSS-reset
5. Check in Tabs that the look of the tabs are intact.

[CDEP-2708]: https://tegel.atlassian.net/browse/CDEP-2708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/card/card.scss
+++ b/packages/core/src/components/card/card.scss
@@ -77,6 +77,7 @@
   }
 
   slot[name='body']::slotted(*) {
+    /* !important is needed here to prevent this from being overwritten by our CSS-reset. */
     font: var(--tds-detail-03) !important;
     letter-spacing: var(--tds-detail-03-ls) !important;
     color: var(--tds-card-body-color);
--- a/packages/core/src/components/card/card.scss
+++ b/packages/core/src/components/card/card.scss
@@ -77,8 +77,8 @@
   }
 
   slot[name='body']::slotted(*) {
-    font: var(--tds-detail-03);
-    letter-spacing: var(--tds-detail-03-ls);
+    font: var(--tds-detail-03) !important;
+    letter-spacing: var(--tds-detail-03-ls) !important;
     color: var(--tds-card-body-color);
     padding: 0 16px;
     margin-bottom: 16px;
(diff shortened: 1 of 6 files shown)


alexandereneroth (approved):

packages/core/src/components/card/card.scss
```diff
@@ -77,8 +77,8 @@
   }
 
   slot[name='body']::slotted(*) {
-    font: var(--tds-detail-03);
-    letter-spacing: var(--tds-detail-03-ls);
+    font: var(--tds-detail-03) !important;
```
timrombergjakobsson, line 80: I feel that using `!important` is a bit scary. But if it is the only way to fix this, then maybe have a comment about why doing it like this? Or potentially, but not part of this pr, look into the css reset we have and see if it can be replaced with something else that does not require the usage of `!important`

mJarsater: Agree, the `!important` seems to be necessary due to the weakness of the `::slotted` pseudo selector. But I will add comments like you suggested 🚀

timrombergjakobsson (approved):

PR #346, opened 2023-10-02
fix: add ESLint rule to prevent importing from Stencils internal package

lkritsimas:
**Describe pull-request**
Some components are currently importing from Stencil's internal package. However, Stencil warns to never do that.
> NOTE!! The @stencil/core/internal package is not meant to be consumed directly by anything other than Stencil internals. It is its own package so that it can be resolved by Stencil, but breaking changes can/will happen at any time. This isn't a "use at your own discretion" moment, but it's more of a "never use this because your code will break" fact.
https://github.com/ionic-team/stencil/blob/main/src/internal/readme.md

Added an ESLint rule to prevent importing from the internals again in the future and changed the imports affected by the rule.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -66,6 +66,12 @@
     "no-restricted-syntax": "warn", // TODO: check this - disables for-in related error
     "default-case-last": "warn",
     "no-restricted-exports": "warn",
+    "no-restricted-imports": ["error",
+      {
+        "name": "@stencil/core/internal",
+        "message": "Use @stencil/core instead."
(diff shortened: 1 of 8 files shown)


timrombergjakobsson:
Wow! Nice catch! @lkritsimas

mJarsater (approved):
Awsome sauce 🥫

timrombergjakobsson (approved):

alexandereneroth (approved):
Great catch. Thanks for your contribution!

PR #353, opened 2023-10-04
fix(modal): prevent text selection when toggling modal

lkritsimas:
**Describe pull-request**  
This fixes an issue where text in the modal was selected when quickly toggling its visibility, or double clicking on the backdrop of a modal that had the `prevent` prop set to `true`.

**How to test** 
1. Spam click on the "Open Modal" button in Storybook.
2. Double click anywhere on the backdrop of a modal that has `prevent` set to `true`.

Modal content should still be selectable and interactable. It should still be possible to close the modal by clicking on the modals backdrop when `prevent` is set to `false`.

**Screenshots**

Before:
modal-before

After:
modal-after

### DIFF
--- a/packages/core/src/components/modal/modal.scss
+++ b/packages/core/src/components/modal/modal.scss
@@ -90,29 +90,6 @@ $modals: (
   }
 }
 
-@mixin modal-host {
-  left: 0;
-  top: 0;
-  position: fixed;
-  height: 100%;
-  width: 100%;
-  z-index: tds-z-index(modal);
-  background-color: var(--tds-modal-backdrop);
-  padding: 0 16px;
-
-  @media (max-width: $screen-s) {
-    padding: 0;
-  }
-
-  &.show {
-    display: flex;
-  }
-
-  &.hide {
-    display: none;
-  }
-}
-
 @mixin modal-header {
   display: flex;
   padding-bottom: var(--tds-spacing-element-8);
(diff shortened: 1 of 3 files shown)


theJohnnyMe (approved):
Nice finding and fix, thank you @lkritsimas !

timrombergjakobsson (approved):
Nice catch!

packages/core/src/components/modal/modal.scss
```diff
@@ -90,29 +90,6 @@ $modals: (
   }
```
mJarsater, line 90: Could we add a `prevent` control to the story for the Modal, for us to be able to test it in storybook.

mJarsater (changes requested):
Could we add a `prevent` control (false by default) to the story for the Modal, for us to be able to test it in storybook.

lkritsimas:
> Could we add a `prevent` control (false by default) to the story for the Modal, for us to be able to test it in storybook.

Done!

By the way, I think the `prevent` prop could be renamed to `persist` or `persistent` to be a bit more self-explanatory. Maybe something to consider for the next major release 😄

mJarsater:
> > Could we add a `prevent` control (false by default) to the story for the Modal, for us to be able to test it in storybook.
> 
> Done!
> 
> By the way, I think the `prevent` prop could be renamed to `persist` or `persistent` to be a bit more self-explanatory. Maybe something to consider for the next major release 😄

Sweet! 🍬

I totally agree with the prop name. It would be great to align it to the same usage in other components. Like you mentioned though, it would have to be something for 2.0. I will make sure to add it to our backlog so we revisit it in the future

mJarsater (approved):
🔥

PR #357, opened 2023-10-09
fix(stepper): change internal event to not be documented in readme file

nathalielindqvist:
**Describe pull-request**  
Added @internal decorator to 'internalTdsPropsChange' event so it's not documented in the readMe.

**Solving issue**  
Fixes: [CDEP-2749](https://tegel.atlassian.net/browse/CDEP-2749)

**How to test**  
1. Go to Storybook link below
2. Check in Stepper -> Notes
3. Check that the event 'internalTdsPropsChange' is not documented

[CDEP-2749]: https://tegel.atlassian.net/browse/CDEP-2749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/stepper/stepper.tsx
+++ b/packages/core/src/components/stepper/stepper.tsx
@@ -51,6 +51,7 @@ export class TdsStepper {
     }
   }
 
+  /** @internal Broadcasts changes to the Stepper props */
   @Event({
     eventName: 'internalTdsPropsChange',
     composed: true,
--- a/packages/core/src/components/stepper/stepper.tsx
+++ b/packages/core/src/components/stepper/stepper.tsx
@@ -51,7 +51,7 @@ export class TdsStepper {
     }
   }
 
-  /** @internal Broadcasts changes to the Stepper props */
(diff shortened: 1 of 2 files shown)


packages/core/src/components/stepper/stepper.tsx
```diff
@@ -51,6 +51,7 @@ export class TdsStepper {
     }
   }
 
+  /** @internal Broadcasts changes to the Stepper props */
```
mJarsater, line 54: I would change this to something like: Broadcasts changes in props to the stepper children.

mJarsater (dismissed):

packages/core/src/components/stepper/stepper.tsx
nathalielindqvist: Nice, updated!

theJohnnyMe (approved):
Nice!

alexandereneroth (approved):
Great catch

PR #361, opened 2023-10-09
fix(dropdown): fix multiselect & filter getting value "undefined"

alexandereneroth:
**Describe pull-request**  
Fixes dropdown getting value "undefined" (as a string) when it was a multiselect, filter, and opened and closed without selecting a value.

> "The reason inputElement.value is "undefined" as a string is because when you assign undefined to an input element's value, it gets converted to the string "undefined". This is a standard behavior in JavaScript: when undefined is coerced into a string, it becomes "undefined"."
-- GPT-4

**Solving issue**  
Fixes: [CDEP-2749](https://tegel.atlassian.net/browse/CDEP-2759)

**How to test**  
1. Go to deploy preview
2. Check in dropdown, add multiselect & filter
3. Open and close the dropdown and verify it looks good.

**Screenshots**  

Screenshot 2023-10-09 at 16 02 15

[CDEP-2749]: https://tegel.atlassian.net/browse/CDEP-2749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -369,6 +369,9 @@ export class TdsDropdown {
 
   private getValue = () => {
     const labels = this.getSelectedChildrenLabels();
+    if (!labels) {
+      return '';
+    }
     return this.filter ? labels?.join(', ') : labels?.toString();
   };
 
@@ -462,6 +465,7 @@ export class TdsDropdown {
                   disabled={this.disabled}
                   onInput={(event) => this.handleFilter(event)}
                   onBlur={(event) => {
+                    console.log('blurrr');
                     this.filterFocus = false;
                     this.handleBlur(event);
                   }}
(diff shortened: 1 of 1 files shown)


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -369,6 +369,9 @@ export class TdsDropdown {
 
   private getValue = () => {
     const labels = this.getSelectedChildrenLabels();
+    if (!labels) {
+      return '';
```
mJarsater, line 373: Any reason for returning a blank string over `null`?

mJarsater (approved):
🧨

packages/core/src/components/dropdown/dropdown.tsx
timrombergjakobsson: Is it not more appropiate to return empty string here, as the function is designed to return a string representation of labels?

alexandereneroth: A text inputs value is always a string, so i think returning an empty string is ok. Null might work as well, but it converted undefined to "undefined" so 🤷‍♂️

PR #363, opened 2023-10-10
build(core/react/angular): added release actions for all packages

mJarsater:
**Describe pull-request**  
This PR adds three github actions:
 - release-core
 - release-react
 - release-angular
 
 This are all actions that are triggered manually and they do a release to npm for each package.
 
 The `release-core` actions requires the following input:
  - Release type (major, minor, patch)
  - Node version (18.16.0 by default) -> this specifies which node package should be built with.
  - Tag -> specifies what npm tag the release should have
  - Dry run -> if true the `npm publish` will run with a dry-run flag

The `release-react` and `release-angular` requires the following input:
  - Tegel version -> which tegel version is the dependency for the package. This will also be the version of the built an published package. For example: if Tegel version is 1.1.0, the tegel-angular package will be built as @scania/tegel@1.1.0 as a dependency **and** released @scania/tegel-angular package will be of version 1.1.0
  - Node version (18.16.0 by default) -> this specifies which node package should be built with.
  - Tag -> specifies what npm tag the release should have
  - Dry run -> if true the `npm publish` will run with a dry-run flag

Example of ran actions: https://github.com/scania-digital-design-system/tegel-react-demo/actions
_These were all run with dry run: true_
 
**Solving issue**  
Fixes:
 - [CDEP-2754](https://tegel.atlassian.net/browse/CDEP-2754)
 - [CDEP-2755](https://tegel.atlassian.net/browse/CDEP-2755)
 - [CDEP-2756](https://tegel.atlassian.net/browse/CDEP-2756)

**How to test**  
Not testable today.

[CDEP-2754]: https://tegel.atlassian.net/browse/CDEP-2754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release-angular.yml
+++ b/.github/workflows/release-angular.yml
@@ -44,10 +44,10 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Angular - CD into correct folder
-        run: echo "cd packages/angular"
+        run: cd packages/angular
 
       - name: Angular - Remove package-lock
-        run: echo "rm -rf package-lock.json"
+        run: rm -rf package-lock.json
 
       - name: Angular - Install latest tegel package
         run: npm install @scania/tegel@${{ github.event.inputs.tegelVersion }}
--- a/.github/workflows/release-angular.yml
+++ b/.github/workflows/release-angular.yml
@@ -1,0 +1,87 @@
+name: Release @scania/tegel-angular
+on:
+  workflow_dispatch:
+    inputs:
+      tegelVersion:
+        description: 'Tegel version'
+        required: true
+        type: string
+
+      nodeVersion:
+        description: 'Node version'
+        required: true
+        default: '18.16.0'
+        type: string
+
+      tags:
+        description: 'Tag'
+        required: true
+        default: 'beta'
+        type: choice
+        options:
+          - latest
+          - beta
+          - dev
+
+      dryRun:
+        description: 'Dry run'
+        required: false
+        default: true
+        type: boolean
+
+jobs:
+  Release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Set up node
+        # Setup .npmrc file to publish to npm
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ github.event.inputs.tegelVersion }}
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Angular - CD into correct folder
+        run: echo "cd packages/angular"
+
+      - name: Angular - Remove package-lock
+        run: echo "rm -rf package-lock.json"
+
+      - name: Angular - Install latest tegel package
+        run: npm install @scania/tegel@${{ github.event.inputs.tegelVersion }}
+
+      - name: Set user
+        run: git config --global user.name "Tegel - Scania" && git config --global user.email "tegel.design.system@gmail.com"
+
+      - name: Angular - Bump version
+        run: npm version ${{ github.event.inputs.tegelVersion }}
+
+      - name: Angular - Read Package.json Version
+        id: version
+        run: echo "version=$(jq -r .version package.json)" >> $GITHUB_OUTPUT
+
+      - name: Check out new branch
+        run: git checkout -b "release/${{ steps.version.outputs.version }}"
+
+      - name: Angular - Run build
+        run: npm run build
+
+      - name: Angular - Publish
+        run: |
+          if [ "${{ github.event.inputs.dryRun }}" == "true" ]; then
+            npm publish --tag ${{ github.event.inputs.tags }} --dry-run
+          else
+            npm publish --tag ${{ github.event.inputs.tags }}
+          fi
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Core - Create PR
+        uses: peter-evans/create-pull-request@v3
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-core.yml
```diff
@@ -0,0 +1,92 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Set up node
+        # Setup .npmrc file to publish to npm
+        uses: actions/setup-node@v3
+        with:
+          node-version: '18.16.0'
```
alexandereneroth, line 48: I think you intended to take this version from the input

alexandereneroth:
I can suggest looking into composite actions to share code between workflows:
https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

alexandereneroth:
> I can suggest looking into composite actions to share code between workflows: https://docs.github.com/en/actions/creating-actions/creating-a-composite-action

Something like this

```
# .github/actions/setup-project/action.yml
name: 'Setup project'
description: 'Checks out the code and sets up Node.js'

inputs:
  node-version:
    description: 'Node.js version'
    required: true
    default: '18.16.0'

runs:
  using: 'composite'
  steps:
    - name: Check out code
      uses: actions/checkout@v2

    - name: Set up node
      uses: actions/setup-node@v3
      with:
        node-version: ${{ inputs.node-version }}
        registry-url: 'https://registry.npmjs.org'
```

and then 

```
# .github/workflows/release-react.yml
jobs:
  Release:
    runs-on: ubuntu-latest
    steps:
      - name: Setup project
        uses: ./.github/actions/setup-project
        with:
          node-version: ${{ github.event.inputs.tegelVersion }}
      # other steps...
```

.github/workflows/release-angular.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Angular - Read Package.json Version
+        id: version
+        run: echo "version=$(jq -r .version package.json)" >> $GITHUB_OUTPUT
+
+      - name: Check out new branch
+        run: git checkout -b "release/${{ steps.version.outputs.version }}"
+
+      - name: Angular - Run build
+        run: npm run build
```
alexandereneroth, line 69: I think it would be better if the build step was before the "checkout and create branch" step, so no branch is created in case the build fails.
Might not be an issue, but also might be

alexandereneroth (commented):
Great work with this 👍

.github/workflows/release-core.yml
mJarsater: Yes!

.github/workflows/release-angular.yml
mJarsater: True!

.github/workflows/release-core.yml
alexandereneroth, line 48:
```suggestion
          node-version: ${{ github.event.inputs.nodeVersion }}
```

alexandereneroth:
I think it looks good, probably next step would be to test it for real (with dry run and then perhaps a 0.0.0-dev beta version or something)

alexandereneroth:
If you feel like looking into something you could try and find out how you could roll back(eg. delete) the tag and branch in case of any errors

.github/workflows/release-core.yml
```diff
@@ -0,0 +1,107 @@
...
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Core - Create PR
+        uses: peter-evans/create-pull-request@v3
+        with:
+          title: release/@scania/tegel@${{ steps.version.outputs.version }}
+          commit-message: Release of @scania/tegel@${{ steps.version.outputs.version }}
+          branch: release/${{ steps.version.outputs.version }}
+          delete-branch: true
+  on-failure:
```
mJarsater, line 93: Something like this could work to remove the branch and git tag on failure @alexandereneroth

.github/workflows/release-core.yml
```diff
@@ -0,0 +1,107 @@
...
+    steps:
+      - name: Notify on Failure
+        env:
+          npm_version: ${{ needs.release.outputs.version }}
+        run: |
+          if [ "${{ job.status }}" == "failure" ]; then
+            tagname="@scania/tegel@$npm_version"
+            git push --delete origin $tagname
+            branchname="release/@scania/tegel@$npm_version"
+            git push origin --delete branch_name
```
mJarsater, line 106:
```suggestion
            git push origin --delete branchname
```

.github/workflows/release-core.yml
alexandereneroth: Nice, will that run automatically?

mJarsater: Yes, it **should** run when there is a failure anywhere in the other step

alexandereneroth:
How would you say we move forward with this? Maybe we could schedule a call and make a dummy beta release to verify that everything works for real.

mJarsater:
> How would you say we move forward with this? Maybe we could schedule a call (I'd like to see) and make a dummy beta release to verify that everything works for real.

Sounds like a good plan. Please set up a meeting and invite everyone interested.

theJohnnyMe (approved):
Logic makes sense and I know you have tested some of it on the demo page.
As you said, it is hard to test it but have it merged and try it dry flad and later with prod test release. 
Great work Martin!

alexandereneroth (approved):

PR #375, opened 2023-10-23
feat(date-picker): inital date-picker development

mJarsater:
Date Picker component.

**To-do**
- [ ] Review and align design
- [ ] UX questions: What happens when I click a date in the next month, should the dates be update to reflect the selected datemonth?
- [ ] Locales - date fns have support for locales, we need to see which are available and expose the available options as a prop on the component.
- [ ] Can the start/selected date be null?
- [ ] Make sure all available props in Date Picker is available in Date Range Picker
- [ ] Enable slots? 
- [ ] What data should be submitted in a form? 
- [ ] The focus state of the next/previous buttons?
- [ ] When should the Date Picker close? Only on click outside? What if you click the Date Picker and then click the input field.
- [ ] The Date Picker needs further control of the popover, it should open when you click the input field. But it should not close when the input field is clicked, if the Date Picker is opened. I do not think this is possible with the implementation of the popover we have today. 
- [x] Error state - prop to set the error state to true and slot to allow for users to pass in a error message.
- [x] Success state - prop to set the error state to true and slot to allow for users to pass in a success message.
- [x] Label & label position props
- [x] Min/Max props
- [x] Disable same date selection
- [x] Something buggy with choosing a date a couple of months ahead

**Solving issue**  _
Fixes: -

**How to test**  
1. Go to Date Picker
2. Check it out
3. Go to Date Range Picker
4. Check it out


mJarsater:
Don't let him die please.. 😢

alfredskeden:
> Don't let him die please.. 😢

I don't think they are letting it die, It's just a lot going in this PI and acute maintenance does not include this one for now :I

mJarsater:
> I am thinking of merging this code as is, but excluding it from the prod Storybook and excluding it in shipping in NPM, but keeping it in the dev Storybook under a separate "exploratory" section.

I think the only thing is to keep it up-to-date with rebasing. It doesn't alter anything in existing component so there shouldn't be any conflicts there. 

Do we have a way of merging it to develop but keeping out of the prod build?

theJohnnyMe:
> > I am thinking of merging this code as is, but excluding it from the prod Storybook and excluding it in shipping in NPM, but keeping it in the dev Storybook under a separate "exploratory" section.
> 
> I think the only thing is to keep it up-to-date with rebasing. It doesn't alter anything in existing components so there shouldn't be any conflicts there.
> 
> Do we have a way of merging it to develop but keeping out of the prod build?
> 
> Edit: rebased it now btw.

I think files can be ignored in the .npmignore file so they are not shipped when doing a release and considering prod Storybook, we do not have to create a story for it so it is not visible, or we can use an environment variable to show only in our local environments.

All of these steps are there to keep the code in until we have time to focus on it more. 

What do you think?

mJarsater:
> > > I am thinking of merging this code as is, but excluding it from the prod Storybook and excluding it in shipping in NPM, but keeping it in the dev Storybook under a separate "exploratory" section.
> > 
> > 
> > I think the only thing is to keep it up-to-date with rebasing. It doesn't alter anything in existing components so there shouldn't be any conflicts there.
> > Do we have a way of merging it to develop but keeping out of the prod build?
> > Edit: rebased it now btw.
> 
> I think files can be ignored in the .npmignore file so they are not shipped when doing a release and considering prod Storybook, we do not have to create a story for it so it is not visible, or we can use an environment variable to show only in our local environments.
> 
> All of these steps are there to keep the code in until we have time to focus on it more.
> 
> What do you think?

Yeah I think it would just have to be tested and verified first. So we don't accidentally include it in a prod build and push it as part of the package. Hiding it in storybook could be just removing the stories for now

packages/core/package.json
```diff
@@ -50,6 +50,8 @@
   "dependencies": {
     "@popperjs/core": "^2.11.8",
     "@stencil/core": "^4.9.0",
+    "date-fns": "^2.30.0",
+    "dotenv": "^16.0.3",
```
theJohnnyMe, line 54: We dropped "dotenv" a couple of releases ago due to the new Node version supporting it, do we need it for something here?

mJarsater: Nope, not needed. I'll update

theJohnnyMe:
OK, so what I did:

- moved date picker folders into new "_beta" folder
- in each of date picker stories there is title which is dependent if STORYBOOK_ENV is equal to "dev", if is, story has a title and shows in menu, otherwise not
- there are two stencil configs due to two tsconfigs

Why two TSConfigs
- One for production
- One for development
In development one we include all files from components, while in production one we exclude "_beta" folder.

How to check it?
- On mac run "export STORYBOOK_ENV="dev""
- From same terminal, run npm run start from root level
- In Storybook, Beta stories should be available
- Digging in "dist" folder of core, you should be able to ee date-picker files
- stop environment
- run npm pack to see file size, should be around 5.1MB
- run "unset STORYBOOK_ENV" in terminal
- run npm run build
- check if "dist" folder of core and see if you find any date-picker files
- run npm pack to see file size, should be around 4MB

Please, consider I did this on late Friday afternoon, it might be I missed something but it was nice to play with the configuration...

mJarsater:
@theJohnnyMe I'm getting some weird `null` story:
Screenshot 2024-01-18 at 15 18 51

theJohnnyMe:
> @theJohnnyMe I'm getting some weird `null` story: Screenshot 2024-01-18 at 15 18 51

That sucks. I will try to find another way then. Sorry for waiting on this one so long, I simply do not have time for everything right now, a bit tuff times :(

theJohnnyMe:
@mJarsater  - I did an update on config and stencil.config, now it should hide stories from showing up in case it is not "dev" environment. Same goes for building them and shipping them. Please check and let me know what do you think.

theJohnnyMe:
@mJarsater  - Please check the last 2 commits. I added improvements to the config so now it should leave Datepicker out while in production. Once you set STORYBOOK_ENV='dev' and run build again, you should be able to see it in Storybook.

mJarsater:
> @mJarsater - Please check the last 2 commits. I added improvements to the config so now it should leave Datepicker out while in production. Once you set STORYBOOK_ENV='dev' and run build again, you should be able to see it in Storybook.

My local setup isn't great. But if SB looks alright and you can pack and make sure it is not part of the bundle for all packages I think it's good to go

nathalielindqvist (approved):
For a beta implementation I think this is a good start! Really nice config setup 👏

theJohnnyMe (approved):

PR #376, opened 2023-10-23
fix(tabs): unset z index

alexandereneroth:
z index were unnecessarily high.

NOTE: Might technically be a breaking change. However, I believe users getting negatively affected would be very rare. As such, I think leaving it will cause far more issues for users than removing it (z-index).

**How to test:**
Open the deploy preview and check that:
- tabs have no z-index set
- scroll buttons still appear and work as expected
- check for other issues

### DIFF
--- a/packages/core/src/components/tabs/inline-tabs/inline-tabs.scss
+++ b/packages/core/src/components/tabs/inline-tabs/inline-tabs.scss
@@ -1,6 +1,5 @@
 @import '../../../mixins/focus-state';
 @import '../../../mixins/box-sizing';
-@import '../../../mixins/z-index';
 
 :host {
   @include tds-box-sizing;
@@ -35,12 +34,12 @@
 
   .scroll-right-button {
     right: 0;
-    z-index: calc(tds-z-index(tab) + 1);
+    z-index: 1;
   }
 
   .scroll-left-button {
     left: 0;
-    z-index: calc(tds-z-index(tab) + 1);
+    z-index: 1;
   }
 
   .scroll-right-button,
--- a/packages/core/src/components/tabs/folder-tabs/folder-tab/folder-tab.scss
+++ b/packages/core/src/components/tabs/folder-tabs/folder-tab/folder-tab.scss
@@ -1,5 +1,4 @@
 @import '../../../../mixins/focus-state';
-@import '../../../../mixins/z-index';
 @import '../../../../mixins/box-sizing';
 
 :host {
--- a/packages/core/src/components/tabs/folder-tabs/folder-tabs.scss
+++ b/packages/core/src/components/tabs/folder-tabs/folder-tabs.scss
@@ -1,6 +1,5 @@
 @import '../../../mixins/focus-state';
 @import '../../../mixins/box-sizing';
-@import '../../../mixins/z-index';
 
 :host {
   @include tds-box-sizing;
@@ -20,12 +19,12 @@
   }
 
   .scroll-right-button {
-    z-index: calc(tds-z-index(tab) + 1);
+    z-index: 1;
     right: 0;
   }
 
   .scroll-left-button {
-    z-index: calc(tds-z-index(tab) + 1);
+    z-index: 1;
     left: 0;
   }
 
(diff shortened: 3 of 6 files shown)


timrombergjakobsson:
Did we discuss this one in a checkin?

alexandereneroth:
@mJarsater What was your reasoning for keeping z-index on tabs? I remember we discussed this.
I think they could be removed. Technically it might be a breaking change - but I believe the benefit of changing it outweighing the drawback of leaving it.

mJarsater:
> as your reasoning for keeping z-index on tabs? I remember we discussed this.
> I think they could be removed. Technically it might be a breaking change - but I believe the benefit of changing it outweighing the drawback of leaving it

I think the only reason was to be sure that the previous/next buttons are showing correctly.

alexandereneroth:
> > as your reasoning for keeping z-index on tabs? I remember we discussed this.
> > I think they could be removed. Technically it might be a breaking change - but I believe the benefit of changing it outweighing the drawback of leaving it
> 
> I think the only reason was to be sure that the previous/next buttons are showing correctly.

Alright, then I think making sure of that can be part of testing this PR.

alexandereneroth:
> https://pr-376.d3fazya28914g3.amplifyapp.com

Don't think so, anyway I think it would not be worth the time

alexandereneroth:
> Did we discuss this one in a checkin?

Don't think so, anyway I think it would not be worth the time

mJarsater (approved):

TatyanaBumazhkina:
Checked deploy preview page for Inline Tabs and Navigation Tabs. Elements don't have z-index in HTML code.

packages/core/src/components/tabs/inline-tabs/inline-tabs.scss
```diff
@@ -35,12 +34,12 @@
 
   .scroll-right-button {
     right: 0;
-    z-index: calc(tds-z-index(tab) + 1);
+    z-index: 1;
```
nathalielindqvist, line 37: Asking because I don't know - will the set z-index not lead to the scroll buttons ending up behind something else if the user has something in the surrounding context that has a higher z-index?

timrombergjakobsson:
> > Did we discuss this one in a checkin?
> 
> Don't think so, anyway I think it would not be worth the time

?

nathalielindqvist (approved):
Looks good! Just a small comment regarding context 😄

theJohnnyMe (approved):
Checked in Storybook, looks OK. I would say please test this in one of demo pages too by packing the code, just to see how it interacts with other components.

alexandereneroth:
> > > Did we discuss this one in a checkin?
> > 
> > 
> > Don't think so, anyway I think it would not be worth the time
> 
> ?

Sorry, I misread. Thought you wrote "Should we discuss this one in a checkin?"

I think I might have brought it up, can't quite remember.

alexandereneroth:
> Checked in Storybook, looks OK. I would say please test this in one of demo pages too by packing the code, just to see how it interacts with other components.

I suggest we can write a Playwright test for it instead, what do you think?
Packing it for the demo pages would be a bit of work now as they use the wrappers.

timrombergjakobsson:
> > Checked in Storybook, looks OK. I would say please test this in one of demo pages too by packing the code, just to see how it interacts with other components.
> 
> I suggest we can write a Playwright test for it instead, what do you think? Packing it for the demo pages would be a bit of work now as they use the wrappers.

Sounds like a good idea! On the demo pages note though, we can still pack it right? I tried earlier, and it worked.

alexandereneroth:
> > > Checked in Storybook, looks OK. I would say please test this in one of demo pages too by packing the code, just to see how it interacts with other components.
> > 
> > 
> > I suggest we can write a Playwright test for it instead, what do you think? Packing it for the demo pages would be a bit of work now as they use the wrappers.
> 
> Sounds like a good idea! On the demo pages note though, we can still pack it right? I tried earlier, and it worked.

I guess. I feel quite confident with this one though. Leave a review comment or reject the PR if you think it should be tested like that before merging. 
That green squash and merge button looks awfully tempting.

timrombergjakobsson:
> > > > Checked in Storybook, looks OK. I would say please test this in one of demo pages too by packing the code, just to see how it interacts with other components.
> > > 
> > > 
> > > I suggest we can write a Playwright test for it instead, what do you think? Packing it for the demo pages would be a bit of work now as they use the wrappers.
> > 
> > 
> > Sounds like a good idea! On the demo pages note though, we can still pack it right? I tried earlier, and it worked.
> 
> I guess. I feel quite confident with this one though. Leave a review comment or reject the PR if you think it should be tested like that before merging. That green squash and merge button looks awfully tempting.

You already have 3 approved reviews, so go ahead ;)

PR #378, opened 2023-10-24
docs(datetime): updated the max prop to use correct prop

mJarsater:
**Describe pull-request**  
The max control in storybook was used to set the min prop. This PR fixes that.

**How to test**  
1. Go to Datetime
2. Set a min/max value
3. Make sure the props work properly.


alexandereneroth (approved):
⚡

alfredskeden (approved):
letsgo 👍

PR #380, opened 2023-10-24
fix(plausible): domain update

theJohnnyMe:
**Describe pull-request**  
Updating Plausible domain for **develop branch**

**Solving issue**  
Fixes: [CDEP-2890](https://tegel.atlassian.net/browse/CDEP-2890)

**How to test**  
1. Check if the domain in code changes is updated to new one (tds-storybook.tegel.scania.com)

[CDEP-2890]: https://tegel.atlassian.net/browse/CDEP-2890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/.storybook/manager-head.html
+++ b/packages/core/.storybook/manager-head.html
@@ -1,6 +1,6 @@
 <!-- .storybook/manager-head.html -->
 
-<script defer data-domain="tegel-storybook.netlify.app" src="https://plausible.io/js/script.manual.js"></script>
+<script defer data-domain="tds-storybook.tegel.scania.com" src="https://plausible.io/js/script.manual.js"></script>
 
 <!-- 1. define the `plausible` function to manually trigger events -->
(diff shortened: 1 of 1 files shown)


alexandereneroth (commented):
I suggest two PRs as you did, but with the same branch, otherwise git will still consider them out of sync, even if they contain the same changes.
We also might have to remove the branch auto deletion on merge.

theJohnnyMe:
> I suggest two PRs as you did, but with the same branch, otherwise git will still consider them out of sync, even if they contain the same changes. We also might have to remove the branch auto deletion on merge.

Good that you commented. I would propose this:

1. Close this PR
2. Merge changes to the main branch so we get tracking in the place (in other PR)
3. Rebase develop with main so develop has whatever main has + its new changes
4. Live happily ever after

@alexandereneroth  - what do you think of this approach?
I think I can do it as I can bypass GitHub rules.

PR #382, opened 2023-10-25
docs(datetime): correct the setting of max prop

mJarsater:
**Describe pull-request**  
Updated the setting of max prop in the story.

**How to test**  
1. Go to Datetime
2. Test and validate the max prop


packages/core/src/components/datetime/datetime.stories.tsx
```diff
@@ -202,14 +202,14 @@ const datetimeTemplate = ({
     </style>
 
   <div class="demo-wrapper">
-
+        hahah
```
theJohnnyMe, line 205:
```suggestion
        
```

theJohnnyMe (approved):

PR #390, opened 2023-10-25
build(release actions): cleanup of release gh actions

mJarsater:
This PR cleans up the github actions used for the releases of @scania/tegel, @scania/tegel-react and @scania/tegel-angular.

For @scania/tegel-react and @scania/tegel-angular, it adds a installation and build step for the core package. This is to be able to generate the wrappers based off of the core package. It also adds the package type (tegel, tegel-react, tegel-angular) to the PR branch it creates when a release is completed.

It also moves the NPM version bump into a seperate action that then creates a PR againt main. 

The release flow would look like this: 

1. Run the `Bump NPM version (core/angular/react)` action
2. Review and merge the PR it created into main
3. Review and merge the same branch into develop
4. Run: `Release @scania/tegel`, `Release @scania/tegel-react` and `Release @scania/tegel-angular`
5. Party 🥳

**Solving issue**  
Fixes: -

### DIFF
--- a/.github/workflows/bump-npm-version.yml
+++ b/.github/workflows/bump-npm-version.yml
@@ -16,10 +16,10 @@ jobs:
   release-angular:
     runs-on: ubuntu-latest
     steps:
-      - name: Fail if branch is not main
-        if: github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main'
+      - name: Fail if branch is not develop
+        if: github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/develop'
         run: |
-          echo "This workflow should not be triggered on a branch other than main"
+          echo "This workflow should not be triggered on a branch other than develop"
           exit 1
 
       - name: Check out code
@@ -46,6 +46,7 @@ jobs:
           title: release/@scania/tegel@${{ github.event.inputs.version }}
           commit-message: Release of @scania/tegel@${{ github.event.inputs.version }}
           branch: release/tegel@${{ github.event.inputs.version }}
+          base: main
           delete-branch: true
 
   on-failure:
--- a/.github/workflows/bump-npm-version.yml
+++ b/.github/workflows/bump-npm-version.yml
@@ -47,7 +47,6 @@ jobs:
           commit-message: Release of @scania/tegel@${{ github.event.inputs.version }}
           branch: release/tegel@${{ github.event.inputs.version }}
           base: main
-          delete-branch: true
 
   on-failure:
     runs-on: ubuntu-latest
--- a/.github/workflows/bump-npm-version.yml
+++ b/.github/workflows/bump-npm-version.yml
@@ -1,0 +1,60 @@
+name: Bump NPM version (core/angular/react)
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release type'
+        required: true
+        default: 'patch'
+        type: choice
+        options:
+          - major
+          - minor
+          - patch
+
+jobs:
+  release-angular:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if branch is not main
+        if: github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main'
+        run: |
+          echo "This workflow should not be triggered on a branch other than main"
+          exit 1
+
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Set Tegel user
+        run: git config --global user.name "Tegel - Scania" && git config --global user.email "tegel.design.system@gmail.com"
+
+      - name: Core - Bump version
+        run: npm version ${{ github.event.inputs.version }}
+        working-directory: packages/core
+
+      - name: Angular - Bump version
+        run: npm version ${{ github.event.inputs.version }}
+        working-directory: packages/angular
+
+      - name: React - Bump version
+        run: npm version ${{ github.event.inputs.version }}
+        working-directory: packages/react
+
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v3
+        with:
+          title: release/@scania/tegel@${{ github.event.inputs.version }}
+          commit-message: Release of @scania/tegel@${{ github.event.inputs.version }}
+          branch: release/tegel@${{ github.event.inputs.version }}
+          delete-branch: true
+
+  on-failure:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remove branch on failure
+        env:
+          npm_version: ${{ github.event.inputs.version }}
+        run: |
+          if [ "${{ job.status }}" == "failure" ]; then
+            git push origin --delete @scania/tegel@${{ github.event.inputs.version }}
+          fi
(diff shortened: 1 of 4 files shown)


.github/workflows/release-angular.yml
```diff
@@ -46,23 +46,24 @@ jobs:
...
 
-      - name: Angular - CD into correct folder
-        run: cd packages/angular
-
       - name: Angular - Bump version
         run: npm version ${{ github.event.inputs.tegelVersion }}
+        working-directory: packages/angular
 
       - name: Angular - Remove package-lock
         run: rm -rf package-lock.json
```
alexandereneroth, line 54: What was the reason for removing this? (just curious)

mJarsater: Do you mean the "cd:ing" into the correct folder? That is all handled by the last step, the creation of the PR.

alexandereneroth (approved):

.github/workflows/release-angular.yml
alexandereneroth: Ah no what was the reason for removing package-lock.json?

alexandereneroth (commented):
Not feedback on the changes in this PR, but to the actions:
If I understand the code correctly it seems like it is written to release from develop and then create a release branch and merge to main. 

Could it be the other way around, that we create a release branch from develop, merge it (when release notes are done), tag, and then do the release in main when it's finalised?

mJarsater:
> Not feedback on the changes in this PR, but to the actions: If I understand the code correctly it seems like it is written to release from develop and then create a release branch and merge to main.
> 
> Could it be the other way around, that we create a release branch from develop, merge it (when release notes are done), tag, and then do the release in main when it's finalised?
> 
> That would seem to be a more standard way of doing things

No, it is set to release for what ever branch you choose. But since it is not yet inside of main, you can't currently do a release from main. So I think the tactic you suggested above should still work.

.github/workflows/release-angular.yml
mJarsater: Ah, that was just due to us being **sure** we are using the latest @scania/tegel version as a base

.github/workflows/bump-npm-version.yml
```diff
@@ -0,0 +1,60 @@
...
+          - major
+          - minor
+          - patch
+
+jobs:
+  release-angular:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if branch is not main
+        if: github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/main'
```
alexandereneroth, line 20: I believe it should be 'develop' as we're branching from develop when preparing a release, and then merging in to main when we are ready to release

.github/workflows/release-core.yml
```diff
@@ -51,27 +48,18 @@ jobs:
...
 
-      - name: Core - CD into correct folder
-        run: cd packages/core
-
-      - name: Core - Bump version
-        run: npm version ${{ github.event.inputs.version }}
-
       - name: Core - Read Package.json Version
         id: version
         run: echo "version=$(jq -r .version package.json)" >> $GITHUB_OUTPUT
```
alexandereneroth, line 53: Will this work? GPT-4 said this:

1. In the "Core - Read Package.json Version" step, you're writing to `$GITHUB_OUTPUT`, but this environment variable is not defined in GitHub Actions. You should use `::set-output name=version::$(jq -r .version package.json)` to set the output of a step. Here's the corrected step:

```
- name: Core - Read Package.json Version
  id: version
  run: echo "::set-output name=version::$(jq -r .version package.json)"
  working-directory: packages/core
```

🤷‍♂️

.github/workflows/release-core.yml
```diff
@@ -80,28 +68,27 @@ jobs:
...
       - name: Remove branch and git tag on failure
         env:
-          npm_version: ${{ needs.release.outputs.version }}
+          npm_version: ${{ needs.release-core.outputs.version }}
         run: |
           if [ "${{ job.status }}" == "failure" ]; then
             tagname="@scania/tegel@$npm_version"
             git push --delete origin $tagname
-            branchname="release/@scania/tegel@$npm_version"
-            git push origin --delete $branchname
```
alexandereneroth, line 106: do you need to check out the repository first?

.github/workflows/release-core.yml
```diff
@@ -80,28 +68,27 @@ jobs:
...
     needs: release-core
     runs-on: ubuntu-latest
 
     steps:
       - name: Remove branch and git tag on failure
         env:
-          npm_version: ${{ needs.release.outputs.version }}
+          npm_version: ${{ needs.release-core.outputs.version }}
         run: |
           if [ "${{ job.status }}" == "failure" ]; then
```
alexandereneroth, line 91: Is this redundant since its already on-failure, or is on-failure just a name?

alexandereneroth:
GPT-4 gave me this feedback on the react workflow, some of which seems to be irrelevant, but thought you might see something i dont.

This GitHub Actions workflow is generally well designed, but there are a few areas for improvement:

1. In the React - Install with latest tegel package step, you're using steps.version.outputs which doesn't output anything. You should use `steps.version.outputs.version` instead.

2. The React - Build step is missing the working-directory property. If the build command needs to be run in the packages/react directory, you should specify it like you did in the other steps.

3. In the on-failure job, you're trying to delete a git tag using `git push origin --delete @scania/tegel@${{ needs.release-react.outputs.version }}`. However, the release-react job doesn't output a version. You should add an output for the version in the release-react job.

4. Also, the on-failure job should probably use run: `git tag --delete ...` instead of run: `git push origin --delete ...` to delete a local tag. If you want to delete a remote tag, you should use run: `git push --delete origin tagName`.

.github/workflows/release-core.yml
mJarsater: on-failure is just a name

.github/workflows/release-core.yml
mJarsater: They are deprecated: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/bump-npm-version.yml
mJarsater: Correcto!

.github/workflows/release-core.yml
mJarsater: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/bump-npm-version.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: npm version ${{ github.event.inputs.version }}
+        working-directory: packages/react
+
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v3
+        with:
+          title: release/@scania/tegel@${{ github.event.inputs.version }}
+          commit-message: Release of @scania/tegel@${{ github.event.inputs.version }}
+          branch: release/tegel@${{ github.event.inputs.version }}
+          delete-branch: true
```
alexandereneroth, line 50: If we're merging this to develop as well after, we shouldn't delete it

mJarsater: Yes, updated now!

theJohnnyMe (approved):

PR #395, opened 2023-10-26
build(node): version 20

mJarsater:
Setting node version 20 as min version. This is needed for us to be able to bump the stencil versions. 

**Solving issue**  
Fixes: -

**How to test**  
1. Check out branch
2. Try running install with node 18 in root - it should fail
3. Try running install with node 20 in root - it should succeed.


timrombergjakobsson:
Checked out the branch and ran `npm run build` with node 18 but it succeeded?

mJarsater:
> Checked out the branch and ran `npm run build` with node 18 but it succeeded?

Sorry, it should be -> run npm install in root -> it should fail

timrombergjakobsson (approved):
Tried it out with installing and building and running. All works!

alexandereneroth (approved):

nathalielindqvist (approved):
Works fine on Windows!

theJohnnyMe (approved):
Looks good to me!

Installing is prevented, however, code can be run on both the 18 and 20 versions of node event it is installed with a newer version.

If you consider this might be causing some issues, something like this can be implemented in tegel's code too: https://github.com/scania-digital-design-system/sdds-website/pull/425/files

That would clearly signify that even for use we require a different version.

PR #400, opened 2023-10-30
fix(dropdown): error focus state style alignment

mJarsater:
This PR fixes the issue of the error state not being correct whenever a Dropdown was focused/opened. 

It also makes sure that the filter input is focus whenever the icon in the dropdown is clicked. 

Before: 
Screenshot 2023-10-30 at 12 02 03

After:
Screenshot 2023-10-30 at 12 01 49

**Solving issue**  
Fixes: -

**How to test**  
1. Go to Dropdown
2. Set error to true
3. Open the Dropdown
4. Make sure the error state aligns with design

### DIFF
--- a/packages/core/src/components/dropdown/dropdown-button.scss
+++ b/packages/core/src/components/dropdown/dropdown-button.scss
@@ -49,6 +49,20 @@
 
     &.error {
       border-bottom: 1px solid var(--tds-negative);
+
+      &:focus {
+        border-bottom-color: transparent;
+
+        &::before {
+          content: '';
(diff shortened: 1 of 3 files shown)


alexandereneroth (approved):
In Figma inputs cant be focused and have error state at the same time, but this is of course needed. 
I think this is a good code solution for now even if it doesn't align with design 101%. 👍

timrombergjakobsson (approved):
Checks out, although one question not related to this PR/fix, should not the dropdown have `cursor: pointer` when hovering?

Issue #407, opened 2023-11-02
[Feature Request]: Remove Dotenv from the dependencies


gokberknur:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [X] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

Hi guys, not a big issue but  I saw you updated the node version. 

with the node 20 you no longer need dotenv as dependencies. 

### Background

you can read a bit more here: https://dev.to/cjreads665/nodejs-2060-say-goodbye-to-dotenv-2ijl

### Idea

make dependency array as clean as possible :)

mJarsater:
Lovely!

PR #408, opened 2023-11-02
build: removed dotenv dependency

mJarsater:
It is no longer needed as of node 20. We are also not really using it today, so I removed our only implementation of it in the storybook main.js. 

Our only use for it was to hide/show certain addons in production. I do not think we are really using these in development today, so I would suggest we remove them all together. 

**Solving issue**  
Fixes: https://github.com/scania-digital-design-system/tegel/issues/407

**How to test**  
1. Check out branch
2. Run `npm i` in core
3. Start storybook
4. Make sure everything runs as before.

### DIFF
--- a/packages/core/.storybook/main.cjs
+++ b/packages/core/.storybook/main.cjs
@@ -1,5 +1,3 @@
-require('dotenv').config();
-
 let addons = [
   '@storybook/addon-links',
   '@storybook/addon-essentials',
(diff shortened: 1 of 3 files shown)


timrombergjakobsson (approved):

theJohnnyMe:
I would propose we keep those tools in even if we do not use them as much as we thought, but do remove dependency on dotenv package.

timrombergjakobsson:
> I would propose we keep those tools in even if we do not use them as much as we thought, but do remove dependency on dotenv package.

I guess that makes sense. I suggest we revisit it further on and if we feel we don't use them at all, we remove them.

timrombergjakobsson (approved):

PR #434, opened 2023-11-17
fix(chip): no double click when switching to active for type radio

timrombergjakobsson:
**Describe pull-request**  
Setting chip to active when type radio should not require to double click. 

**Solving issue**  
Fixes: [CDEP-2995](https://tegel.atlassian.net/browse/CDEP-2995)

**How to test**  
1. Go to Storybook
2. Check in Chip component, and change to type radio, setting to active should not require double click

[CDEP-2995]: https://tegel.atlassian.net/browse/CDEP-2995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


timrombergjakobsson:
Related to a bug report in the support channel, linked to it in the Jira. Please have a look, will also add specs for it.

theJohnnyMe (changes requested):
Try this in PR preview link: 

1. Select "Radio" as Type
2. Click on a single Chip for a couple of times - you will be able to deselect it, I think it should not happen.

packages/core/src/components/chip/test/basic/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Badge - Basic</title>
```
nathalielindqvist, line 5:
```suggestion
    <title>Chip - Basic</title>
```

packages/core/src/components/chip/test/radio/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Badge - Basic</title>
```
nathalielindqvist, line 5:
```suggestion
    <title>Chip - Basic</title>
```

nathalielindqvist (commented):
I don't know if it's just for me, but the snapshots don't look like they are of the Chip component? This is what I see:
image
image

theJohnnyMe (changes requested):
Compared to production:

1. Button type is emitting 2 click events on single click
2. Radio type stops emitting Custom Events after clicking between all Chips in the row

Please double check it :)

timrombergjakobsson:
> Compared to production:
> 
> 1. Button type is emitting 2 click events on single click
> 2. Radio type stops emitting Custom Events after clicking between all Chips in the row
> 
> Please double check it :)

Could you please clarify, you mean these 2 things are in this PR, or in production?

timrombergjakobsson:
> Compared to production:
> 
> 1. Button type is emitting 2 click events on single click
> 2. Radio type stops emitting Custom Events after clicking between all Chips in the row
> 
> Please double check it :)

In production it also stops emitting sometimes after changing, also I could reproduce double emit on click in production also 🤔.

theJohnnyMe:
Button type seems to work as expected
Radio type seems to stop emitting after every Chip has been selected/deselected
Checkbox type seems to always send two CustomEvents

There is some issue with Radio and Checkbox type that has to be checked.

alfredskeden:
> Button type seems to work as expected Radio type seems to stop emitting after every Chip has been selected/deselected Checkbox type seems to always send two CustomEvents
> 
> There is some issue with Radio and Checkbox type that has to be checked.

Production:
I noticed that in storybook if you flip between for example radio and checkbox more events fire. So I flip once 2 events fire. I flip between the 5 times 6 events fire so it like remembers that. 
Regarding radio type its so strange behaviour that needs to be fixed. Radio chips should never be able to be "unchecked"?

In this pr:
yes type radio only trigger once for each "chip" and never for the starting one.
But same event trigger problem that it "builds up" an give more events for every time you flip between the types.

packages/core/src/components/chip/chip.tsx
```diff
@@ -55,7 +55,16 @@ export class TdsChip {
...
+    if (this.type === 'checkbox') {
+      // Toggle the prop on click
+      this.checked = !this.checked;
+    } else if (this.type === 'radio') {
+      // Always set it to true to enforce visual update for selected state
+      this.checked = true;
+    } else {
+      console.warn('Unsupported type in Chip component!');
+    }
+
```
theJohnnyMe, line 67: @timrombergjakobsson, I have been looking into this one for longer and found a bug that causes this.

The type button and type checkbox do not have this issue as they can be used as a single component. On the other side, the radio button is a component where you use at least two instances of it. During debugging, I was looking at what happens with the "checked" attribute in the DOM. Once it is assigned to the Chip that is radio type it stays there forever. Clicking on another Chip of the same radio type selects that one, but the previous one still has a "checked" attribute. Looking into native HTML input with type radio and other Design systems we realized the same happens there. We have tested it in form too, so form, on submitting will pick up "last" Chip on which user has clicked - no matter all of them might have "checked" as attribute.

What are possible options:

1. Add @Listen decorator to Chip component so it listens to its own tdsChange. If that specific instance of Chip component is emitting, it will set its own checked prop to true, while all the others will set it to false. In this example, we would need to enable Chip components with same "name" prop to be able to communicate to each other or better say - be aware of each other.
2. On click of Chip component, set checked prop to true (even if it already is true) to enforce styling update to blue. 

Number 2. you already did so kudos for that, I just reverted some changes to keep changes smaller. Please review it too and ping me if something more is needed.

This was indeed very interesting issue to debug. 

PS: Nevermind for multiple console logs on tdsChange, it seems to be issue related to Storybook, Chrome. I have checked it in Angular demo and there seems fine. So sorry to bother you around it. I noticed different behaviors of local environment if Chome and Firefox as well as in incognito mode.

theJohnnyMe:
Besides this one, there is a unit test issue remaining that Natahlie raised. 
Feel free to fix it in this one or if it is easier do it in a separate PR, your call. ;)

packages/core/src/components/chip/chip.tsx
timrombergjakobsson: @theJohnnyMe wow, interesting little bug we discovered here. So I reviewed your changes, and they seem to work. However, the local behavior of changeEvents makes it hard to trust that they are working properly with your changes. Did you `npm pack`and tried it out locally in Angular env to confirm that it worked?  If so, I'm ok with the changes you made, as it visually seems to work as intended.

timrombergjakobsson:
> Besides this one, there is a unit test issue remaining that Natahlie raised. Feel free to fix it in this one or if it is easier do it in a separate PR, your call. ;)

I can have a look!

packages/core/src/components/chip/chip.tsx
theJohnnyMe: @timrombergjakobsson  - I agree with you, I have not tested in Angular, but we should do it.
 It might be a bit tricky but not impossible:

- We would need to pack the core
- Then put it in angular and pack that one
- Test it in Angular demo

Let's try that during the day and see how it goes

mJarsater:
🥷 

"PS: Nevermind for multiple console logs on tdsChange, it seems to be issue related to Storybook, Chrome. I have checked it in Angular demo and there seems fine. So sorry to bother you around it. I noticed different behaviors of local environment if Chome and Firefox as well as in incognito mode."

Im pretty sure this is due to us adding an eventlistener in the story template. The template is gonna run each time you navigate to the story in storybook. This means that:
1. You navigate to the Chip in SB the first time
2. An evenlistener is added
3. You navigate away or do a change in a control
4. **Another** evenlistener is added
5. You navigate away or do a change in a control
6. **Another** evenlistener is added
7. You click the Chip
8. In the console you have 3 events emitted. 

It's not that the event is getting emitted 3 times, is just that we have added 3 different eventlisteners to the same event.

alfredskeden:
> It's not that the event is getting emitted 3 times, is just that we have added 3 different eventlisteners to the same event.

aha gotcha! :) 👍

timrombergjakobsson:
Btw, where do we stand on this one? These days been so divided so would be good to wrap it up and don't leave it hanging @theJohnnyMe!

theJohnnyMe:
> Btw, where do we stand on this one? These days been so divided so would be good to wrap it up and don't leave it hanging @theJohnnyMe!

I did not had time these days to take a look at code due to number of other actions. :( If anyone else has time to pack the code and test it Angular demo page, please do so. Also please drop the unit test files and add it to separate PR so we keep it small.

theJohnnyMe:
I will put this one to draft as it has a unit test too and I will open a new PR to address the bug.

theJohnnyMe:
PR for bug available here: https://github.com/scania-digital-design-system/tegel/pull/441
I will close this one actually but leave the branch alive if we would like to recycle it for the Chip unit test task.

timrombergjakobsson:
> > Btw, where do we stand on this one? These days been so divided so would be good to wrap it up and don't leave it hanging @theJohnnyMe!
> 
> I did not had time these days to take a look at code due to number of other actions. :( If anyone else has time to pack the code and test it Angular demo page, please do so. Also please drop the unit test files and add it to separate PR so we keep it small.

Should we really separate unit tests from a bugfix?

theJohnnyMe:
> > > Btw, where do we stand on this one? These days been so divided so would be good to wrap it up and don't leave it hanging @theJohnnyMe!
> > 
> > 
> > I did not had time these days to take a look at code due to number of other actions. :( If anyone else has time to pack the code and test it Angular demo page, please do so. Also please drop the unit test files and add it to separate PR so we keep it small.
> 
> Should we really separate unit tests from a bugfix?

Agreement from standup: In this case, I would say yes. We have not implemented test cases yet for chip, but let's check with Tatyana so we have a test case for this one in the case spreadsheet.

timrombergjakobsson:
> > > > Btw, where do we stand on this one? These days been so divided so would be good to wrap it up and don't leave it hanging @theJohnnyMe!
> > > 
> > > 
> > > I did not had time these days to take a look at code due to number of other actions. :( If anyone else has time to pack the code and test it Angular demo page, please do so. Also please drop the unit test files and add it to separate PR so we keep it small.
> > 
> > 
> > Should we really separate unit tests from a bugfix?
> 
> Agreement from standup: In this case, I would say yes. We have not implemented test cases yet for chip, but let's check with Tatyana so we have a test case for this one in the case spreadsheet.

Ah ok, I still believe we should add it. But if the majority in the team think not, then we go with that 👍

PR #435, opened 2023-11-21
test(dropdown): added unit tests

alfredskeden:
**Describe pull-request**  
Followed the test steps in ticket/excel to test.

**Solving issue**  
Fixes: [CDEP-2960](https://tegel.atlassian.net/browse/CDEP-2960)

**How to test**  
1. Navigate to packages/core.
2. Check the code for tests in dropdown component.

**Additional context** 
We are still in exploration for good testing.

[CDEP-2960]: https://tegel.atlassian.net/browse/CDEP-2960


nathalielindqvist (approved):
Nice work! LGTM :)

packages/core/src/components/dropdown/test/default/dropdown.e2e.ts
```diff
@@ -0,0 +1,82 @@
...
+    await page.goto(componentTestPath);
+    const dropdown = page.locator('tds-dropdown');
+    expect(dropdown).toBeTruthy();
+    await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });
+  });
+
+  test('default dropdown have label outside', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const dropdown = page.locator('tds-dropdown');
+    const labelOutside = dropdown.locator('div[class="label-outside"]');
```
alexandereneroth, line 17: This test still passed when i changed to label inside 💡 Something's not working

alexandereneroth: However, I think just having the screenshot could be good enough in this case as this is testing implementation details 🤔

alfredskeden: Kinda updated this I think, The playwright locator still tries to find the label-outside even tho you put it for label inside. So i just now check if the text contains withinit and it breaks if you change to label inside.

packages/core/src/components/dropdown/test/default/dropdown.e2e.ts
```diff
@@ -0,0 +1,82 @@
...
+  test('renders default dropdown correctly', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const dropdown = page.locator('tds-dropdown');
+    expect(dropdown).toBeTruthy();
+    await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });
+  });
+
+  test('should have a label outside with text', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const labelOutside = page.locator('tds-dropdown > div[class="label-outside"]');
```
alexandereneroth, line 16: If we refactor the dropdown in the future and afterwards the div inside doesn't exist or the class "label-outside" is called something else, for example, the test will break. 
Unfortunately tests like these will be in the way for future work. 

I can recommend this article on the topic https://kentcdodds.com/blog/testing-implementation-details

alexandereneroth: Playwright also mentions this here https://playwright.dev/docs/best-practices#prefer-user-facing-attributes-to-xpath-or-css-selectors 
getByRole seems like a good one, but it requires that the component have been implemented in an accessible way, which is perhaps not the case a lot of the time in this code base unfortunately.

alexandereneroth: Sometimes, screenshots might be the only way we can test something in a resilient way.

alexandereneroth: I am sorry, should have been clearer about this before

alfredskeden: Hehe no problem ma dude :) @alexandereneroth Sounds like a really nice article and I think this is what I feel when I do these kind of tests. "There must be a better way" but then I just bite and try to test the specs from the excel sheet. I think we have to do another think around these with a lot more tries to see what fits us. I agree totally the tests will break if something changes and we need to other checks. Screenshots are so good tests!

packages/core/src/components/dropdown/test/default/index.html
```diff
@@ -20,9 +20,14 @@
       helper="Helper text"
       size="lg"
       open-direction="auto"
+      data-testid="tds-dropdown-testid"
```
timrombergjakobsson, line 23: How come using `data-testid`?

alfredskeden: There is a way of getting that element using 

```
const dropdown = page.getByTestId('tds-dropdown-testid');
```
and the frameworks recommend to use that for getting a specific element instead of `page.locate('tds-dropdown');`

packages/core/src/components/dropdown/test/error/dropdown.e2e.ts
```diff
@@ -6,79 +6,34 @@ const componentTestPath = 'src/components/dropdown/test/error/index.html';
...
-    const dropdownSelect = dropdown.locator('div[class="dropdown-select lg"]');
-    const button = dropdownSelect.locator('button');
-    await button.click();
-    const dropdownOptionOne = dropdown.locator('tds-dropdown-option[value="option-1"]');
-    await dropdownOptionOne.click();
-    await expect(dropdown).toHaveAttribute('value', 'option-1');
-  });
+    /* before clicking dropdownlist should not be visible, the button should be */
+    await expect(placeholderElement).toBeVisible();
+    await expect(dropdownOptionOne).not.toBeVisible();
```
alexandereneroth, line 31: Nice that there is a check for it being hidden before as well. 
An alternative here could be .toBeHidden(). However no need as this does the same thing.

alfredskeden: Oh yeah its kinda the same check. feels like the not makes it easier to "read" the code that it should not be visible than toBeHidden and then to be visible but that might just be me? @timrombergjakobsson any inputs?

packages/core/src/components/dropdown/test/default/dropdown.e2e.ts
```diff
@@ -0,0 +1,97 @@
...
+    /* check so only one "Option 1" text exists */
+    const option1Element = page.getByText(/Option 1/);
+    await expect(option1Element).toHaveCount(1);
+
+    /* Click Option 1 element */
+    await dropdownOptionOne.click();
+
+    /* Check so its dropdown list is not visible anymore and there now 2 "Option 1" texts in the dom */
+    await expect(dropdownOptionOne).not.toBeVisible();
+    await expect(option1Element).toHaveCount(2);
```
alexandereneroth, line 70: I think that testing that there are two "Option 1" texts in the DOM is testing an implemention detail, since only one is visible. 
A different implementation might remove the "Option 1" that is in the dropdown menu from the DOM entirely when it closes. 
Seems like .toBeVisible could be used here instead as it's supposed to pass if at least one element in the list is visible 🤔

alfredskeden: Yes I learned so much since this, I'll check and update some 👍

packages/core/src/components/dropdown/test/default/index.html
```diff
@@ -0,0 +1,35 @@
...
+
+  <body>
+    <tds-dropdown
+      name="dropdown"
+      label="Label text"
+      label-position="outside"
+      placeholder="Placeholder"
+      helper="Helper text"
+      size="lg"
+      open-direction="auto"
```
alexandereneroth, line 22: I'd say the most important setup to test would be the tds-dropdown without any attributes set (or as few as possible).
What do you think?
I believe this is just the properties that exist in our storybook by default right?

packages/core/src/components/dropdown/test/error/dropdown.e2e.ts
timrombergjakobsson: I think if they both are doing the same stuff, then it does not really matter which one. Although myself I think `.toBeHidden()` is easier to read hahah. But it is just cos it maps pretty nicely with `.toBeVisible()`

packages/core/src/components/dropdown/test/error/dropdown.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    /* Check diff on screenshot */
+    await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });
+  });
+
+  test('find helper text and check if visible and have icon', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const helperText = page.getByText(/Helper text/);
+    await expect(helperText).toBeVisible();
+    const helperTextIcon = helperText.locator('tds-icon');
```
alexandereneroth, line 20: Hmm this is actually testing an implementation detail, since the tds-icon is added inside of our component declaration. 
I think we could leave this out and rely on the screenshot, WDYT?

alexandereneroth: It is a bit of an edge case though, as we're unlikely to use any other element for the icon in the future 🤔

packages/core/src/components/dropdown/test/error/dropdown.e2e.ts
alfredskeden: Ok have updated.

packages/core/src/components/dropdown/test/error/dropdown.e2e.ts
alfredskeden: No you are right, I do getByRole('img') instead so that we can use any icon there instead

packages/core/src/components/dropdown/test/default/index.html
alfredskeden: Added a basic aswell without any props

packages/core/src/components/dropdown/test/multiselect/dropdown.e2e.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    /* click the button */
+    const placeholderElement = page.getByText(/Placeholder/);
+    await placeholderElement.click();
+
+    /* get all checkboxes */
+    const dropdownListElementOneButton = page
+      .locator('tds-dropdown-option')
+      .filter({ hasText: /Option 1/ })
+      .locator('tds-checkbox');
```
alexandereneroth, line 52: Same thing here as what I mentioned with the tds-icon in the error text actually. Maybe we could just remove the .locator('tds-checkbox') and it would still work right?

alexandereneroth: ah you're looking for the checked attribute 🤔

alexandereneroth: Hmm, there is a ".toBeChecked()" might be better (assuming it works with our component)
https://playwright.dev/docs/input#checkboxes-and-radio-buttons

alexandereneroth: I guess this might be impossible if we haven't properly made it accessible

alfredskeden: I can check on it

alfredskeden: [3537a14](https://github.com/scania-digital-design-system/tegel/pull/435/commits/3537a140bde6204b64b5dbfdfdd1c48635d66009)
Found this solution, what do you think? there are 4 places that updates when something gets checked.

tds-dropdown-option get aria-selected
tds-checkbox gets the checked
input gets the aria-checked
and the dropdown value becomes updates with all checked values...

alexandereneroth: Nice, I think that works. I believe it always needs the aria-selected attribute since they can never be just native checkboxes.
Lets go with this if you agree 👍

alexandereneroth (approved):
LGTM 💯

packages/core/src/components/dropdown/test/multiselect/dropdown.e2e.ts
alfredskeden: cool sounds cool with me.

PR #374, opened 2023-10-20
test: playwright setup 

mJarsater:
This PR adds the initial setup for using Playwright as testing framework for our components. 

It installs "playwright" and "stencil-playwright", the second one being a wrapper to allow for the use of web components with Playwright.

It also adds some basic/initial tests for the Accordion. 

**How to test:**
1. Check out the branch
2. cd into `packages/core`
3. Run `npm install`
4. Run `npx playwright test`

PR #436, opened 2023-11-21
test(textarea): unit test

alfredskeden:
**Describe pull-request**  
Followed the test steps in ticket/excel to test.

**Solving issue**  
Fixes: [CDEP-2975](https://tegel.atlassian.net/browse/CDEP-2975)

**How to test**  
1. Navigate to packages/core.
2. Check the code for tests in textarea component.

**Additional context** 
We are still in exploration for good testing practices.

[CDEP-2975]: https://tegel.atlassian.net/browse/CDEP-2975


packages/core/src/components/textarea/test/default/textarea.e2e.ts
```diff
@@ -0,0 +1,37 @@
+import { test } from 'stencil-playwright';
+import { expect } from '@playwright/test';
+
+const componentTestPath = 'src/components/textarea/test/default/index.html';
+
+test.describe('tds-textarea', () => {
+  test('renders default textarea correctly', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const textarea = page.locator('tds-textarea');
+    await expect(textarea).toHaveClass(/hydrated/);
```
timrombergjakobsson, line 10: Just a wonder, should we really test this? That the class `hydrated` is present?

alfredskeden: Looks like the reviewers where ok with it in #374 
I just took that test from there and I guess if hydrated would somehow get removed as class is no bueno and the test should catch that?

timrombergjakobsson: Ping @alexandereneroth @theJohnnyMe @nathalielindqvist @EmelieLitwin

alfredskeden: Oh I didn't know that was stencil stuff :) Sorry im so new. yeah then we shouldn't test it according to the guidelines and stencil should do those tests.

timrombergjakobsson: No worries man, we are also new :)! I think Martin added it additionally, and then we briefly touched upon it in a meeting the other week. So like I said, from my perspective it feels weird to test for it. But, if we in the team feels it provides value then we can do it.

alfredskeden: Ok, I removed it for now and we can add those tests again later I think. This is not a final PR for adding tests, it only handles some of the tests that we have in the [unit test spec sheet](https://scaniaazureservices-my.sharepoint.com/:x:/r/personal/tatyana_bumazhkina_scania_com/Documents/Unit%20test%20scenarios.xlsx?d=wca48bf6d92cf4e8c8f456337e2002ec4&csf=1&web=1&e=9iV53q&nav=MTVfezdBNzNBRjE3LTNBNjUtNDMyQy1BRkZFLTU0NkQ4OEE0QTRBRX0)

nathalielindqvist: Yeah this came up on the testing start-up meeting as well. I think we said we don't need to test it since it's Stencil specific like you said @timrombergjakobsson, nice catch! We can probably remove it from the tests for now and we can add it back in if we see that it provides value later on :) great work so far @alfredskeden! 👏🎉

nathalielindqvist (commented):
Don't know if it's just for me, but the snapshots don't seem to look exactly like the component? Or maybe the tests can't fully recreate the component as it looks in Storybook? This is what I see in the PR:
image
image

And this is what they look like in Storybook:
image
image

packages/core/src/components/textarea/test/default/textarea.e2e.ts
timrombergjakobsson: Yes, I think it is better that we focus on verifying the actual behavior and functionality of our components, such as interactions, event handling, rendering correctness, and expected output etc. We have to trust Stencil.js and their internal testing of their own "stuff" :). But awesome work @alfredskeden! Fun stuff happening!!

timrombergjakobsson:
> Don't know if it's just for me, but the snapshots don't seem to look exactly like the component? Or maybe the tests can't fully recreate the component as it looks in Storybook? This is what I see in the PR: image image
> 
> And this is what they look like in Storybook: image image

Yes, Im a bit confused on this part. I guess it depends if we do:
 `await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });`
 vs
`await expect(component).toHaveScreenshot({ maxDiffPixels: 0 });`
But a bit unsure on the correct practice here.

alfredskeden:
> Yes, Im a bit confused on this part. I guess it depends if we do: `await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });` vs `await expect(component).toHaveScreenshot({ maxDiffPixels: 0 });` But a bit unsure on the correct practice here.

Hmm yes it will still detect but I guess it takes a screenshot of the whole page and compares.
Its good to check the whole page so that the component dont change outside of its component somehow.
and if I only target `tds-textarea` for the screenshot I get this
from html inspect:
image

Screenshot:
tds-textarea-renders-default-textarea-correctly-1-linux

timrombergjakobsson:
> Don't know if it's just for me, but the snapshots don't seem to look exactly like the component? Or maybe the tests can't fully recreate the component as it looks in Storybook? This is what I see in the PR: image image
> 
> And this is what they look like in Storybook: image image

Yes, it's a question of 
`await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });`
vs
`await expect(component).toHaveScreenshot({ maxDiffPixels: 0 });`
Both is valid, but I guess the question is when to use it? 
I guess they both have pros and cons. 

`page`:

- Use Cases: Use this approach for end-to-end tests, visual regression tests, or when you want to ensure that the entire page looks as expected.
- Drawbacks: It might capture more than what you need for component-level tests, leading to larger and potentially less focused screenshots.

`component`:

- Use Cases: Use this approach for component-level tests or when you want to verify the appearance of a specific element or component in isolation.
- Limitations: It may not cover interactions between components or how the component looks in the context of the entire page layout.

packages/core/src/components/textarea/test/default/textarea.e2e.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  test('default textarea - click on text area - cursor should appear', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const textarea = page.locator('tds-textarea').locator('textarea');
+    const myEventSpy = await page.spyOnEvent('click');
+    await textarea.click();
+    /* Expect to have received an event from clicking on the textarea */
+    expect(myEventSpy).toHaveReceivedEvent();
+    const textareaCursorState = await textarea.evaluate(
+      (button2) => getComputedStyle(button2).cursor,
```
nathalielindqvist, line 23: Is there a benefit to renaming "button2" to something else in this context?

packages/core/src/components/textarea/test/read-only/textarea.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+    await expect(textarea).toHaveAttribute('readonly');
+    const readonlyIcon = page.locator('tds-icon[name="edit_inactive"]');
+    /* Expect a tds-icon to be present when readonly mode is active */
+    expect(readonlyIcon).not.toBeNull();
+  });
+
+  // Error: locator.fill: Target closed
+  // Not able to update the since we get Playwright error here when trying to input a readonly input.
+  // Skipping for now.
+  test.skip('Type text in textarea - no possibility to type text', async ({ page }) => {
```
nathalielindqvist, line 31: Could we test that the component does not receive a click event instead to test that it's not interactable?

packages/core/src/components/textarea/test/default/textarea.e2e.ts
alfredskeden: hmm totally, updated to style instead. But yeah `button2` was a copy pasta :D

packages/core/src/components/textarea/test/read-only/textarea.e2e.ts
alfredskeden: Hmm I tried to do this, Its just that the `await textarea.click()` does get the error. I wonder if just checking if it has the disabled attribute for now.

nathalielindqvist: That could be a way to do it. Maybe this should be discussed with Tatyana though?

timrombergjakobsson: Maybe we could test for that when you try and type text into it, it remains empty?

```
await textarea.type('Adding some text');
  
 // Check that the textarea remains empty (no text was entered)
 expect(await textarea.inputValue()).toBe('');`
```

timrombergjakobsson: If we wanna test also for the disabled attribute we could do:

```
const isDisabled = await textarea.getAttribute('disabled');
expect(isDisabled).toBeTruthy();
```

alfredskeden: > Maybe we could test for that when you try and type text into it, it remains empty?
> 
> ```
> await textarea.type('Adding some text');
>   
>  // Check that the textarea remains empty (no text was entered)
>  expect(await textarea.inputValue()).toBe('');`
> ```

I still get `Error: locator.fill: Target closed` when trying to do stuff to a readonly field. might just check for the readonly tag. Since its not disabled, its "readonly"

theJohnnyMe:
In the [spreadsheet](https://scaniaazureservices-my.sharepoint.com/:x:/r/personal/tatyana_bumazhkina_scania_com/Documents/Unit%20test%20scenarios.xlsx?d=wca48bf6d92cf4e8c8f456337e2002ec4&csf=1&web=1&e=1Lzg7J&nav=MTVfezdBNzNBRjE3LTNBNjUtNDMyQy1BRkZFLTU0NkQ4OEE0QTRBRX0) we need to check for "No label" works.

1. If it is still relevant should we check if whatever string is assigned to `label` prop exists anywhere under tds-textarea if `lable-position` is set to `no-label`?
2. I see we have basic and default folders for test with very small differences between them - aka taking an extra screenshot in the default compared to basic, should we remove basic folder then?

alfredskeden:
> 1. If it is still relevant should we check if whatever string is assigned to `label` prop exists anywhere under tds-textarea if `lable-position` is set to `no-label`?

Good catch. before I wasnt sure how to test this but now it quite clear. Added the tests for default and read-only to check for label.
also would like to test label-position="inside" also but might be able to add that later.

> 2. I see we have basic and default folders for test with very small differences between them - aka taking an extra screenshot in the default compared to basic, should we remove basic folder then?

Regarding this @alexandereneroth wanted to test the component when it does not have any props to make sure it still works as expected. And I think thats we have one basic folder with no props. and then a default test for how the default code looks like in storybook.

packages/core/src/components/textarea/test/read-only/textarea.e2e.ts
```diff
@@ -0,0 +1,33 @@
+import { test } from 'stencil-playwright';
+import { expect } from '@playwright/test';
+
+const componentTestPath = 'src/components/textarea/test/read-only/index.html';
+
+test.describe.only('tds-textarea-read-only', () => {
```
alfredskeden, line 6: @alexandereneroth @theJohnnyMe just noticed this one sneaked passed me. Opening a new small pr where I removed this.

PR #440, opened 2023-11-28
test(button): refactor unit tests

alfredskeden:
**Describe pull-request**  
This pull request refactors the unit tests for button. follow this more
https://kentcdodds.com/blog/testing-implementation-details
with less implementation details for better testing and developer experience

Also updates all images to be page specific and not component specific

**Solving issue**  
Fixes: [CDEP-2961](https://tegel.atlassian.net/browse/CDEP-2961)

**How to test**  
1. Navigate to packages/core.
2. Check the code for tests in button component.

**Additional context** 
We are still in exploration for good testing.

[CDEP-2961]: https://tegel.atlassian.net/browse/CDEP-2961

### DIFF
--- a/packages/core/src/components/button/test/basic/index.html
+++ b/packages/core/src/components/button/test/basic/index.html
@@ -12,6 +12,6 @@
   </head>
 
   <body>
-    <tds-button id="basicButton" text="Button" data-testid="tds-button-testid"></tds-button>
+    <tds-button text="Button" data-testid="tds-button-testid"></tds-button>
   </body>
 </html>
--- a/packages/core/src/components/button/test/danger/index.html
+++ b/packages/core/src/components/button/test/danger/index.html
@@ -13,7 +13,6 @@
 
   <body>
     <tds-button
-      id="basicButton"
       text="Button"
       variant="danger"
       size="sm"
--- a/packages/core/src/components/button/test/disabled/index.html
+++ b/packages/core/src/components/button/test/disabled/index.html
@@ -13,7 +13,6 @@
 
   <body>
     <tds-button
-      id="disabledButton"
       type="button"
       variant="primary"
       size="lg"
--- a/packages/core/src/components/button/test/ghost/index.html
+++ b/packages/core/src/components/button/test/ghost/index.html
@@ -13,7 +13,6 @@
 
   <body>
     <tds-button
-      id="basicButton"
       text="Button"
       variant="ghost"
       size="xs"
--- a/packages/core/src/components/button/test/icon/index.html
+++ b/packages/core/src/components/button/test/icon/index.html
@@ -12,7 +12,7 @@
   </head>
 
   <body>
-    <tds-button id="basicButton">
+    <tds-button>
       <tds-icon
         slot="icon"
         size="20px"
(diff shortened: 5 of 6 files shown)


theJohnnyMe (approved):
Looks good to me buddy, nice work!

packages/core/src/components/button/test/basic/index.html
```diff
@@ -12,6 +12,6 @@
   </head>
 
   <body>
-    <tds-button id="basicButton" text="Button"></tds-button>
+    <tds-button id="basicButton" text="Button" data-testid="tds-button-testid"></tds-button>
```
theJohnnyMe, line 15: Only thing you can get rid off is "id" in all index.html files accross button test. I think it is now used any more so it is not needed...

packages/core/src/components/button/test/basic/button.e2e.ts
```diff
@@ -3,97 +3,40 @@ import { expect } from '@playwright/test';
 
 const componentTestPath = 'src/components/button/test/basic/index.html';
 
-test.describe('tds-button', () => {
+test.describe('tds-button-basic', () => {
   test('renders basic button correctly', async ({ page }) => {
     await page.goto(componentTestPath);
-    const button = page.locator('tds-button');
-    await expect(button).toHaveScreenshot({ maxDiffPixels: 0 });
+    const button = page.getByTestId('tds-button-testid');
```
timrombergjakobsson, line 9: Question: should we use `getByRole` here also? As we can. Or does using `getByTestId` here provide anything else?

packages/core/src/components/button/test/basic/button.e2e.ts
```diff
@@ -3,97 +3,40 @@ import { expect } from '@playwright/test';
...
+    await expect(button).toBeVisible();
   });
 
   test('Check so that height is correct to lg/default measurements', async ({ page }) => {
     await page.goto(componentTestPath);
-    const button = page.locator('tds-button');
-    const innerButton = button.locator('button');
-    const buttonHeight = await innerButton.evaluate((button2) => getComputedStyle(button2).height);
+    const button = page.getByText('Button', { exact: true });
+    const buttonHeight = await button.evaluate((style) => getComputedStyle(style).height);
```
timrombergjakobsson, line 39: Nice!

alfredskeden: Yeah not checking the height might be a breaking change in some applications using the code so needs to be checked :D

packages/core/src/components/button/test/basic/button.e2e.ts
alfredskeden: Erm yeah I can see few. the test id selects `tds-button` while `.getByRole('button')` would get the <button> element. So its really what we would like to test here. Check if tds-button is in the dom or if it has a <button> inside of it...

alfredskeden: Updated comment didnt see the `<button>` dissapeared at first. But no problem, I guess we have to decide on a pattern here on what we would like to test on the first test with the screenshot.

timrombergjakobsson: Yeah! What do you think @theJohnnyMe and @alexandereneroth?

PR #444, opened 2023-12-04
feat(dropdown): added support for normalized text when searching

alfredskeden:
**Describe pull-request**  
Requested in the support channel to have the dropdown search for accent answers for international users. opt-out able.

**Solving issue**  
Fixes: [CDEP-3001](https://tegel.atlassian.net/browse/CDEP-3001)

**How to test**  
1. Run storybook
2. Go do dropdown
3. Enable filter
4. Start searching for iles or aland
5. îles Åland should show up in the dropdown list
6. Turn off Normalized Text in storybook
7. Do the same search again and notice that its harder to find the dropdown item in the list.

**Suggested test steps**
- [x] Storybook controls
- [x] Added/Updated unit tests that needs to be checked.

**Additional context**  
This is my first PR working with stencil and the this big storybook that you guys have so I'm sure that I've missed something or named something super wrong.

[CDEP-3001]: https://tegel.atlassian.net/browse/CDEP-3001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alfredskeden:
Noticed something strange when working with this but very minor.

Accent characters top get cut off in dropdown input and only on some characters in the dropdown list.
(Same in tds-textfield)
see image: 
image

packages/core/src/components/dropdown/test/filter/normalize-text-false.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Dropdown - Filter</title>
```
alfredskeden, line 5: Was thinking just a quick one on this since I wrote this test case before implementation of the code. 

Should it be on its own, Or can it be together with the filter folder and tests just update the path to this one where `normalized-text` is added and changed? Should be just add it with an document.addAttribute before doing the tests? How you guys feel?

Like this works but think we can do better.

theJohnnyMe:
> Noticed something strange when working with this but very minor.
> 
> Accent characters top get cut off in dropdown input and only on some characters in the dropdown list. (Same in tds-textfield) see image (noticed the top of å and top of î: image

Nice catch! It is related to line-height property of font. I have increased as original line-height does not work for sings like: ÅÖÄ etc...

packages/core/src/components/dropdown/test/filter/normalize-text-false.html
theJohnnyMe: Sorry, I do not follow, can you please elaborate more?

alfredskeden: No problem :) Erm i'm like just questioning myself on how to organise specific tests regarding the "same" kinda of tests. So I just made another folder inside the filter folder and moved the "new" test inside of there. Think it works but please see if you think it makes sense.

alfredskeden:
> Nice catch! It is related to line-height property of font. I have increased as the original line-height does not work for signs like: ÅÖÄ etc...

Dude awesome thanks a bunch, And hope it does not break any of the designs you didn't have to update any of the screenshots?

theJohnnyMe (approved):
LGTM! Nice work man!

alfredskeden:
> The test failed initially on commit, but I ran them again and everything went OK.

Also sometimes get strange screenshot diff on like components that should not get diff and then checking the diff screenshots its not clear what happens just some kind of lag. then run the tests again and it works.

theJohnnyMe:
> > The test failed initially on commit, but I ran them again and everything went OK.
> 
> Also sometimes get strange screenshot diff on like components that should not get diff and then checking the diff screenshots its not clear what happens just some kind of lag. then run the tests again and it works.

Exactly that happened to me too!

timrombergjakobsson (approved):
💯

PR #449, opened 2023-12-06
fix(header-cell): changed sorting arrow direction in header cell

Wadzabi:
**Describe pull-request**  
Current direction of sorting arrow does not align with standard direction found in common  web design libraries. Changed direction of arrow to align with standard sorting conventions. Arrow up is used for ascending order and arrow down is used for descending order. Also replaced hover icon when switching sorting key with arrow up to make it clear what direction the table will sort in on first click.

**How to test**  
1. Go to sorting section of table component in storybook
2. Click on header cell and see direction of arrow on ascending and descending sorting events.

**Screenshots**  
Old sorting:
Sorting_arrow_old

New sorting:
Sorting_arrow_new

Material design documentation about sorting in tables:
Screenshot 2023-12-06 at 19 50 50
Screenshot 2023-12-06 at 19 51 29

**Additional context**  
[Material design table documentation](https://m2.material.io/components/data-tables#anatomy)


theJohnnyMe:
@Wadzabi  - Thanks a lot buddy for this great idea and work. I will share this with our designers and we will come back to you ;)

packages/core/src/components/table/table-header-cell/table-header-cell.tsx
```diff
@@ -42,7 +42,7 @@ export class TdsTableHeaderCell {
 
   @State() textAlignState: string;
 
-  @State() sortingDirection: 'asc' | 'desc' = 'asc';
+  @State() sortingDirection: 'asc' | 'desc' = 'desc';
```
mJarsater, line 45: Breaking?

packages/core/src/components/table/table-header-cell/readme.md
```diff
@@ -20,7 +20,7 @@
 
 | Event     | Description                                                                                                                                                          | Type                                                                                      |
 | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------- |
-| `tdsSort` | Sends unique Table identifier, column key and sorting direction to the tds-table-body component, can also be listened to in order to implement custom-sorting logic. | `CustomEvent<{ tableId: string; columnKey: string; sortingDirection: "asc" \| "desc"; }>` |
+| `tdsSort` | Sends unique Table identifier, column key and sorting direction to the tds-table-body component, can also be listened to in order to implement custom-sorting logic. | `CustomEvent<{ tableId: string; columnKey: string; sortingDirection: "desc" \| "asc"; }>` |
```
mJarsater, line 23: Would be great to solve this linter issue.

packages/core/src/components/table/table-header-cell/table-header-cell.tsx
alexandereneroth: Yea, good catch! 
I believe this change should be excluded from the PR if we don't want the PR hanging around until a possible future 2.0 version.

We could instead introduce a prop for the user to change the default sorting direction, for example.

Wadzabi: I changed this as I felt the current behaviour felt unintended and confusing. Currently on first click after render the sorting goes to descending. But when clicking for the first time on another column it goes to ascending. This only changes the behaviour for the first ever sort click. 

I would also find it strange if this causes braking changes since it's only changing the order of the events. Would only be a braking change if someone has tightly coupled their logic with the first event being dispatched being descending. 

From a feature standpoint I could see it being even better if you could set value yourself in the case that you have a default sorting direction on render. If I am not wrong there is currently no way to load the table with a sorting arrow already visible.

alexandereneroth: @alfredskeden @theJohnnyMe We should look into this.

packages/core/src/components/table/table-header-cell/readme.md
alfredskeden: It is when we build, So should be fixed with a commit? isnt it building this readme correct?

mJarsater: Yes, but I don't think we have specified the "correct" order in our lint config. So if someone has other settings locally it gonna be overridden

theJohnnyMe:
Couple of cents from my side:

1. Fix for selection on first column and making it "asc" is a good one, I think it makes sense to have it like that as all other columns start with asc.
2. Removal of hover icon which indicates sorting and replacing it with just "one way" icon has to be checked and accepted by designers. We can not move forward without designer input. It will also require a change in Figma. 
3. Once we have designer input we can proceed further if/when/how to release it.
4. The issue with linter that always changes asc/desc in the readme file is something to look into in a separate ticket.

Wadzabi:
Hi tegel team, any updates on when this will be reviewed by a designer?

charlottelorensson:
Hello Martin! 

Thank you for your work and the contribution! :) We had a look at the whole flow and were thinking of making some additional changes as well. We though that from the suggested new behavior and the old one, there are a few things we could consider changing as well:

1. Keeping/adding a neutral state. Currently you cannot “get out” of the sorting mode. Introducing a 3rd state/click, that would lead you back to sorting arrows that appear on hover, could solve this. This would suggest that we keep the sorting icons and therefore:

2. Instead of having the Arrow icon visible on hover we could have the Sorting icon appear on hover (also from the very first time you hover, we don't currently have this). Link to behavior example: https://carbondesignsystem.com/components/data-table/usage/

3. Looking into changing so that we do not directly enter the “Hover on selected” state after pressing but rather just selected and then introducing hover on selected the second time you hover.

We are still in planning but we will try and have a look when we can pick this up. It probably won't be anytime soon unfortunately so, if you have the time and resources, feel free to take a stab at it :)

theJohnnyMe:
Closing it due to inactivity.

PR #454, opened 2023-12-07
fix(slider): update on value should not send a tdsChange event

alfredskeden:
**Describe pull-request**  
From the support channel - "From a quick glance the tds slider is the only input component that dispatches event when changing the property from the outside. TdsSlider has a watch on value property and is then disptaching the event. This is not the case in for example text-field or text-area"

This moves makes so that changing the value of value does not trigger onTdsChange updates

**Solving issue**  
Fixes: [CDEP-3056](https://tegel.atlassian.net/browse/CDEP-3056)

**How to test**  
1. Go to storybook -> slider
2. Updating the value manually in <tds-slider value="50"> in the inspector to something should not fire a tdsChange event! 

**Suggested test steps**
- [x] Events

[CDEP-3056]: https://tegel.atlassian.net/browse/CDEP-3056?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/slider/slider.tsx
```diff
@@ -168,7 +168,9 @@ export class TdsSlider {
 
   @Watch('value')
   handleValueUpdate(newValue: string) {
-    this.forceValueUpdate(newValue);
+    this.calculateThumbLeftFromValue(newValue);
```
mJarsater, line 171: I think this makes a lot of sense. But could it be a breaking change?

alfredskeden: @mJarsater Nice that it makes sense to you, had some talk about it in the support channels and it could be a breaking change if someone updates the `value` outside and of slider and expects a tdsChange to fire. Likelihood of that is very small and something that they hopefully will notice. But since this more correlates to how all the inputs should work and how other components work it should be updated. We just have con communicate it in a good way.

theJohnnyMe (approved):
We will inform the users about this inconsistent behaviour in release notes with !important tag. In a way it is a breaking change, but hopefully no one relyed on this one.

timrombergjakobsson (approved):

PR #461, opened 2023-12-11
test(header): Adding unit tests

alfredskeden:
**Describe pull-request**  
Unit tests for header component

**Solving issue**  
Fixes: [CDEP-3003](https://tegel.atlassian.net/browse/CDEP-3003)

**How to test**  
1. Checkout code
2. `npm run test` the core package to check that tests run

[CDEP-3003]: https://tegel.atlassian.net/browse/CDEP-3003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/header/test/default/header.e2e.ts-snapshots/tds-header-default-renders-default-header-correctly-1-linux.png
alfredskeden, line 1: I think it is weird that playwright does not get the brand icon on the screenshot for some reason. I tried finding a solution for it, but not able to find one :(

alfredskeden: I found that footer has the same problem in the "brand".

I think it has to do with that we load the image as a background-image and it uses a var()? to get it?

nathalielindqvist (approved):
Apart from the missing brand symbol it looks good to me :)

packages/core/src/components/header/test/default/header.e2e.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  test('Header title exists and is "Example: default"', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const headerComponentHeaderText = page.getByText('Example: default');
+    await expect(headerComponentHeaderText).toBeVisible();
+  });
+
+  test('Header luncher icon exists', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const headerComponentLuncherIcon = page.getByRole('img');
```
alexandereneroth, line 24: Could this also select the brand icon?
Perhaps it would be good to combine this locator with another, like the button label or a test-id

packages/core/src/components/header/test/default/header.e2e.ts
```diff
@@ -0,0 +1,35 @@
...
+    await expect(headerComponentLuncherIcon).toBeVisible();
+    await expect(headerComponentLuncherIcon).toHaveCount(1);
+  });
+
+  test('Header brand label with link exists', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const headerComponentBrandLink = page.getByRole('link');
```
alexandereneroth, line 31: I guess this locator could unintentionally select links in the header-launcher as well if the implementation of that was be changed so that the launcher list is not removed from the DOM while it is hidden.

packages/core/src/components/header/test/default/header.e2e.ts-snapshots/tds-header-default-renders-default-header-correctly-1-linux.png
alexandereneroth: Yea, could be related to the relative path used in the CSS file to find the image. That path might not be correct anymore when it is fetched from that html file?

alexandereneroth: NVM its not relative its on the CDN

packages/core/src/components/header/test/default/header.e2e.ts
alfredskeden: It can not since brand icon is an css background-image

packages/core/src/components/header/test/default/header.e2e.ts
alfredskeden: Yes then this test changes and has to be updated by that changer of that logic?.

alfredskeden: Yes updated: [2d0bc34](https://github.com/scania-digital-design-system/tegel/pull/461/commits/2d0bc348527c6ace24db0439ac34cbdf3c0d206b)

packages/core/src/components/header/test/default/header.e2e.ts
alfredskeden: updated: [2d0bc34](https://github.com/scania-digital-design-system/tegel/pull/461/commits/2d0bc348527c6ace24db0439ac34cbdf3c0d206b)

theJohnnyMe (approved):

PR #462, opened 2023-12-11
test(footer): Adding unit tests

alfredskeden:
**Describe pull-request**  
Adding unit tests for footer component

**Solving issue**  
Fixes: [CDEP-3004](https://tegel.atlassian.net/browse/CDEP-3004)

**How to test**  
1. Checkout branch
2. Check so that `npm run tests` goes thru

[CDEP-3004]: https://tegel.atlassian.net/browse/CDEP-3004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/footer/test/default/footer.e2e.ts-snapshots/tds-footer-default-renders-default-footer-correctly-1-linux.png
alfredskeden, line 1: Same problem here as with header.
The brand background-image is not rendering on screenshot

packages/core/src/components/footer/test/default/footer.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+  test('Footer contains copyright text', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const footerCopyrightText = page.getByText('Copyright © 2023 Scania');
+    await expect(footerCopyrightText).toBeVisible();
+  });
+
+  test('Footer contains brand label (Scania)', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const footerBrandText = page.getByText('Scania', { exact: true });
+    await expect(footerBrandText).toBeHidden();
```
alfredskeden, line 37: I'm checking is hidden here because the p tag is hidden for accesability I guess and then a brand with an background-image should be show here but it should still be there but hidden, the text "Scania" that is.

packages/core/src/components/footer/test/default/footer.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+  test('Footer contains four links "Link text"', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const footerLinks = page.getByText('Link text');
+    await expect(footerLinks).toHaveCount(4);
+  });
+
+  test('Footer contains three truck icons', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const footerTruckIcons = page.getByRole('img');
+    await expect(footerTruckIcons).toHaveCount(3);
```
alexandereneroth, line 25: The page.getByRole('img'); locator could also find the Scania logo (if the implementation was different) and then I believe this test would fail.

alfredskeden: hmm not sure I follow, might need a talk about this because if eels like you have been thinking about something that I don't follow fully on.

timrombergjakobsson: I guess he means that the test is brittle in the sense that if the Scania logo was implemented as a image, the test would fail?

alexandereneroth: Yes, i mean that then the getByRole('img') would find two elements, i believe

alfredskeden: Hmm ok, I'll rethink the locator.

alfredskeden: How about these? @alexandereneroth @timrombergjakobsson [dd8b96e](https://github.com/scania-digital-design-system/tegel/pull/462/commits/dd8b96ea1df9b59e5e063702a740de430255af69)

theJohnnyMe: I think this works OK. Nice workaround :)

PR #466, opened 2023-12-12
fix(message): refactor unit tests

alfredskeden:
**Describe pull-request**  
Refactor message unit tests to have less implementation details but same tests.


nathalielindqvist (approved):
🍰

theJohnnyMe:
Error one has 4 steps in it, while other ones (info, warning, success) have just one test inside. Why is that?

alfredskeden:
> Error one has 4 steps in it, while other ones (info, warning, success) have just one test inside. Why is that?

The test result wont change it outcome based on the variant you put in it, So we only have to do tests in one tds-message to cover all of the "variants" and not repeat them for each folder.

Then just each folder check for screenshot if it would change we got it there.

theJohnnyMe:
> > Error one has 4 steps in it, while other ones (info, warning, success) have just one test inside. Why is that?
> 
> The test result wont change it outcome based on the variant you put in it, So we only have to do tests in one tds-message to cover all of the "variants" and not repeat them for each folder.
> 
> Then just each folder check for screenshot if it would change we got it there.

Get you! ;)

theJohnnyMe (approved):

PR #471, opened 2023-12-18
test(modal): adding unit tests

timrombergjakobsson:
**Describe pull-request**  
Addin unit tests to Modal

**Solving issue**  
Fixes: [CDEP-2973](https://tegel.atlassian.net/jira/software/projects/CDEP/boards/1?assignee=63ff6d6c314f5088138187f0&selectedIssue=CDEP-2973)

[CDEP-2973]: https://tegel.atlassian.net/browse/CDEP-2973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/modal/test/default/modal.e2e.ts-snapshots/tds-modal-renders-modal-correctly-1-linux.png
timrombergjakobsson, line 1: Hm, I see now that it is not showing?

mJarsater: Modal needs to be set to `show="true"`, no?

packages/core/src/components/modal/test/default/index.html
```diff
@@ -0,0 +1,26 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0" />
+    <link rel="stylesheet" href="../../../../../dist/tegel/tegel.css" />
+    <script type="module">
+      import { defineCustomElements } from '../../../../../loader/index.es2017.js';
+      defineCustomElements();
+    </script>
+  </head>
+
+  <body>
+    <tds-modal header="This is a header" data-testid="tds-modal-testid" size="lg">
```
mJarsater, line 15:
```suggestion
    <tds-modal header="This is a header" data-testid="tds-modal-testid" size="lg" show>
```

packages/core/src/components/modal/test/open-close/modal.e2e.ts
```diff
@@ -0,0 +1,55 @@
...
+    const tdsModal = page.locator('tds-modal');
+    const hasShowAttribute = await tdsModal.getAttribute('show');
+    if (hasShowAttribute) {
+      await expect(tdsModal).toBeVisible(); // Modal is initially open
+      await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });
+    }
+  });
+
+  test('Clicking [Delete] button closes the modal', async ({ page }) => {
+    const deleteButton = page.getByText('Delete');
```
alfredskeden, line 21:
```suggestion
    const deleteButton = page.getByRole('button').filter({ hasText: /Delete/ });
```

I think this test makes sure its a button as expected and also that the text is inside of it. So the developer dont change it to a clickable span or something :D

packages/core/src/components/modal/test/open-close/modal.e2e.ts
```diff
@@ -0,0 +1,55 @@
...
+    await deleteButton.click(); // Click [Delete] button
+    const tdsModal = page.locator('tds-modal');
```
alfredskeden, line 23: You can also see if the button that you just clicked is still visible or not I think.


```
    const deleteButton = page.getByRole('button').filter({ hasText: /Delete/ });
    await expect(deleteButton).toBeVisible(); // button to be visible
    await deleteButton.click(); // Click [Delete] button
    await expect(deleteButton).toBeHidden(); // button is removed
```

packages/core/src/components/modal/test/open-close/index.html
```diff
@@ -0,0 +1,26 @@
...
+    <link rel="stylesheet" href="../../../../../dist/tegel/tegel.css" />
+    <script type="module">
+      import { defineCustomElements } from '../../../../../loader/index.es2017.js';
+      defineCustomElements();
+    </script>
+  </head>
+
+  <body>
+    <tds-button id="my-modal-button" data-testid="tds-modal-button-testid" text="Open Modal"></tds-button>
+    <tds-modal header="This is a header" data-testid="tds-modal-testid" size="lg" id="my-modal-button" show>
```
alfredskeden, line 16: Think it's good to test "selector" prop somewhere also.

packages/core/src/components/modal/test/open-close/modal.e2e.ts
```diff
@@ -0,0 +1,55 @@
...
+
+test.describe('tds-modal-open', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto(componentTestPath);
+  });
+
+  test('1. Modal is opened by default', async ({ page }) => {
+    const tdsModal = page.locator('tds-modal');
+    const hasShowAttribute = await tdsModal.getAttribute('show');
+    if (hasShowAttribute) {
```
alfredskeden, line 14: I'm not sure we need to use conditionals here.

Additionally if tds-modal will always have this show attribute if you set it in the index.html tag so it does not give a fair representation of a good test, lets say for example a developer would still be able to make the element have the attribute show without showing the modal.

Instead I would remove the getAttribute show and look for the modal and see if it has all its visible elements

packages/core/src/components/modal/test/open-close/modal.e2e.ts
```diff
@@ -0,0 +1,55 @@
...
+  test('[Open Modal] button exists', async ({ page }) => {
+    const openModalButton = page.getByTestId('tds-modal-button-testid');
+    await expect(openModalButton).toBeVisible(); // [Open Modal] button exists and is visible
+  });
+
+  test('Clicking [Open Modal] button opens the modal', async ({ page }) => {
+    const openModalButton = page.getByTestId('tds-modal-button-testid');
+    const tdsModal = page.locator('tds-modal');
+
+    await openModalButton.dispatchEvent('click');
```
alfredskeden, line 36: After removing this click in the test it still comes back success. Because after clicking the button the modal should close right? Since its open by default.

I think this tests needs to be where the modal is not open by default to be tested. Also instead of getting the Open modal button like that since it would not be a <button> element we need to get the element inside tds-button with

```
const openModalButton = page.getByRole('button').filter({ hasText: /Open Modal/ });
```

packages/core/src/components/modal/test/prevent-true/index.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Badge - Basic</title>
```
alfredskeden, line 5:
```suggestion
    <title>Modal - prevent=true</title>
```

packages/core/src/components/modal/test/default/modal.e2e.ts
```diff
@@ -0,0 +1,25 @@
+import { test } from 'stencil-playwright';
+import { expect } from '@playwright/test';
+
+const componentTestPath = 'src/components/modal/test/default/index.html';
+
+test.describe('tds-modal', () => {
```
alfredskeden, line 6:
```suggestion
test.describe('tds-modal-default', () => {
```

packages/core/src/components/modal/test/default/modal.e2e.ts
```diff
@@ -0,0 +1,25 @@
...
+  test('renders modal correctly', async ({ page }) => {
+    await page.goto(componentTestPath);
+    const tdsModal = page.getByTestId('tds-modal-testid');
+    await expect(tdsModal).toHaveCount(1);
+    await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });
+  });
+
+  test('modal contains two buttons', async ({ page }) => {
```
alfredskeden, line 14: modal actually contains 3 buttons. also the close button right?

So something like

```
  test('modal contains three buttons', async ({ page }) => {
    // Locate the buttons within the modal
    const modalButtons = page.getByRole('button');

    // Assert that there are three buttons
    await expect(modalButtons).toHaveCount(3);
  });
```

Note that this only finds the elements if the modal is show by default

timrombergjakobsson: True!

packages/core/src/components/modal/test/open-close/modal.e2e.ts
timrombergjakobsson: I'm not following you? Maybe the modal should be initially hidden? Wondering what should be the correct structure and approach. Maybe hidden by default and take a screenshot to verify? And then test to verify that it opens and one for closing.

alfredskeden: Yes I think "default" can be open by default.
"Open - Close" tests can be hidden by default and then does all the tests to check that the button opens and closes the modal.

timrombergjakobsson: Yes, I think so too. Although, removing the show attribute and then trying to test that clickin the open modal button and that the modal should be visible, always fails?

`test('Clicking [Open Modal] button opens the modal', async ({ page }) => {
    const openModalButton = page.getByRole('button').filter({ hasText: /Open Modal/ });
    const tdsModal = page.locator('tds-modal');

    await openModalButton.click();
    await expect(tdsModal).toBeVisible(); // Modal should be open after clicking [Open Modal]
  });`
  
  Then I just get:
  
  ` Locator: locator('tds-modal')
    Expected: visible
    Received: `hidden`

alfredskeden: Yes the user don't know what a `page.locator('tds-modal')` so could either, take a screenshot to verify or to check for an element inside the modal to see if that is visible?

Im not 💯 why tds-modal is still hidden when its clearly should be visible. Mby it just dosnt know if it is somehow I havnt tested it myself.

alfredskeden: It works for me with the above code if I add the selector="#my-modal-button" to the code

timrombergjakobsson: > Did you add the selector to the modal in the index.html? you can also open the index.html on `Live Server` extension to see it and test it.
> 
> Edit: It works for me with the above code if I add the selector="#my-modal-button" to the code

Ah yea? Hm, I have it in index.html, but still it fails?

alfredskeden: Hmm strange, Then I don't know :(

packages/core/src/components/modal/test/default/modal.e2e.ts
```diff
@@ -0,0 +1,25 @@
...
+    await page.goto(componentTestPath);
+    const tdsModal = page.getByTestId('tds-modal-testid');
+
```
theJohnnyMe, line 17: [Unit test scenario](https://scaniaazureservices-my.sharepoint.com/:x:/r/personal/tatyana_bumazhkina_scania_com/Documents/Unit%20test%20scenarios.xlsx?d=wca48bf6d92cf4e8c8f456337e2002ec4&csf=1&web=1&e=64N7Zh&nav=MTVfezIyRjA3ODQ0LUIwRkItNDIzNS05MDk4LTIwOUM4ODU1Q0MwNn0) suggests testing the Header and Body test too, can that be added too?

packages/core/src/components/modal/test/default/modal.e2e.ts
```diff
@@ -0,0 +1,25 @@
+import { test } from 'stencil-playwright';
```
theJohnnyMe, line 1: This test generates a second screenshot image once you branch out and run it on the local machine.
I guess it is due to some naming changes.
Please delete one that is not needed.

packages/core/src/components/modal/test/open-close/modal.e2e.ts
```diff
@@ -0,0 +1,55 @@
...
+    await expect(tdsModal).toBeVisible(); // Modal should be open after clicking [Open Modal]
+  });
+
+  test('Clicking [Close] button closes the modal', async ({ page }) => {
+    const closeButton = page.locator('button.tds-modal-close');
```
theJohnnyMe, line 41:
```suggestion
    const closeButton = page.locator('button[aria-label="close"]');
```

Just as idea to avoid class names...

packages/core/src/components/modal/test/open-close/modal.e2e.ts
theJohnnyMe: <tds-modal needs to have `selector` prop that targets button's ID. I can see on my local that it is missing.

packages/core/src/components/modal/test/open-close/index.html
```diff
@@ -0,0 +1,26 @@
...
+      <span slot="body">
+        'Where you can put anything you want!'
+        <span slot='actions' class='tds-u-flex tds-u-gap2'>
+          <tds-button data-dismiss-modal size="md" text="Delete" variant="danger"></tds-button>
+          <tds-button data-dismiss-modal size="md" text="Cancel" variant="secondary"></tds-button>
+        </span>
```
theJohnnyMe, line 22:
```suggestion
      <span slot="body">
        'Where you can put anything you want!'
        </span>
        <span slot='actions' class='tds-u-flex tds-u-gap2'>
          <tds-button data-dismiss-modal size="md" text="Delete" variant="danger"></tds-button>
          <tds-button data-dismiss-modal size="md" text="Cancel" variant="secondary"></tds-button>
        </span>         
```

Closing tag for body slot's span.

packages/core/src/components/modal/test/prevent-true/index.html
```diff
@@ -0,0 +1,28 @@
...
+
+  <body>
+    <tds-button id="my-modal-button" data-testid="tds-modal-button-testid" text="Open Modal"></tds-button>
+    <tds-modal header="This is a header" data-testid="tds-modal-testid" size="lg" id="my-modal-button" show prevent="true">
+      <span slot="body">
+        'Where you can put anything you want!'
+        <span slot='actions' class='tds-u-flex tds-u-gap2'>
+          <tds-button data-dismiss-modal size="md" text="Delete" variant="danger"></tds-button>
+          <tds-button data-dismiss-modal size="md" text="Cancel" variant="secondary"></tds-button>
+        </span>
```
theJohnnyMe, line 25:
```suggestion
      <span slot="body">
        'Where you can put anything you want!'
        </span>
        <span slot='actions' class='tds-u-flex tds-u-gap2'>
          <tds-button data-dismiss-modal size="md" text="Delete" variant="danger"></tds-button>
          <tds-button data-dismiss-modal size="md" text="Cancel" variant="secondary"></tds-button>
        </span>         
```

Missing closing span for body slot.

theJohnnyMe (approved):

nathalielindqvist (approved):

PR #476, opened 2023-12-29
fix(dropdown, text-field, textarea): global css bleeding in label and helper text

theJohnnyMe:
**Describe pull-request**  
there is a background color bleeding from global styles in dropdown, text-field and textarea. It is noticible when components have lable outside or helper text.

**Solving issue**  
_In case of GitHub issue add # plus the number of the issue (for example #123) OR if it is Azure then AB# and number of ticket_
Fixes: [CDEP-2186](https://tegel.atlassian.net/browse/CDEP-2186)

**How to test**  
1. Go to dropdown/text-field/textarea
2. Change Dark/Light modes, toggle between mode variants.
3. Background of label outside and helper text is now transparent.

**Additional context**  
Issues comes from rule in global css file. It is viisble in components that beside main part of component have other neighbouring elements that are under same Host. When those fragments appear above or belo main part of component then bleeding issue is visible. 

[CDEP-2186]: https://tegel.atlassian.net/browse/CDEP-2186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/global/global.scss
```diff
@@ -99,3 +99,20 @@ body {
     background-color: var(--tds-grey-900);
   }
 }
+
+//FIXME: In Tegel 2.0 reconsider usage of lines 68 - 100 in this file.
+
+/* Overwrite of global rules that bleed in some components */
+:root,
```
mJarsater, line 106: Can we not do this in each individual file instead? I think this might be make debugging harder in the future.

theJohnnyMe: I agree, however, the import of that file would have to remain at the bottom of this one as it overrides behavior specified a couple of lines ahead. In previous projects, we usually named it "_shame.scss" to reflect that we are not proud to have it and also to indicate that we need to fix it future. 

Please let me know if you think of a better name.

mJarsater: Ah I see what you mean. You could set the background in the `:host` within each file, but I think you would have to throw an `!important` on it. For example: 

Screenshot 2024-01-08 at 13 28 07

I guess it's personal preference if you'd rather go with your solution above and fix it for 2.0. Or with the `!important` and fix that for 2.0.

theJohnnyMe: Yeah, exactly. I was also checking the !important rule but we might forget about it, or if ever in future we need to overright it will be chaos. 
I will prepare a separate file as maybe in the future we might need some more "dirty & ugly" code we can place it there until we reach a point for Tegel 2.0, when part of the improvements will be getting rid of shame.scss - meaning fixing the cause of issues rather than patching it.

packages/core/src/components/footer/test/default/footer.e2e.ts
```diff
@@ -27,7 +27,7 @@ test.describe('tds-footer-default', () => {
 
   test('Footer contains copyright text', async ({ page }) => {
     await page.goto(componentTestPath);
-    const footerCopyrightText = page.getByText('Copyright © 2023 Scania');
+    const footerCopyrightText = page.getByText('Copyright © 2024 Scania');
```
mJarsater, line 30:
```suggestion
    const footerCopyrightText = page.getByText(`Copyright © ${new Date().getFullYear()}  Scania`);
```

Not sure if this works in the testing environment, but would be nice to not have to update this every year.

theJohnnyMe: Does work Sir, I just removed extra space ;) Thanks for your help man!

mJarsater (approved):

nathalielindqvist (approved):
Solves the bug, nice work 👍

timrombergjakobsson (approved):

PR #477, opened 2024-01-09
feat(icon): enable Storybook change of title and description

theJohnnyMe:
**Describe pull-request**  
Improvement of SVG title and description text

**Solving issue**  
Fixes: [CDEP-3086](https://tegel.atlassian.net/browse/CDEP-3086)

**How to test**  
1. Go to the preview link
2. Change SVG title from Storybook controls 
3. Hover over the icon for a longer period
4. The same text should appear in a tooltip
5. Make text empty
6. Hover over the icon for a longer period
7. The icon name should be visible in the tooltip 

**Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

**Additional context**  
There are more ways to improve it, but it might be more work and we might force users to specify description prop too. If both title and descriptions are available, we might implement advice from here: [SVG A11Y](https://css-tricks.com/accessible-svgs/#aa-2-inline-svg).
Another way around is that we always specify title and description by default (probably it would be the same text) if user does not provide anything under svg-title and svg-desc props. Please let me know what you think.


mJarsater (approved):
LGTM

nathalielindqvist (commented):
Nice! How do we want to handle decorative icons? This solution enforces an SVG title on all icons, including decorative ones.

nathalielindqvist:
> > Nice! How do we want to handle decorative icons? This solution enforces an SVG title on all icons, including decorative ones.
> 
> I think that's covered by adding the `element.name` as a title if non is submitted: `<title>{this.svgTitle ??`${element.name} icon`}</title>`
> 
> I think this is a sensible fallback, another alternative would be to only add the title is an `svgTitle` is submitted, otherwise we treat it as purely decorative and add no title, and no aria-labelledby.

I agree that it's a sensible fallback if we accept that there will be some sort of tooltip text regardless. From what I could see the guidelines only say decorative icons don't _have_ to have a title/alt text, but guessing there is nothing wrong with having one anyway?

mJarsater:
> > > Nice! How do we want to handle decorative icons? This solution enforces an SVG title on all icons, including decorative ones.
> > 
> > 
> > I think that's covered by adding the `element.name` as a title if non is submitted: `<title>{this.svgTitle ??`${element.name} icon`}</title>`
> > I think this is a sensible fallback, another alternative would be to only add the title is an `svgTitle` is submitted, otherwise we treat it as purely decorative and add no title, and no aria-labelledby.
> 
> I agree that it's a sensible fallback if we accept that there will be some sort of tooltip text regardless. From what I could see the guidelines only say decorative icons don't _have_ to have a title/alt text, but guessing there is nothing wrong with having one anyway?

Yeah I think that something you'll have to decide internally. Setting the `title` based on the `element.name` is the way the icon has worked before. The `svgTitle` was just introduce a simple override, which was requested in the support channel. 

I don't think there is any a11y issue in setting the `title` as `element.name`. Either way is fine, removing it might be considered a breaking change though.

packages/core/src/components/icon/icon.tsx
```diff
@@ -10,15 +10,19 @@ import { iconsCollection } from './iconsArray';
...
 
-  /** Pass a size of icon as a string, for example: 32px, 1rem, 4em... */
+  /** Pass a size of icon as a string, for example, 32px, 1rem, 4em... */
   @Prop({ reflect: true }) size: string = '16px';
 
   /** Override the default title for the svg. */
   @Prop() svgTitle: string;
 
+  /** Set description for the svg. */
+  @Prop() svgDesc: string;
```
mJarsater, line 24:
```suggestion
  @Prop() svgDesc?: string;
```

This should probably be optional, since it's not required in our render function.

theJohnnyMe: Nice one!

packages/core/src/components/icon/icon.tsx
```diff
@@ -10,15 +10,19 @@ import { iconsCollection } from './iconsArray';
...
+  /** Pass the name of the icon.
+   * For icon names, refer to Storybook Icon controls dropdown or https://tegel.scania.com/foundations/icons/icon-library */
   @Prop({ reflect: true }) name: string = 'truck';
 
-  /** Pass a size of icon as a string, for example: 32px, 1rem, 4em... */
+  /** Pass a size of icon as a string, for example, 32px, 1rem, 4em... */
   @Prop({ reflect: true }) size: string = '16px';
 
   /** Override the default title for the svg. */
   @Prop() svgTitle: string;
```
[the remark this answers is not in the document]
theJohnnyMe: Nice one! ;)

theJohnnyMe:
1. Check the spelling on the description 
2. use svgDescription instead of svgDesc
3. provide explanation on what svgDescription is used for

mJarsater (approved):

nathalielindqvist (approved):
LGTM!

timrombergjakobsson (approved):
Nice!

PR #478, opened 2024-01-09
fix(dropdown): empty no-result-text prop

theJohnnyMe:
**Describe pull-request**  
If the user uses lazy loading, there is a bug in which we display a "No results" message for a short time while user logic is fetching results. If the user sets noResultsText to an empty string now, it wont trigger no results message to pop up.
The second fix is to send an error in the console if no tds-dropdown option is available once users interact with the component.

**Solving issue**  

Fixes: [CDEP-2974](https://tegel.atlassian.net/browse/CDEP-2974)

**How to test**  
1. Go to the preview link.
2. Enable filtering
3. Try to input a query that will not match any data.
4. No result message should pop up.
5. Delete "No result text" from Storybook control.
6. Try to search again for a query that will not match any data.
7. The message "No result" should NOT pop up.
8. Branch on you local
9. Drop all <tds-dropdown-option> from story file.
10. Interact with component. Nothing should happen.
11. Take a look at to console, there should be an error message saying that the dropdown is missing data.

**Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

[CDEP-2947]: https://tegel.atlassian.net/browse/CDEP-2947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CDEP-2974]: https://tegel.atlassian.net/browse/CDEP-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -65,7 +65,8 @@ export class TdsDropdown {
   /** Normalizes input text for fuzzier search */
   @Prop() normalizeText: boolean = true;
 
-  /** Text that is displayed if filter is used and there are no options that matches the search. */
+  /** Text that is displayed if filter is used and there are no options that matches the search.
+   * Setting it to an empty string disables message from showing up. */
   @Prop() noResultText: string = 'No result';
```
mJarsater, line 70:
```suggestion
  @Prop() noResultText?: string;
```

Might be a better change to make this optional and undefined if it's not set. And then in your new check on line 573 would instead be:

```
{this.filterResult === 0 && this.noResultText  && (
```

Would be breaking though.

theJohnnyMe: Yep, dropping the message text and keeping it empty would be a breaking change as I think some of the users would get confused. However, I will add this to Tegel 2.0 to reconsider how we control showing the "No results" message.

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -340,11 +341,17 @@ export class TdsDropdown {
...
-    Array.from(this.host.children).filter(
+  /* Returns a list of all children that are tds-dropdown-option elements */
+  private getChildren = () => {
+    const tdsDropdownOptions = Array.from(this.host.children).filter(
       (element) => element.tagName === 'TDS-DROPDOWN-OPTION',
     ) as Array<HTMLTdsDropdownOptionElement>;
+    if (tdsDropdownOptions.length === 0) {
+      console.error(
+        'TDS DROPDOWN: Data is missing! Please check if you have <tds-dropddown-option> elements!',
+      );
```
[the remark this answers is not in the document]
theJohnnyMe: Yep, I can even change it from error to warning, as you say, it might jump into various scenarios when it is not an error.

packages/core/src/components/dropdown/dropdown.tsx
mJarsater: Great, should be made optional no though I think?

theJohnnyMe:
Message type updated to warning + unit test added for case: no-result-test being empty means missing dropdown message with "No result".

packages/core/src/components/dropdown/test/filter/hide-no-result-message/dropdown.e2e.ts
mJarsater, line 1: I don't know what other tests we have in place for this component. But could rename this to `no-result-message` and test for both when a no-result-message should show (when the prop is set) and when it shouldn't show (when no value is set)?

theJohnnyMe: In "normalize" test case we check if "No result" is there, do you think we should add a separate test just for it?

mJarsater: Ah nice! Yeah I thinks that's fine 👍🏻

mJarsater (approved):

packages/core/src/components/dropdown/dropdown.tsx
theJohnnyMe: Just added "?" to it ;)

timrombergjakobsson (approved):
Looks nice!

PR #480, opened 2024-01-11
fix(dropdown): fix selected value when erased

timrombergjakobsson:
**Describe pull-request**  
When filter is set to true and you have selected a value from dropdown by typing it. Before there was a bug that when you start erasing selected value and then type it again to select it, it would not be selected. This fixes that bug. 

**Solving issue**  
Fixes: [CDEP-3077](https://tegel.atlassian.net/jira/software/projects/CDEP/boards/1?assignee=63ff6d6c314f5088138187f0&selectedIssue=CDEP-3077)

**How to test**  
1. Go to Storybook
2. Check in Dropdown, set Filter to true
3. Select a value from dropdown by typing it, and then start erasing the selected value, when typing it again to select it, it should get selected. 

[CDEP-3077]: https://tegel.atlassian.net/browse/CDEP-3077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


theJohnnyMe (approved):
LGTM

packages/core/src/components/dropdown/dropdown.tsx
mJarsater, line 1: Might be worth adding a test to avoid regression on this one.

timrombergjakobsson: Agree! We decided to add it afterwards!

nathalielindqvist (approved):
LGTM

PR #484, opened 2024-01-16
test(link): added unit test

theJohnnyMe:
**Describe pull-request**  
Adding unit tests for Link component

**Solving issue**  
Fixes: [CDEP-3012](https://tegel.atlassian.net/browse/CDEP-3012)

**How to test**  
1. Check [unit test spreadsheet for Link component](https://scaniaazureservices-my.sharepoint.com/:x:/r/personal/tatyana_bumazhkina_scania_com/Documents/Unit%20test%20scenarios.xlsx?d=wca48bf6d92cf4e8c8f456337e2002ec4&csf=1&web=1&e=zOXTNJ&nav=MTVfe0VCNDI3Rjc2LUEzMTQtNEM4Ri1CQkMyLUMxMzE2MTY3NTJDN30).
2. Make sure unit tests in PR do the checks written in the form above 
3. Run code on your local to see that tests pass


nathalielindqvist (approved):
Looks good! Just a thought about checking if the component is disabled. You probably know this better than I do, but is there a standardized way you check for it across components? Just thinking that not all disabled components have the cursor event 'none', or is that fixed now? Maybe we could check for the attribute 'disabled' instead in that case? Just a though, this works too :)

packages/core/src/components/link/test/disabled/link.e2e.ts
```diff
@@ -0,0 +1,43 @@
+import { test } from 'stencil-playwright';
+import { expect } from '@playwright/test';
+
+const componentTestPath = 'src/components/link/test/disabled/index.html';
+
+test.describe('tds-link-default', () => {
+  test('is disabled link rendered correctly', async ({ page }) => {
```
mJarsater, line 7:
```suggestion
  test('disabled link rendered correctly', async ({ page }) => {
```

timrombergjakobsson: Or maybe even:
```suggestion
  test('disabled link is rendered correctly', async ({ page }) => {
```
?

packages/core/src/components/link/test/default/link.e2e.ts
```diff
@@ -0,0 +1,53 @@
+import { test } from 'stencil-playwright';
+import { expect } from '@playwright/test';
+
+const componentTestPath = 'src/components/link/test/default/index.html';
+
+test.describe('tds-link-default', () => {
+  test('is default link rendered correctly', async ({ page }) => {
+    await page.goto(componentTestPath);
+
+    const tdsLink = page.getByTestId('tds-link-testid');
```
timrombergjakobsson, line 10: Should we use `page.getByRole('link');` instead?

mJarsater: I don't think so. That will target the slotted `<a>` no? Not the `<tds-link>`.

theJohnnyMe: @timrombergjakobsson  - I was thinking in first place of that, but tds-link component is not more than a wrapped that enforces styles on anchor tag that user is passing in, so I had to relay on testid.

timrombergjakobsson (approved):

PR #485, opened 2024-01-17
feat(table-header-cell): added slot

theJohnnyMe:
**Describe pull-request**  
Adding a slot to the table-header-cell component to enable users to use it for content like HTML in more complex use cases, like filtering over a single column or adding an icon to the title or when passing content between parenthesis. 

**Solving issue**  
Fixes: [CDEP-3092](https://tegel.atlassian.net/browse/CDEP-3092)

**How to test**  
1. Open a branch on your local
2. Try to add icon or any other HTML code between the parenthesis of <tds-table-header-cell>
3. Your component should be shown in story file (in case you test in Storybook)

**Additional context**  
I made other props optional by adding "?", does not change anything on the user side, but it is a small Typescript improvement. 

[CDEP-3092]: https://tegel.atlassian.net/browse/CDEP-3092?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/table/table-header-cell/table-header-cell.tsx
+++ b/packages/core/src/components/table/table-header-cell/table-header-cell.tsx
@@ -19,6 +19,10 @@ const relevantTableProps: InternalTdsTablePropChange['changed'] = [
   'verticalDividers',
 ];
 
+/**
+ * @slot <default> - <b>Unnamed slot.</b> For the cell contents.
+ */
+
 @Component({
   tag: 'tds-header-cell',
   styleUrl: 'table-header-cell.scss',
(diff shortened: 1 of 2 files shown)


packages/core/src/components/table/table-header-cell/table-header-cell.tsx
```diff
@@ -201,6 +208,7 @@ export class TdsTableHeaderCell {
         <button class="tds-table__header-button" onClick={() => this.sortButtonClick()}>
           <span class="tds-table__header-button-text" style={{ textAlign: this.textAlignState }}>
             {this.cellValue}
+            <slot />
```
mJarsater, line 211: Great stuff. Should we update the CSS to style this element also? I think not, the slots to me are for advanced users and we should not dabble in their stuff.

timrombergjakobsson (approved):
:shipit:

nathalielindqvist (approved):
Nice! Works well!

packages/core/src/components/table/table-header-cell/table-header-cell.tsx
theJohnnyMe: I agree, I think styles should not be part of it. Do you see some of styles that should be removed?

PR #492, opened 2024-01-26
fix(dropdown): adding check for value

timrombergjakobsson:
**Describe pull-request**  
Fixing errors in console when choosing in dropdown when filter is not set. Also included fix for problem with clicking chevron and opening dropdown. 

**Solving issue**  
Fixes: [CDEP-3098](https://tegel.atlassian.net/jira/software/projects/CDEP/boards/1?selectedIssue=CDEP-3098)

**How to test**  
1. Go to Storybook
2. Check in dropdown
3. Select a value from dropdown, make sure that no error is logged in web inspector console.

[CDEP-3098]: https://tegel.atlassian.net/browse/CDEP-3098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -146,7 +146,9 @@ export class TdsDropdown {
     if (this.filter) {
       this.inputElement.value = this.getValue();
     } else {
-      this.inputElement.value = selection.length > 0 ? selection[0].label : '';
+      if (this.inputElement) {
+        this.inputElement.value = selection.length > 0 ? selection[0].label : '';
+      }
```
theJohnnyMe, line 151: As we talked, this line of code seems to be skipped. I tried putting a console.log in condition, but it never comes up.
We need to find another way to resolve it.

nathalielindqvist (approved):

PR #496, opened 2024-01-30
feat(text-field): adding number type and min-max attributes

gokberknur:
**Describe pull-request**  

Currently from our application can pass type=number attribute to tegel-text field, but we cant define min, max attributes from our-side. That effects how user can interact with increase/decrease buttons inside the input field.

**Solving issue**  
Adding support for  customisable min, max values for the input type number.

**How to test**  
1. Go to text-field storybook
2. select input type number
3. define min,max values,
4. interact with increment/decrement buttons

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

**Screenshots**  
_If applicable, add screenshots to help explain_
Screenshot 2024-01-30 at 14 14 20
Screenshot 2024-01-30 at 14 14 03

**Additional context**  
_Add any other context about the pull-request here._

### DIFF
--- a/packages/core/src/components/text-field/readme.md
+++ b/packages/core/src/components/text-field/readme.md
@@ -14,7 +14,9 @@
 | `helper`        | `helper`         | Helper text                                 | `string`                              | `undefined`  |
 | `label`         | `label`          | Label text                                  | `string`                              | `''`         |
 | `labelPosition` | `label-position` | Position of the label for the Text Field.   | `"inside" \| "no-label" \| "outside"` | `'no-label'` |
+| `max`           | `max`            | Max allowed value for input type number     | `number \| string`                    | `undefined`  |
 | `maxLength`     | `max-length`     | Max length of input                         | `number`                              | `undefined`  |
+| `min`           | `min`            | Min allowed value for input type number     | `number \| string`                    | `undefined`  |
 | `modeVariant`   | `mode-variant`   | Mode variant of the Text Field              | `"primary" \| "secondary"`            | `null`       |
 | `name`          | `name`           | Name property                               | `string`                              | `''`         |
 | `noMinWidth`    | `no-min-width`   | Unset minimum width of 208px.               | `boolean`                             | `false`      |
@@ -22,7 +24,7 @@
 | `readOnly`      | `read-only`      | Set input in readonly state                 | `boolean`                             | `false`      |
 | `size`          | `size`           | Size of the input                           | `"lg" \| "md" \| "sm"`                | `'lg'`       |
 | `state`         | `state`          | Error state of input                        | `"default" \| "error" \| "success"`   | `'default'`  |
-| `type`          | `type`           | Which input type, text, password or similar | `"password" \| "text"`                | `'text'`     |
+| `type`          | `type`           | Which input type, text, password or similar | `"number" \| "password" \| "text"`    | `'text'`     |
 | `value`         | `value`          | Value of the input text                     | `string`                              | `''`         |
 
 
--- a/packages/core/src/components/text-field/text-field.stories.tsx
+++ b/packages/core/src/components/text-field/text-field.stories.tsx
@@ -50,9 +50,31 @@ export default {
       control: {
         type: 'radio',
       },
-      options: ['Text', 'Password'],
+      options: ['Text', 'Password', 'Number'],
       table: {
         defaultValue: { summary: 'text' },
+      },
+    },
+    min: {
+      name: 'Min',
+      description: 'Minumum acceptable value when input type is number',
+      control: {
+        type: 'number',
+      },
+      if: {
+        arg: 'type',
+        eq: 'Number',
+      },
+    },
+    max: {
+      name: 'Max',
+      description: 'Maximum acceptable value when input type is number',
+      control: {
+        type: 'number',
+      },
+      if: {
+        arg: 'type',
+        eq: 'Number', 
       },
     },
     size: {
@@ -181,6 +203,8 @@ export default {
     prefixType: 'Icon',
     suffix: false,
     suffixType: 'Icon',
+    min: "0",
+    max: "10",
     maxLength: 0,
     noMinWidth: 'Default',
     readonly: false,
@@ -192,6 +216,8 @@ const Template = ({
   modeVariant,
   state,
   type,
+  min,
+  max,
   size,
   label,
   labelPosition,
@@ -207,6 +233,8 @@ const Template = ({
   disabled,
 }) => {
   const maxlength = maxLength > 0 ? `max-length="${maxLength}"` : '';
+  min = min  ? `min="${min}"` : '';
+  max = max  ? `max="${max}"` : '';
   const stateValue = state.toLowerCase();
   const sizeLookUp = {
     Large: 'lg',
@@ -233,6 +261,8 @@ const Template = ({
       label-position="${labelPosition.toLowerCase()}"
       ${helper ? `helper="${helper}"` : ''}
       ${maxlength}
+      ${min}
+      ${max}
       ${disabled ? 'disabled' : ''}
       ${readonly ? 'read-only' : ''}
       ${noMinWidth ? 'no-min-width' : ''}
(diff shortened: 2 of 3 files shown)


gokberknur:
I havent used storybook before, so I am not sure if everything setup correctly. But feel free to correct them.

theJohnnyMe (approved):
Really nice addition to Text-field component @gokberknur !
I like how you added conditional rendering of min/max controls in Storybook in case type "number" is used. 
I was thinking we could even add email to the list of types as I feel people might need it to rely on HTML validation, but it can be part of some other PR. 

Awesome work and thanks for the contribution.
One more approval is needed from our team and then it can be merged ;)

packages/core/src/components/text-field/text-field.stories.tsx
```diff
@@ -50,11 +50,33 @@ export default {
...
       },
-      options: ['Text', 'Password'],
+      options: ['Text', 'Password', 'Number'],
       table: {
         defaultValue: { summary: 'text' },
       },
     },
+    min: {
+      name: 'Min',
+      description: 'Minumum acceptable value when input type is number',
```
nathalielindqvist, line 60:
```suggestion
      description: 'Minimum acceptable value when input type is number',
```

nathalielindqvist (approved):
Great stuff, thank you for the contribution @gokberknur! Just a super tiny comment on one of the descriptions, otherwise looks great! 🎉

gokberknur:
> Great stuff, thank you for the contribution @gokberknur! Just a super tiny comment on one of the descriptions, otherwise looks great! 🎉

aah, thanks great catch!

timrombergjakobsson (approved):
:shipit:

mJarsater:
Just playing the devils advocate here a bit. Textfield type="number" to me has some confusion about it. I think you should consider if it wouldn't a better approach to create a `tds-number-field` or even `tsd-input` that accepts either number, text, email, etc.

packages/core/src/components/text-field/readme.md
```diff
@@ -14,15 +14,17 @@
...
 | `name`          | `name`           | Name property                               | `string`                              | `''`         |
 | `noMinWidth`    | `no-min-width`   | Unset minimum width of 208px.               | `boolean`                             | `false`      |
 | `placeholder`   | `placeholder`    | Placeholder text                            | `string`                              | `''`         |
 | `readOnly`      | `read-only`      | Set input in readonly state                 | `boolean`                             | `false`      |
 | `size`          | `size`           | Size of the input                           | `"lg" \| "md" \| "sm"`                | `'lg'`       |
 | `state`         | `state`          | Error state of input                        | `"default" \| "error" \| "success"`   | `'default'`  |
-| `type`          | `type`           | Which input type, text, password or similar | `"password" \| "text"`                | `'text'`     |
+| `type`          | `type`           | Which input type, text, password or similar | `"number" \| "password" \| "text"`    | `'text'`     |
```
mJarsater, line 27: Left comment in PR but to me it's bit confusing to have text-field type="number".

mJarsater: If we'd separate it into another component we wouldn't need to have put the caveat that `min`/`min` only is valid for type `number`

mJarsater: See: https://github.com/scania-digital-design-system/tegel/pull/498#issuecomment-1920893280

gokberknur:
> Just playing the devils advocate here a bit. Textfield type="number" to me has some confusion about it. I think you should consider if it wouldn't a better approach to create a `tds-number-field` or even `tsd-input` that accepts either number, text, email, etc.
> 
> See: [#498 (comment)](https://github.com/scania-digital-design-system/tegel/pull/498#issuecomment-1920893280)

Very valid point.. I would say tds-input maybe could be better name. because tegel uses html input tag under the hood. I don't know if there is a specific reason why it is named like tds-textfield because as you stated, it is a very explicit name for "text". 

So it is up to you guys, how you would like to move forward. What is the cost of changing selector from tds-textfield to tds-input? @theJohnnyMe

mJarsater:
> > Just playing the devils advocate here a bit. Textfield type="number" to me has some confusion about it. I think you should consider if it wouldn't a better approach to create a `tds-number-field` or even `tsd-input` that accepts either number, text, email, etc.
> > See: [#498 (comment)](https://github.com/scania-digital-design-system/tegel/pull/498#issuecomment-1920893280)
> 
> Very valid point.. I would say tds-input maybe could be better name. because tegel uses html input tag under the hood. I don't know if there is a specific reason why it is named like tds-textfield because as you stated, it is a very explicit name for "text".
> 
> So it is up to you guys, how you would like to move forward. What is the cost of changing selector from tds-textfield to tds-input? @theJohnnyMe

I guess we can't change the name of the element before tegel 2.0 since it would be breaking. What you could do is introduce a new component called `tds-input` which can live alongside `tds-text-field` until it can be removed (at 2.0). I think one thing to take into consideration is that I think we should stay away from type specific props, which only works for one type and not for the others. For that reason it might be better to separate the components altogether

gokberknur:
> > > Just playing the devils advocate here a bit. Textfield type="number" to me has some confusion about it. I think you should consider if it wouldn't a better approach to create a `tds-number-field` or even `tsd-input` that accepts either number, text, email, etc.
> > > See: [#498 (comment)](https://github.com/scania-digital-design-system/tegel/pull/498#issuecomment-1920893280)
> > 
> > 
> > Very valid point.. I would say tds-input maybe could be better name. because tegel uses html input tag under the hood. I don't know if there is a specific reason why it is named like tds-textfield because as you stated, it is a very explicit name for "text".
> > So it is up to you guys, how you would like to move forward. What is the cost of changing selector from tds-textfield to tds-input? @theJohnnyMe
> 
> I guess we can't change the name of the element before tegel 2.0 since it would be breaking. What you could do is introduce a new component called `tds-input` which can live alongside `tds-text-field` until it can be removed (at 2.0). I think one thing to take into consideration is that I think we should stay away from type specific props, which only works for one type and not for the others. For that reason it might be better to separate the components altogether

hmmm yeah.. but input tag has different type attributes. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

so it seems the naming choice "tds-textfield" is the thing here, unless you have reason I think "tds-input" would be better for clear naming to end user, so they know it is an input field and they can pass attributes as they do in native element.

 I don't know if there was a specific case  so you had to name it textfield (in sdds and tegel). 
 For me it was confusing what to choose textfield or textarea for example in the beginning , where I started to work with tegel :D 
 
But on the other-hand some design systems separating number and text that like you explained. having 2 separate components for numbers and textfields (like microsoft fast or  ibm carbon: https://carbondesignsystem.com/components/number-input/usage/#behaviors) 

Maybe it is a better idea to have something like this "sdds-number-input", where increment/decrement buttons are customized to look like "Scania". 

Either way, i think naming of (tds-textfield) can be good to discuss. (e.g; tds-text-input) or maybe we can ask in the channel?  (if you havent discuss this topic before obviously).

Some UI libraries doesnt do this seperation and using just single component. (like angular material, <mat-input> ) for me this looks better, but there might be solid reasoning why microsoft and ibm did other-way around, since i am not expert in the area, so i trust you whatever you choose :) 
 
@theJohnnyMe

theJohnnyMe:
@gokberknur and @mJarsater  - I support idea of separating concerns, but in this case we might create a component that is 99% same as tds-text-field, so we end up having a lot of code duplication. Sonar Cloud also warns about it. 
I understand that tds-text-fields is not a best component name for one that can accept text, number, password etc... but for me it is a UX discussion that needs to happen. I am all for to change its name in tegel 2.0 to tds-input as it is then agnostic to type... it could potentially merge date-time one into it too, as only data type is different, rest stays the same. 

Personally I was thinking to add type email too to text-field + enable valid prop + enable pattern etc...  

Long story short

1.  I would propose closing this PR
2. Merging Gökberks one
3. Make a ticket for Tegel 2.0 to make tds-input that cover current data types and even new ones.

PR #498, opened 2024-02-01
feat: number input

mJarsater:
Example of number-field over text-field type number.

This is just a suuuuper quick (mostly copy-pasted from text-field) variant of a `tds-number-field` component. I personally think this is a better approach than using the `tds-text-field` with `type="number"`. Easier to understand the API, better type safety, separation of concerns and so on.

Improvement area: 
- [ ] Remove all unnecessary props in number-field.
- [ ] Remove unnecessary CSS.
- [ ] Couple the CSS vars of the text-field and number-field by agnostic name
- [ ] Remove number type from `text-field`
For example: 
```
  --tds-number-field-background-primary: var(--tds-grey-50);
```
and 
```
  --tds-text-field-background-primary: var(--tds-grey-50);
```
are the same, so should probably only need one: 
```
  --tds-input-field-background-primary: var(--tds-grey-50);
```

### DIFF
--- a/packages/core/src/components/icon/readme.md
+++ b/packages/core/src/components/icon/readme.md
@@ -31,6 +31,7 @@
  - [tds-message](../message)
  - [tds-modal](../modal)
  - [tds-navigation-tabs](../tabs/navigation-tabs)
+ - [tds-number-field](../number-field)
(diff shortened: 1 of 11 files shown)


theJohnnyMe:
Closing this one in favour of https://github.com/scania-digital-design-system/tegel/pull/496

PR #500, opened 2024-02-05
feat(datetime): setValue method

theJohnnyMe:
**Describe pull-request**  
Enabling setting "value" to another value during the runtime.

**Solving issue**  
Fixes: [CDEP-3110](https://tegel.atlassian.net/browse/CDEP-3110)

**How to test**  
1. Open the preview link
2. Navigate to Datetime component
3. From the console, try to use setValue method on input by setting new value like: component.setValue('2000-01-13T14:15') 

**Additional context**  
Sometimes you need to find element in "Element" tab in Chrome before using console as sometimes targeting element ends in "null". At least that happened to me. 

[CDEP-3110]: https://tegel.atlassian.net/browse/CDEP-3110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/datetime/datetime.tsx
+++ b/packages/core/src/components/datetime/datetime.tsx
@@ -1,4 +1,4 @@
-import { Component, State, h, Prop, Listen, Event, EventEmitter } from '@stencil/core';
+import { Component, State, h, Prop, Listen, Event, EventEmitter, Method } from '@stencil/core';
 
 @Component({
   tag: 'tds-datetime',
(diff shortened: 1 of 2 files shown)


EmelieLitwin (approved):
Works for me! 👍 
Don't forget to update docs with new message 🙂

timrombergjakobsson (approved):

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -82,6 +82,12 @@ export class TdsDatetime {
   })
   tdsFocus: EventEmitter<FocusEvent>;
 
+  /** Setting a new value of the datetime element */
```
mJarsater, line 85:
```suggestion
  /** Method that sets the value of the datetime element */
```

PR #508, opened 2024-02-14
fix(header): flash of dropdown on page reload

timrombergjakobsson:
**Describe pull-request**  
When page is loaded a quick flash of the dropdown can be seen. To prevent the flash of content before the dropdown is clicked, we can conditionally render the slot content based on the open state.
**Solving issue**  
Fixes: [CDEP-3118](https://tegel.atlassian.net/jira/software/projects/CDEP/boards/1?assignee=63ff6d6c314f5088138187f0&selectedIssue=CDEP-3118)

**How to test**  
1. Go to Storybook
2. Check in Navigation -> User menu
3. Reload the page and see that flash does not happen. To compare with the bug, it is present on demo pages and in our Storybook in prod. 

[CDEP-3118]: https://tegel.atlassian.net/browse/CDEP-3118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathalielindqvist (approved):
Nice catch! 🎣

theJohnnyMe (approved):
Really nice and neat solution! LGTM

PR #513, opened 2024-02-19
fix(slider): emit tdsChange on mouse-up and touch-end events

theJohnnyMe:
**Describe pull-request**  
Slider component spam users as we emit tdsChange on every move of the thumb. That made users use timeout functions to prevent it. The solution is to emit tdsChange once the user lifts the finger from the mouse or there is an end of touch (for touch devices).

**Solving issue**  
Fixes: [CDEP-3199](https://tegel.atlassian.net/browse/CDEP-3199)

**How to test**  
1. Open the preview Storybook of this branch and the production version of Storybook.
2. Open inspector and observe Console.
3. Move the thumbs of the slider left and right. Production one emits events all the time, while in this PR it happens only once the user releases the mouse button or stops touch.
4. Also, check behavior for snapping points, increase via +/- buttons as well as input. 

- [x] All existing tests pass

**Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

**Additional context**  
There is noticed flickering of the Thumb once going slowly towards the left (decreasing). Behavior is observed in both prod and development. There is another ticket for it. It is not noticeable when using Slider for a range of 0 - 1000 for example.

### DIFF
--- a/packages/core/src/components/slider/readme.md
+++ b/packages/core/src/components/slider/readme.md
@@ -29,9 +29,10 @@
 
 ## Events
 
-| Event       | Description                                 | Type                              |
-| ----------- | ------------------------------------------- | --------------------------------- |
-| `tdsChange` | Sends the value of the slider when changed. | `CustomEvent<{ value: string; }>` |
+| Event       | Description                                                                                      | Type                              |
+| ----------- | ------------------------------------------------------------------------------------------------ | --------------------------------- |
+| `tdsChange` | Sends the value of the slider when changed. Fires after mouse up and touch end events.           | `CustomEvent<{ value: string; }>` |
+| `tdsInput`  | Sends the value of the slider while moving the thumb. Fires on mouse move and touch move events. | `CustomEvent<{ value: string; }>` |
 
 
 ## Methods
--- a/packages/core/src/components/slider/slider.stories.tsx
+++ b/packages/core/src/components/slider/slider.stories.tsx
@@ -254,12 +254,21 @@ const Template = ({
     
      <!-- Script tag for demo purposes -->
     <script>
-      slider = document.querySelector('tds-slider')
+      const slider = document.querySelector('tds-slider');
       
-      slider.removeEventListener('tdsChange', null)
-      slider.addEventListener('tdsChange', (event) => {
-        console.log(event);
-      });
+      function handleTdsChange(event) {
+        console.log('tdsChange event:', event);
+      }
+      
+      slider.removeEventListener('tdsChange', handleTdsChange);
+      slider.addEventListener('tdsChange', handleTdsChange);
+
+      function handleTdsInput(event) {
+        console.log('tdsInput event:', event);
+      }
+
+      slider.removeEventListener('tdsInput', handleTdsInput);
+      slider.addEventListener('tdsInput', handleTdsInput);
     </script>
   `);
 
--- a/packages/core/src/components/slider/slider.stories.tsx
+++ b/packages/core/src/components/slider/slider.stories.tsx
@@ -254,7 +254,7 @@ const Template = ({
     
      <!-- Script tag for demo purposes -->
     <script>
-      const slider = document.querySelector('tds-slider');
+      slider = document.querySelector('tds-slider');
       
       function handleTdsChange(event) {
         console.log('tdsChange event:', event);
--- a/packages/core/src/components/slider/slider.tsx
+++ b/packages/core/src/components/slider/slider.tsx
@@ -303,6 +303,11 @@ export class TdsSlider {
     const inputElement = event.target as HTMLInputElement;
     let newValue = parseInt(inputElement.value);
 
+    // Check if the new value is different from the current value
+    if (newValue === parseInt(this.value)) {
+      return; // Exit the function if the new value is the same as the current value
+    }
+
     if (newValue < parseFloat(this.min)) {
       newValue = parseFloat(this.min);
     } else if (newValue > parseFloat(this.max)) {
@@ -558,6 +563,9 @@ export class TdsSlider {
                   readOnly={this.readOnly}
                   onBlur={(event) => this.updateSliderValueOnInputChange(event)}
                   onKeyDown={(event) => this.handleInputFieldEnterPress(event)}
+                  type="number"
+                  min={this.min}
+                  max={this.max}
                 />
               </div>
             </div>
--- a/packages/core/src/components/slider/slider.tsx
+++ b/packages/core/src/components/slider/slider.tsx
@@ -83,7 +83,7 @@ export class TdsSlider {
 
   private resizeObserverAdded: boolean = false;
 
-  /** Sends the value of the slider when changed. */
+  /** Sends the value of the slider when changed. Fires after mouse up and touch end events. */
   @Event({
     eventName: 'tdsChange',
     composed: true,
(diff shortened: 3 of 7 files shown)


mJarsater (commented):
Just asking, but if someone is using the slider to let say increase the volume or zoom in/out would they not want the event of every tick? Might be better to create separate events to cater to both use cases?

theJohnnyMe:
> Just asking, but if someone is using the slider to let say increase the volume or zoom in/out would they not want the event of every tick? Might be better to create separate events to cater to both use cases?

That is a good idea. I think I will add the `tdsInput` event to reflect immediate changes aka while still there is a touch event or mouse button is pressed. What do you think?

mJarsater:
> > Just asking, but if someone is using the slider to let say increase the volume or zoom in/out would they not want the event of every tick? Might be better to create separate events to cater to both use cases?
> 
> That is a good idea. I think I will add the `tdsInput` event to reflect immediate changes aka while still there is a touch event or mouse button is pressed. What do you think?

Yeah that sounds good, seems like that's the way the native `input type="range"` does it. 

Input: https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event

Change:https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event

packages/core/src/components/slider/slider.stories.tsx
```diff
@@ -254,12 +254,21 @@ const Template = ({
     
      <!-- Script tag for demo purposes -->
     <script>
-      slider = document.querySelector('tds-slider')
+      const slider = document.querySelector('tds-slider');
```
mJarsater, line 257: this will error in the console if you navigate from the docs and back, can't use const in the code preview since we don't know how ofter the code will run:

https://github.com/scania-digital-design-system/tegel/assets/62651103/80734d31-179d-479b-961c-1e37f82c5c6e

packages/core/src/components/slider/readme.md
```diff
@@ -29,9 +29,10 @@
 
 ## Events
 
-| Event       | Description                                 | Type                              |
-| ----------- | ------------------------------------------- | --------------------------------- |
-| `tdsChange` | Sends the value of the slider when changed. | `CustomEvent<{ value: string; }>` |
+| Event       | Description                                                                                      | Type                              |
+| ----------- | ------------------------------------------------------------------------------------------------ | --------------------------------- |
+| `tdsChange` | Sends the value of the slider when changed. Fires after mouse up and touch end events.           | `CustomEvent<{ value: string; }>` |
```
mJarsater, line 34: Since the `tdsChange` used to run in every change (input) this would be a breaking change, no?

packages/core/src/components/slider/slider.stories.tsx
theJohnnyMe: Good catch!

packages/core/src/components/slider/readme.md
theJohnnyMe: Indeed it is, but I will do my best to speak to users in the announcements channel about this one. Pretty much those that do not need "realtime" data, can keep using tdsChange. For those that want exactly what tdsChange has done so far, they need to use tdsInput instead.

mJarsater: Alright, I have no idea where you are at for 2.0. But if this absolutely can't wait for that then go for it 👍🏻

nathalielindqvist (commented):
When changing value of the Slider using the input control the TdsChange event fires twice, is this by design? :)

theJohnnyMe:
> When changing value of the Slider using the input control the TdsChange event fires twice, is this by design? :)

Most probably not, probably it is an issue in the component so I will try to check it out. Thanks for raising it ;)

mJarsater:
> https://pr-513.d3fazya28914g3.amplifyapp.com

I can't replicate this issue: 

https://github.com/scania-digital-design-system/tegel/assets/62651103/30376de5-9ef4-4640-8b37-7315d16c0986

theJohnnyMe:
> > https://pr-513.d3fazya28914g3.amplifyapp.com
> 
> I can't replicate this issue:
> 
>  Screen.Recording.2024-02-26.at.10.57.41.mov

I find this correct. When the user navigates the slider with the keyboard, change happen immediately as button is pressed while with the mouse and touch use can slide left and right while having the mouse button presses or uninterrupted touch, once mouse button is released or finger is lifted from the screen that is the moment the value is passed in tdsChange.

What do you think?

theJohnnyMe:
@mJarsater - I think I misunderstood you and you misunderstood Nathalie :D 

I think this is the issue Nathalie speaks of: 

Screenshot 2024-02-27 at 10 12 40

theJohnnyMe:
@mJarsater and @nathalielindqvist, improvement is done in this [commit](https://github.com/scania-digital-design-system/tegel/pull/513/commits/437946597fab1125bd31f159acc8bc42d247ac53).

Please check it, test it and let me know how it looks :)

nathalielindqvist (approved):
Nice! LGTM 🔥

packages/core/src/components/slider/slider.tsx
```diff
@@ -121,33 +132,33 @@ export class TdsSlider {
...
     this.thumbInnerElement.classList.remove('pressed');
-    this.updateValue();
+    this.updateValue(event);
 
     this.trackElement.focus();
   }
 
   @Listen('touchend', { target: 'window' })
-  handleTouchEnd() {
+  handleTouchEnd(event: TouchEvent) {
```
timrombergjakobsson, line 148: Not very critical, but Sonarcloud reports duplication on these two methods. Maybe something to refactor in another PR?
https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=513&id=scania-digital-design-system_tegel&open=AY3FsqZABUMWrAS_OREb

Maybe into a one method to hold the shared logic and the other ones can just call that one?

```
@Listen('mouseup', { target: 'window' })
handleMouseUp() {
  this.handleRelease();
}

@Listen('touchend', { target: 'window' })
handleTouchEnd() {
  this.handleRelease();
}

// New method that contains the shared logic
private handleRelease() {
  if (!this.thumbGrabbed) {
    return;
  }

  this.thumbGrabbed = false;
  this.thumbInnerElement.classList.remove('pressed');
  this.updateValue();
  this.trackElement.focus();
}
```

theJohnnyMe: Good call, I will check if I can sort it out in this one. Should not be biggy...

theJohnnyMe: Change pushed ;)

timrombergjakobsson:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=scania-digital-design-system_tegel&pullRequest=513) **Quality Gate failed**
> Failed conditions  [B Reliability Rating on New Code](https://sonarcloud.io/dashboard?id=scania-digital-design-system_tegel&pullRequest=513) (required ≥ A)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=scania-digital-design-system_tegel&pullRequest=513)
> 
> idea Catch issues before they fail your Quality Gate with our IDE extension SonarLint [SonarLint](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

Sonarlint seems to be configured for react/jsx syntax? 
`Unknown property 'class' found, use 'className' instead`
We can look at those things later on.

mJarsater (approved):
Nice, make sure to clarify this change in the announcement since it is technically breaking

PR #465, opened 2023-12-11
feat: dropdown fixes

alexandereneroth:
The idea is to explore fixing [CDEP-2998](https://tegel.atlassian.net/browse/CDEP-2998) in this one. (Clicking on an item in the header dropdown doesn't close it.)

First some cleaning up of the logic had to be done.

Not sure if "fixing" this would be considered a breaking change, so i added a new prop for it instead.

This PR contains a failing test that needs to pass before this is merged.

[CDEP-2998]: https://tegel.atlassian.net/browse/CDEP-2998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #517, opened 2024-02-20
fix(header-launcher): null check

timrombergjakobsson:
**Describe pull-request**  
Adding a check for tds-header-launcher-list. 
If an element exists, we should add a class to the launcher list enabling it to take all available height on the screen.

**Solving issue**  
We get an console.error that it is null. 
Fixes: [CDEP-3126](https://tegel.atlassian.net/browse/CDEP-3126)

**How to test**  
1. Go to Storybook
2. Check in the Header and see that there is no error in the console.
3. Check this story `/?path=/story/patterns-navigation--few-navigation-items` in preview. It does not have tds-header-launcher-list, so launcher takes as much height is needed to present all items, but not more. In other stories tds-header-launcher-list is present so launcher takes 100% of available height. 

See screenshot for error:
Skärmavbild 2024-02-19 kl  16 51 46

**Aditional notes**
It seems we complicated way of looking for element so I tried to simplify it. 

[CDEP-3126]: https://tegel.atlassian.net/browse/CDEP-3126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/header/header-launcher/header-launcher.tsx
+++ b/packages/core/src/components/header/header-launcher/header-launcher.tsx
@@ -34,20 +34,8 @@ export class TdsHeaderLauncher {
   }
 
   componentDidLoad() {
-    const slotElement = this.host.shadowRoot.querySelector('slot:not([name])') as HTMLSlotElement;
-
-    if (slotElement) {
-      const slottedElements = slotElement.assignedElements();
-      const hasListTypeMenu = slottedElements.some(
-        (element) => element.tagName.toLowerCase() === 'tds-header-launcher-list',
-      );
-
-      if (hasListTypeMenu) {
-        this.hasListTypeMenu = true;
-      }
-    } else {
-      console.warn('Slot element not found or assigned elements.');
-    }
+    const hasListTypeMenu = !!this.host.querySelector('tds-header-launcher-list');
+    this.hasListTypeMenu = hasListTypeMenu;
   }
 
   toggleLauncher() {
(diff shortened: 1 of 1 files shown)


timrombergjakobsson:
> When browsing stories under Patterns -> Navigation I am getting warnings per story there. Is this expected? What should be done not to have this number of warnings?
> 
> Screenshot 2024-02-20 at 15 10 25

Yeah, we need to look into that, cos now also, `console.warn('Slot element not found or assigned elements.');` that I added hits. So something is incorrect in how we structured the component?

nathalielindqvist (commented):
Looks good but I agree with Johnnys comment - should the user always get a warning if they're not using the slot?

theJohnnyMe (dismissed):
Solution needs to be improve. Iit seems that too many warnings or errors we emit now in regular usage of component.

timrombergjakobsson:
> Solution needs to be improve. Iit seems that too many warnings or errors we emit now in regular usage of component.

You have a suggestion :)?

theJohnnyMe:
> > Solution needs to be improve. Iit seems that too many warnings or errors we emit now in regular usage of component.
> 
> You have a suggestion :)?

Not at the moment, I would need time to sit and understand the issue more. Just saw there are a lot of warnings popping up in the console in regular use of components.

timrombergjakobsson:
> > > Solution needs to be improve. Iit seems that too many warnings or errors we emit now in regular usage of component.
> > 
> > 
> > You have a suggestion :)?
> 
> Not at the moment, I would need time to sit and understand the issue more. Just saw there are a lot of warnings popping up in the console in regular use of components.

Yeah, Il have a look as well.

theJohnnyMe:
I am not sure if there is anything that can help, but worth checking Alexes old PR: https://github.com/scania-digital-design-system/tegel/pull/465

nathalielindqvist (approved):
LGTM! Nice work! 🏆

timrombergjakobsson (commented):
Nice one @theJohnnyMe!

theJohnnyMe (approved):

PR #521, opened 2024-02-21
fix(modal): align with figma

EmelieLitwin:
**Describe pull-request**  
Some paddings didn't match figma and storybook story didn't match either

**Solving issue**  
_Please link to the issue this pull request is solving by adding ticket number after CDEP-_
Fixes: [CDEP-2617](https://tegel.atlassian.net/browse/CDEP-2617)

**How to test**  
1. Go to storybook
2. Compare with ui library

- [x] All existing tests pass

[CDEP-2617]: https://tegel.atlassian.net/browse/CDEP-2617?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/modal/modal.scss
```diff
@@ -92,20 +92,12 @@ $modals: (
 
 @mixin modal-header {
   display: flex;
-  padding-bottom: var(--tds-spacing-element-8);
+  padding: var(--tds-spacing-element-8);
+  padding: 16px;
```
nathalielindqvist, line 96: Do we need both padding declarations? I feel like they might just override each other, but maybe I'm missing something? :)

packages/core/src/components/modal/modal.tsx
```diff
@@ -151,7 +151,7 @@ export class TdsModal {
         <div class="tds-modal-backdrop" />
         <div class={`tds-modal tds-modal__actions-${this.actionsPosition} tds-modal-${this.size}`}>
           <div class="header">
-            {this.header && <div class="header">{this.header}</div>}
+            {this.header && <div class="header-text">{this.header}</div>}
```
nathalielindqvist, line 154: Asking because I don't know - was the class name updated only to not collide with the other class called "header" or is there another reason as well?

packages/core/src/components/modal/modal.scss
```diff
@@ -261,6 +246,12 @@ button.tds-modal-close {
     margin-left: var(--tds-spacing-element-48);
   }
 }
+@media (min-width: 736px) {
```
mJarsater, line 249: Feel like this pixel value should be turned into a variable?

theJohnnyMe:
@EmelieLitwin - What is the status of this ticket?

packages/core/src/components/modal/modal.scss
```diff
@@ -122,14 +113,9 @@ $modals: (
   color: var(--tds-modal-text);
   font: var(--tds-body-01);
   letter-spacing: var(--tds-body-01-ls);
-  padding-bottom: 40px;
   overflow-y: visible;
-  padding-right: 16px;
+  padding: 0 16px 16px;
```
mJarsater, line 117:
```suggestion
  padding: 0 var(--tds-spacing-element-16) var(--tds-spacing-element-16);
```

mJarsater: @EmelieLitwin @theJohnnyMe @nathalielindqvist @timrombergjakobsson I think is something that should be update in a lot of places. Preferring vars over pixel values

theJohnnyMe (changes requested):
There is a bug that is visible here and in prod.
Please try to fix it.

How to replicate it:
- add a long title in Header so it goes to two or more lines
- add quite a long text in the body
- select "Sticky" as actions buttons position

Result: Text in the body is cut off. User can not read all of it. See if it can be fixed.

packages/core/src/components/modal/modal.scss
```diff
@@ -147,7 +147,7 @@ $modals: (
     .body {
       font: var(--tds-body-01);
       letter-spacing: var(--tds-body-01-ls);
-      padding-bottom: 96px;
+      padding-bottom: 35vh;
```
theJohnnyMe, line 150: Be careful when using relative values, we might end up in a scenario like this as it is dependent on screen height.





Screenshot 2024-04-02 at 16 59 46


See issue is in how slots affect each other. If title goes to two or more lines, it affect body. Body is calculated with calc which seems to get confused if Title gets to push it. Please do play with flex, flex basis and flex shrink work, maybe it can help out here.

Check this article and see if we can use something like that on Modal component here: https://css-tricks.com/understanding-flex-grow-flex-shrink-and-flex-basis/
Maybe it is possible to use tricks form here, but with flex-orientation: column. 
Do play with it, it is a nice challenge :)

nathalielindqvist: Realized about an hour after I went home that this fix wouldn't work, gotta love those "wait a minute" moments in development 😅 I'll take a look at flex and see what I can do!

timrombergjakobsson: Good suggestion! Yes, most likely then the modal would be set to display: flex; so:
`
.tds-modal {
  display: flex;
  flex-direction: column;
  `
  kinda.

theJohnnyMe (approved):
Really nice improvement Nathalie ;)

timrombergjakobsson (approved):
Awesome work @nathalielindqvist!

PR #538, opened 2024-03-11
fix(toast): content width and close button margins

nathalielindqvist:
Removed the content ellipsis so header and subheader can both be unlimitedly long. Also adjusted the margin of the close button to align with component in Figma.

**Solving issue**  
Fixes: [CDEP-3081](https://tegel.atlassian.net/browse/CDEP-3081)

**How to test**  
1. Go to Storybook link below.
2. Check in Toast component.
3. Check that header and subheader can be very long without breaking into an ellipsis or pushing the text outside the component.
4. Compare with component in Figma to make sure it's aligned.

- [ ] All existing tests pass


packages/core/src/components/toast/toast.scss
```diff
@@ -102,10 +103,6 @@
       color: var(--tds-toast-subheadline-color);
       font: var(--tds-detail-02) !important;
       letter-spacing: var(--tds-detail-02-ls) !important;
-      max-width: 252px;
-      white-space: nowrap;
-      overflow: hidden;
-      text-overflow: ellipsis;
```
mJarsater, line 108: Might be good to highlight this in the announcement since this could be considered breaking. People might have a lot of content in their toasts which they expect to be truncated.

theJohnnyMe (approved):
Nice one @nathalielindqvist ! 🚀

PR #544, opened 2024-03-12
ci: release all script improvement

theJohnnyMe:
**Describe pull-request**  
Fix for release script 

**Solving issue**  
Fixing issues in the release "all script"

**How to test**  
1. Check if the code makes sense it is hard to test it without testing in prod

### DIFF
--- a/.github/workflows/release-all.yml
+++ b/.github/workflows/release-all.yml
@@ -26,38 +26,219 @@ on:
 jobs:
   release-core:
     runs-on: ubuntu-latest
-    outputs:
-      version: ${{ steps.core-release.outputs.version }}
+    env:
+      WORKING_DIRECTORY: packages/core
     steps:
-      - name: Trigger Core Release
-        id: core-release
-        uses: benc-uk/workflow-dispatch@v1
+      - name: Check out code
+        uses: actions/checkout@v4
         with:
-          workflow: release-core.yml
-          token: ${{ secrets.GITHUB_TOKEN }}
-          inputs: '{"nodeVersion": "${{ github.event.inputs.nodeVersion }}", "tags": "${{ github.event.inputs.tags }}", "dryRun": "${{ github.event.inputs.dryRun }}"}'
+          ref: develop
+
+      - name: Set up node
+        uses: actions/setup-node@v3
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-all.yml
```diff
@@ -26,38 +26,219 @@ on:
 jobs:
   release-core:
     runs-on: ubuntu-latest
+    env:
+      WORKING_DIRECTORY: packages/core
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+        with:
+          ref: develop
```
nathalielindqvist, line 35: When trying to do a manual release we encountered an error code that had to do with the inability to check out the branch on which we were currently on, develop in this case. I'm not sure exactly how this script is configured, but is that something we need to keen in mind?

.github/workflows/release-all.yml
```diff
@@ -26,38 +26,219 @@ on:
...
+        run: npm run build
+
+      - name: Core - Publish
+        working-directory: ${{ env.WORKING_DIRECTORY }}
+        run: |
+          if [ "${{ github.event.inputs.dryRun }}" == "true" ]; then
+            npm publish --tag ${{ github.event.inputs.tags }} --dry-run
+          else
+            npm publish --tag ${{ github.event.inputs.tags }}
+          fi
```
nathalielindqvist, line 69: What does "fi" stand for? Is it some sort of flag?

timrombergjakobsson: `fi` marks the end of the conditional block.

.github/workflows/release-all.yml
theJohnnyMe: The problem in the last release was in "Manual create release PR" which was trying to use develop as a reference and reset it before bumping the version number. Now what we did in that script it so pull the latest develop and create a release branch as the first step so there should not be issues. Tim and I have tested it so the script runs OK.

In this script, when it comes to real release, we are doing everything on develop branch, there are no changes to it made, just building and publishing so I hope it will run smooth.

timrombergjakobsson (approved):
I would say we go for it and try it out. It will only create an additional workflow and the existing is left intact.

nathalielindqvist (approved):
Let's try this out and see how it goes! 🤘

PR #552, opened 2024-03-15
feat(modal): re-initilize selector and clean-up method

theJohnnyMe:
**Describe pull-request**  
Modal component, when rendered dynamically and simultaneously using selector/reference has an issue. It looks like the reference is lost after the initial render. That can be observed if the user navigates to other pages and returns to one where the component is. On the initial visit of the page, Modal opens, on revisit - it does not. 

What I did is to have the initialization of selector/reference in a later and more repetitive lifecycle method. The user can call the same method in case the lifecycle fails, just as a fallback solution. An additional thing was adding the removal of listeners once the component is destroyed - just to be neat and prevent performance and memory issues. 

**Solving issue**  
Fixes: [CDEP-3063](https://tegel.atlassian.net/browse/CDEP-3063)

**How to test**  
1. Go check the code and see if it makes sense.
2. Current version fo code in develop is available [here](https://github.com/scania-digital-design-system/tegel-angular-demo/pull/87) to test. 
3. Version of code with this PR fixes is available [here](https://github.com/scania-digital-design-system/tegel-angular-demo/pull/85).
4. So try to visit the page, go to "Test page", open/close different modals. Navigate to another page and then come back to "Test page".
5. New code should make sure that Modla always open while in current one it open only on inital load, refresh is needed to make it work again after navigating to another pages.

- [x] All existing tests pass

[CDEP-3063]: https://tegel.atlassian.net/browse/CDEP-3063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -74,14 +74,43 @@ export class TdsModal {
     if (this.show !== undefined) {
       this.isShown = this.show;
     }
-    this.setDismissButtons();
-    this.setShowButton();
+    this.initializeModal();
 
     if (this.header && hasSlot('header', this.host)) {
       console.warn(
         "Tegel Modal component: Using both header prop and header slot might break modal's design. Please use just one of them. ",
       );
     }
+  }
+
+  componentWillLoad() {
+    this.initializeModal();
+  }
+
+  disconnectedCallback() {
+    this.cleanupModal();
+  }
+
+  /** Initializes or re-initializes the modal, setting up event listeners. */
+  @Method()
+  async initializeModal() {
+    this.setDismissButtons();
+    this.setShowButton();
+  }
+
+  /** Cleans up event listeners and other resources. */
+  @Method()
+  async cleanupModal() {
+    if (this.selector || this.referenceEl) {
+      const referenceEl = this.referenceEl ?? document.querySelector(this.selector);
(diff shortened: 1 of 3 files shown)


packages/core/src/components/modal/modal.tsx
```diff
@@ -84,6 +83,36 @@ export class TdsModal {
...
+
+  /** Initializes or re-initializes the modal, setting up event listeners. */
+  @Method()
+  async initializeModal() {
+    this.setDismissButtons();
+    this.setShowButton();
+  }
+
+  /** Cleans up event listeners and other resources. */
+  @Method()
```
mJarsater, line 102: Should be marked as `@internal` if it not something a user might use

theJohnnyMe: I am thinking that maybe users would need to use it in case something goes off with the lifecycle. I can not predict the case to be frank, but I would rather leave it possible to use it.

nathalielindqvist (approved):
Nice catch on this one! 🎣  The Modal works as expected with the new implementation - LGTM! 🌟

timrombergjakobsson (approved):
Nice!

Issue #522, opened 2024-02-21
[Bug report]: Text field read only with icon suffix


apparecidoo:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

@scania/tegel@1.6.0

### Browser

Chrome

### Framework

Angular

### Version

Angular 16

### Reproduction steps

1. Copy and paste the code example

### Code example

```
<tds-text-field type="text" size="lg" state="default" label="Label" label-position="no label" read-only no-min-width placeholder="Placeholder">
    <tds-icon slot="suffix" name="truck" size="20px"></tds-icon>
  </tds-text-field>
```

### Expected behaviour

Not override the icons or text

mJarsater:
Might be an expected behaviour from design? @theJohnnyMe 

I would recommend conditionally toggling the suffix icon if readonly is true?

theJohnnyMe:
Adding an image here so we understand what it is about:

![Screenshot 2024-03-12 at 15 15 24](https://github.com/scania-digital-design-system/tegel/assets/17004063/4669a34a-a0c8-4c3c-9b15-0269bae15018)

I am pinging our designers to check what is expected behavior here. We will keep you posted.

theJohnnyMe:
I got input from designers: Long story short -> "When using suffix icon in read-only state, read-only state icon takes priority and suffix icon should be hidden"
@apparecidoo: I did update in PR for that so you do not have to do anything atm, just be patient, I think we will push this out this Thursday in a release so once you upgrade your code you will have this change in too ;)

theJohnnyMe:
Closing the issue as PR is merged. Stayed tuned for the upcoming release.

PR #554, opened 2024-03-18
fix(text-field): read-only hides suffix icon

theJohnnyMe:
**Describe pull-request**  
When using the suffix icon and read-only, the read-only icon takes priority, and the suffix icon remains hidden. 

**Solving issue**  
Fixes: [CDEP-3182](https://tegel.atlassian.net/browse/CDEP-3182)
https://github.com/scania-digital-design-system/tegel/issues/522

**How to test**  
1. Add suffix icon
2. Turn on "read-only"
3. Suffx icon should be hidden
4. PS. Please check that other controls like size, mode-variant, and state are working too. I had to change from using string literals to using an object way of writing CSS as string literals had some hiccups when changing values in runtime. 

- [x] All existing tests pass

**Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

[CDEP-3182]: https://tegel.atlassian.net/browse/CDEP-3182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/text-field/text-field.tsx
+++ b/packages/core/src/components/text-field/text-field.tsx
@@ -26,10 +26,10 @@ export class TdsTextField {
   /** Label text */
   @Prop() label: string = '';
 
-   /** Min allowed value for input type number */
+  /** Min allowed value for input type number */
   @Prop() min: string | number;
 
-   /** Max allowed value for input type number */
+  /** Max allowed value for input type number */
   @Prop() max: string | number;
 
   /** Helper text */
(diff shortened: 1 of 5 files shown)


packages/core/src/components/text-field/text-field.tsx
```diff
@@ -203,7 +211,15 @@ export class TdsTextField {
...
 
           {usesSuffixSlot && (
-            <div class={`text-field-slot-wrap-suffix text-field-${this.state}`}>
+            <div
+              class={{
+                'text-field-slot-wrap-suffix': true,
+                'text-field-error': this.state === 'error',
+                'text-field-success': this.state === 'success',
+                'text-field-default': this.state === 'default',
+                'tds-u-display-none': this.readOnly,
```
mJarsater, line 220: This is the big change right? 

Are the class changes above for some other reason?

mJarsater: Could add this behaviour as part of the description for the `readOnly` and `slot="suffix"`

theJohnnyMe: @mJarsater  - Yes, when adding new class I realized we use string literal so I updated it to use object like structure that works better in realtime updates. Since I had to change it for lines you selected I felt it makes sense to update rest of the code. I wrote about it in description too.

theJohnnyMe: > Could add this behaviour as part of the description for the `readOnly` and `slot="suffix"`

Good idea, makes sense so we do not confuse users.

mJarsater (approved):

nathalielindqvist (commented):
Don't know if this should be part of this PR, but it looks like the suffix icon and the read-only icon have different margins/positioning when the components it put in Medium or Small size. Should that be handled somehow?
Medium:
image
image

Small:
image
image

theJohnnyMe:
> Don't know if this should be part of this PR, but it looks like the suffix icon and the read-only icon have different margins/positioning when the components it put in Medium or Small size. Should that be handled somehow? Medium: image image
> 
> Small: image image
Nice finding @nathalielindqvist  - Is it the case in prod version too or only in this PR ?

theJohnnyMe:
Seems like it is in prod too @nathalielindqvist. I will check it and if it is simple (it seems), I will fix it in this PR ;) Nice finding really ;)

theJohnnyMe:
@nathalielindqvist  - Change pushed! ;)

nathalielindqvist (approved):
Looks great! ⭐

PR #562, opened 2024-03-22
fix: create branch based on upcoming release

timrombergjakobsson:
**Describe pull-request**  
Attempt to fix release script problem. Yesterday's problem with minor/patch existing on remote might be solved with this.


theJohnnyMe (approved):
LGTM. Give it a shot after merge to see if it works expected. Test with both using "minor" and "patch" from version dropdown in GitHub Action.

nathalielindqvist (approved):
👌

PR #563, opened 2024-03-22
feat(tds-dropdown): added a clear filter button

finnlake:
### Snapshot tests fail since there now is a new button with cross icon where there is a dropdown with a filter.

**Added a clear filter button**  
_Added a clear filter button since users gets confused how to use the filter ie that they have to delete the existing text in the input._

**Solving issue**  
-

**How to test**  
1. Go to...
2. Check in...
3. Run ...

- [ ] All existing tests pass

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

**Screenshots**  
Selecting the dropdown
image
After clicking the reset button
image

**Additional context**  

### DIFF
--- a/packages/core/src/components/dropdown/dropdown-filter.scss
+++ b/packages/core/src/components/dropdown/dropdown-filter.scss
@@ -121,6 +121,11 @@
 
     tds-icon {
       margin-right: 16px;
+
+      &.hide {
+        display: none;
+        visibility: hidden;
+      }
     }
   }
 }
--- a/packages/core/src/components/dropdown/dropdown-filter.scss
+++ b/packages/core/src/components/dropdown/dropdown-filter.scss
@@ -123,10 +123,6 @@
       cursor: pointer;
     }
 
-    .menu-icon {
-      margin-right: 16px;
-    }
-
     .clear-icon {
       margin: 0 8px;
       color: var(--tds-dropdown-clear-icon-color);
--- a/packages/core/src/components/dropdown/dropdown-filter.scss
+++ b/packages/core/src/components/dropdown/dropdown-filter.scss
@@ -123,6 +123,10 @@
       cursor: pointer;
     }
 
+    .menu-icon {
+      margin-right: 16px;
+    }
+
     .clear-icon {
       margin: 0 8px;
       color: var(--tds-dropdown-clear-icon-color);
--- a/packages/core/src/components/dropdown/dropdown-filter.scss
+++ b/packages/core/src/components/dropdown/dropdown-filter.scss
@@ -121,6 +121,7 @@
 
     tds-icon {
       margin-right: 16px;
+      cursor: pointer;
 
       &.hide {
         display: none;
--- a/packages/core/src/components/dropdown/dropdown-filter.scss
+++ b/packages/core/src/components/dropdown/dropdown-filter.scss
@@ -120,8 +120,22 @@
     }
 
     tds-icon {
-      margin-right: 16px;
       cursor: pointer;
+    }
+
+    .menu-icon {
+      margin-right: 16px;
+    }
+
+    .clear-icon {
+      margin: 0 8px;
+      color: var(--tds-dropdown-clear-icon-color);
+      padding-right: 8px;
+      border-right: 1px solid var(--tds-dropdown-clear-icon-color);
+
+      &:hover {
+        color: var(--tds-dropdown-clear-icon-hover-color);
+      }
 
       &.hide {
         display: none;
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -530,13 +530,27 @@ export class TdsDropdown {
                 />
               </div>
               <tds-icon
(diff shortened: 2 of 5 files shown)


timrombergjakobsson:
@finnlake thanks a lot for your contribution! However there are some things to first consider. We don't have a design on this functionality in Figma, so it needs to be looked at by a designer, ping @charlottelorensson. Personally, I don't see why this could not be a nice addition of functionality to the component. However I think we should have some regression tests added, as it is functionality visible only with filter.

packages/core/src/components/dropdown/dropdown-filter.scss
```diff
@@ -121,6 +121,11 @@
 
     tds-icon {
       margin-right: 16px;
+
```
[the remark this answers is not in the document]
finnlake: I was thinking of that, but that was not set on the chevron.
But maybe it should be then?

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -595,4 +601,10 @@ export class TdsDropdown {
       </Host>
     );
   }
+
+  private handleFilterReset = () => {
```
mJarsater, line 605: Could we move this to before the `render` func? Just for consistency.

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -522,6 +528,12 @@ export class TdsDropdown {
                   }}
                 />
               </div>
+              <tds-icon
+                onClick={this.handleFilterReset}
+                class={`${this.open && this.inputElement.value !== '' ? '' : 'hide'}`}
+                name="cross"
+                size="16px"
+              ></tds-icon>
```
mJarsater, line 536: I'd consider making this reachable by keyboard, wrapping it in a button would create semantically incorrect HTML I believe. But should be able to move it outside of the input element in the DOM and then position it correctly using CSS.

finnlake: Ok, will have a look at it when I get the time.

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -530,13 +530,27 @@ export class TdsDropdown {
                 />
               </div>
               <tds-icon
+                tabIndex={0}
+                role="button"
```
timrombergjakobsson, line 534: 👍

theJohnnyMe:
@finnlake thanks a lot for this nice contribution!

I did take a look on Figma and comments so this is what I changed:

- Rebased the branch with latest changes from develop branch
- Color updates for icon and border according to Figma file
- Added reset method to handleFilterReset so it clears dropdown menu selection
- Improved a11y for using keyboard to navigate clear and menu button 

@finnlake, @timrombergjakobsson, @mJarsater please let me know what do you think of it :)

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -523,8 +530,28 @@ export class TdsDropdown {
...
+                onKeyDown={(event) => {
+                  if (event.key === 'Enter') {
+                    this.handleFilterReset();
+                  }
+                }}
+                class={`clear-icon ${this.open && this.inputElement.value !== '' ? '' : 'hide'}`}
+                name="cross"
+                size="16px"
+              ></tds-icon>
+              <tds-icon
```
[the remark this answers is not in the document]
theJohnnyMe: Great idea, added! ;)

timrombergjakobsson:
Looks great! Just one thought, now when we hit `esc`, we close the dropdown, but it also breaks the filter, as we can't resume to type and filter. Should that be intended behavior? If not, maybe we should reset the filter when hitting escape, and not merely set the open state to false?

theJohnnyMe:
> Looks great! Just one thought, now when we hit `esc`, we close the dropdown, but it also breaks the filter, as we can't resume to type and filter. Should that be intended behavior? If not, maybe we should reset the filter when hitting escape, and not merely set the open state to false?

Good spotting, the issue exists in the current production code. I would say that we make a ticket out of it and sort it in another PR and leave this one within area of contribution.

timrombergjakobsson:
> > Looks great! Just one thought, now when we hit `esc`, we close the dropdown, but it also breaks the filter, as we can't resume to type and filter. Should that be intended behavior? If not, maybe we should reset the filter when hitting escape, and not merely set the open state to false?
> 
> Good spotting, the issue exists in the current production code. I would say that we make a ticket out of it and sort it in another PR and leave this one within area of contribution.

Yes it does! Let's create a ticket for it.

timrombergjakobsson (approved):

finnlake:
Thank you for accepting the pr!
Nice work @theJohnnyMe , there was a considerable amount of work needed that I didnt think of... :)

timrombergjakobsson:
> Thank you for accepting the pr! Nice work @theJohnnyMe , there was a considerable amount of work needed that I didnt think of... :)

Great work @finnlake!

PR #564, opened 2024-03-22
fix(table-body-cell): map text-align prop in lifecycle

theJohnnyMe:
**Describe pull-request**  
tds-body-cell, when generated via for/while/map loop, missed internal event for text-align property from parent. That said, no matter body cell has prop for text-align itself, it does not get mapped to textAlignState as mapping was happening only in internal event. Now mapping is added to the lifecycle too. 

**How to test**  
1. Check if prop will work in cases where cells are generated from JSON file, local textAlign prop on tds-body-cell has to have effect on text alignment.

React demo example with this code used via npm pack: https://github.com/scania-digital-design-system/tegel-react-demo/pull/176

PS: I increased tolerance on table test to 5% and made new screenshots too for two tests.


nathalielindqvist (approved):
Nicely done! Works well in the React demo 👌 Great that you increased the tolerance for the tests too, they all pass on local 😄

timrombergjakobsson (approved):
👍

PR #572, opened 2024-03-27
test: optimisation of tests execution 

theJohnnyMe:
**Describe pull-request**  
1. Fixing failing tests
2. Splitting large tests in a couple of smaller
3. Increasing number of workers from 4 to 8
4. Adding "parallel" setting to all current test cases

**How to test**  
1. Run the regular develop branch and run `npm run test`
2. Check the time it took and outcome (fails sometimes) 
3. Run this branch and run `npm run test`
4. Check the time it took and the outcome (should pass all the times)
5. Time-saving should be around 25%

**Additional context**  
If having time, try steps 1 and 3 ten times each. I hope that this branch has improved test stability.


packages/core/playwright.config.ts
```diff
@@ -7,13 +7,14 @@ const config: PlaywrightTestConfig = {
   testMatch: '*.e2e.ts',
   use: {
     baseURL: 'http://localhost:3333',
+    browserName: 'chromium',
```
mJarsater, line 10: Would it not make sense to include other rendering engines? Like Webkit for Safari and Gecko for Firefox? I think these automated tests is a great place for catching inconsistencies between browsers, rather than you guys having to check all browsers.

timrombergjakobsson: Think that makes sense, I think you can set it up something like the suggested change below.

packages/core/playwright.config.ts
```diff
@@ -7,13 +7,14 @@ const config: PlaywrightTestConfig = {
...
   },
   webServer: {
     command: 'serve -p 3333',
     port: 3333,
     reuseExistingServer: !process.env.CI,
   },
-  workers: 4,
+  workers: 8,
```
timrombergjakobsson, line 17:
```suggestion
  workers: 8,
  reporter: 'line',
  projects: [
    {
      name: 'Chromium',
      use: { browserName: 'chromium' },
    },
    {
      name: 'Firefox',
      use: { browserName: 'firefox' },
    },
    {
      name: 'WebKit',
      use: { browserName: 'webkit' },
    },
  ],
```

timrombergjakobsson: In Playwright you just specify 'Firefox' I believe.

packages/core/playwright.config.ts
mJarsater: Don't know if this will impact the amount of snapshots though, but would be good to try out

theJohnnyMe: I agree, I think we can add more of those like Tim suggested in the comment below. My concern is how long it should take to run the test when committing too. Right now it is round 20sec on my Mac, with changes in this branch it goes down to 14sec. 

I want to optimize it so it does not take that much time. Maybe we need to consider running only test of components you work on currently and not entire system? Maybe entire system check should happen on opened PR instead and block merging if it is failing?

packages/core/playwright.config.ts
theJohnnyMe: Lets give it a shot in afternoon, I want to see how much time it will take to run all the tests if we have all the browsers :)

packages/core/playwright.config.ts
mJarsater: I think you could move the running of tests to the github action only, and remove it from the pre-commit hook?

packages/core/playwright.config.ts
theJohnnyMe: @mJarsater & @timrombergjakobsson here is the outcome:

Screenshot 2024-03-28 at 15 20 51

Running test on local took over 2min. Every browser generated its own screenshot. I think we need to change approach to testing only component changed on commit and leave entire system check once PR is submitted.

packages/core/playwright.config.ts
theJohnnyMe: we are not alone: https://github.com/microsoft/playwright/issues/15075

timrombergjakobsson:
> For now, I would like to propose merging this one and exploring making extensive tests for all browsers that would run on PR opening as GitHub action. Proposal is also to make it mandatory for that test to pass before merging.

Yes! Let's do that @theJohnnyMe!

nathalielindqvist (approved):

PR #577, opened 2024-04-02
Feat/cdep 3033 datetime unittest

timrombergjakobsson:
**Describe pull-request**  
Adding [unit tests](https://scaniaazureservices-my.sharepoint.com/:x:/r/personal/tatyana_bumazhkina_scania_com/_layouts/15/doc2.aspx?sourcedoc=%7BCA48BF6D-92CF-4E8C-8F45-6337E2002EC4%7D&file=Unit%20test%20scenarios.xlsx&nav=MTVfezZFMkQzNzcyLUREQjAtNENFOS1BMDk3LTgwNzg2NDgzNUQyQ30&action=default&mobileredirect=true) for datetime component. 

**Solving issue**  
Fixes: [CDEP-3033](https://tegel.atlassian.net/browse/CDEP-3033)


packages/core/src/components/datetime/test/error/datetime.e2e.ts
```diff
@@ -0,0 +1,71 @@
...
+    );
+
+    const errorIcon = page.locator('tds-icon[name="error"]');
+    await expect(errorIcon).toBeVisible();
+  });
+
+  test('Clock icon focuses the time input', async ({ page }) => {
+    await page.goto(componentTestPath);
+
+    await page.click('input[type="time"]');
```
nathalielindqvist, line 49: Clicking on the icon itself opens the dropdown to choose a time in the component, should the test be written to incorporate that as well or is this enough? 🙂

packages/core/src/components/datetime/test/default/datetime.e2e.ts
```diff
@@ -0,0 +1,63 @@
...
+    await expect(helperText).toHaveText('Helper text');
+
+    await expect(calendarIcon).toBeVisible();
+  });
+
+  test('icon click triggers native datetime picker', async ({ page }) => {
+    await page.goto(componentTestPath);
+
+    // Assuming the calendar icon can be clicked to open the datetime picker
+    await page.click('input[type="datetime-local"]');
```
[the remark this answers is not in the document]
timrombergjakobsson: Same as below :)

packages/core/src/components/datetime/test/error/datetime.e2e.ts
nathalielindqvist: I'm wondering if it would be fruitful to generate a new screenshot once the icon has been clicked to verify that the picker actually opens, but maybe it will create more problems than it will solve? Not sure if the native part of it all will create discrepancies between OS, or if it would resolve itself since we're using Docker?

timrombergjakobsson: On second thought, we can't capture the opening of the native calendar. So best we can do is capture the focus of the input and also the values is the correct format etc.

packages/core/src/components/datetime/test/default/datetime.e2e.ts
```diff
@@ -0,0 +1,63 @@
...
+  }) => {
+    await page.goto(componentTestPath);
+
+    // Check for the label text
+    const label = page.locator('.tds-datetime-label');
+    const dateTime = page.locator('tds-datetime');
+    const helperText = page.locator('.tds-datetime-helper .tds-helper');
+    const calendarIcon = page.locator('tds-icon[name="calendar"]');
+
+    await expect(dateTime).toHaveAttribute('size', 'lg');
```
theJohnnyMe, line 25: Checking attributes is good but it can be false positive, for example, attribute can be there but if CSS is wrong, like instead of 56px there is a value of 48px. Please do check height of element too to confirm proper height. I think "toHaveCSS" can be used there :)

packages/core/src/components/datetime/test/default/datetime.e2e.ts
```diff
@@ -0,0 +1,63 @@
...
+  test('verifies label, helper text, size, and calendar icon for datetime-local component', async ({
+    page,
+  }) => {
+    await page.goto(componentTestPath);
+
+    // Check for the label text
+    const label = page.locator('.tds-datetime-label');
+    const dateTime = page.locator('tds-datetime');
+    const helperText = page.locator('.tds-datetime-helper .tds-helper');
+    const calendarIcon = page.locator('tds-icon[name="calendar"]');
```
theJohnnyMe, line 23: The spreadsheet is asking for a check of these props/attributes too:

- min
- max
- disabled 
- no-min-width

timrombergjakobsson: Yes, I saw that too, but only empty values, and when empty it is not rendered right?

packages/core/src/components/datetime/test/default/datetime.e2e.ts
timrombergjakobsson: True, although we do that quite a lot in our tests :). Checking for `toHaveAttribute()`

packages/core/src/components/datetime/test/default/datetime.e2e.ts
```diff
@@ -0,0 +1,63 @@
...
+    // Assuming the calendar icon can be clicked to open the datetime picker
+    await page.click('input[type="datetime-local"]');
+
+    await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });
+
+    // Verify the input is indeed focused. This is an indirect test to ensure
+    // that actions leading up to the display of the native picker are correctly initiated.
+    const inputIsFocused = await page.evaluate(
+      () => document.activeElement.type === 'datetime-local',
+    );
```
theJohnnyMe, line 46:
```suggestion
    const inputIsFocused = await page.evaluate(
  () => (document.activeElement as HTMLInputElement).type === 'datetime-local',
);
```

This line trows an error in my IDE, this seems to fix it.

packages/core/src/components/datetime/test/default/datetime.e2e.ts
theJohnnyMe: I know and we would have to improve on that. For example, I think if size is component changes the screenshot will fail first, however, it won't tell us more than that screenshots do not match. If we do check the height, it would be an additional test that failed that can help us to isolate the issue faster.

I think it improves test precision, makes it easier for us and contributors to isolate issues much faster. 
What do you think?

timrombergjakobsson: Yes, I counted at least 33 instances of the usage toHaveAttribute that we need to have a look at then, some very recently merged. Maybe it is better to add toHaveCSS as a compliment instead? An additional check, maybe that is what you meant?

theJohnnyMe: Yeah, exactly, to compliment and really confirm that attribute made an effect on component.

packages/core/src/components/datetime/test/success/datetime.e2e.ts
```diff
@@ -0,0 +1,31 @@
...
+    await expect(label).toHaveText('Label text');
+
+    // Check for the helper text
+    const helperText = page.locator('.tds-datetime-helper .tds-helper');
+    await expect(helperText).toBeVisible();
+    await expect(helperText).toHaveText('Helper text');
+
+    // For example, checking for the "state" attribute to be "success"
+    await expect(dateTime).toHaveAttribute('state', 'success');
+    await expect(dateTime).toHaveAttribute('size', 'md');
```
theJohnnyMe, line 29: Here maybe it would be good to check some CSS to confirm the success state and correct height for medium.

theJohnnyMe: Spreasheet specifies checking more attributes and doing selections and checking today's date that are not part of unit test here. Was leaving them out intentional?

packages/core/src/components/datetime/test/error/datetime.e2e.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    // Check if the time input is focused after clicking the icon
+    // This assumes the time input has a specific ID or class you can target
+    const isTimeInputFocused = await page.evaluate(() => document.activeElement.type === 'time');
```
theJohnnyMe, line 53:
```suggestion
     const isTimeInputFocused = await page.evaluate(
      () => (document.activeElement as HTMLInputElement).type === 'time',
    );
```
Improvment for IDE and type issue.

packages/core/src/components/datetime/test/success/datetime.e2e.ts
timrombergjakobsson: There are some cases that are related to interactions with the datepicker we open, but as it is a native one there are some limitations to what we can do in unit tests. If that is what you mean?

theJohnnyMe (approved):
Nice work Tim! LGTM!

nathalielindqvist (approved):
LGTM, great work! 🚀

PR #593, opened 2024-04-09
fix(header-dropdown): close on slotted items click

theJohnnyMe:
**Describe pull-request**  
Close the header dropdown after clicking on one of its children

**Solving issue**  
Fixes: [CDEP-2998](https://tegel.atlassian.net/browse/CDEP-2998)

**How to test**  
1. Go to the [this PR](https://github.com/scania-digital-design-system/tegel-react-demo/pull/178). It is built using develop. Observe what happens on clicking on one of the items in Header Dropdown, Dropdown menu stays opened. 
2. Go to [PR with this fix](https://github.com/scania-digital-design-system/tegel-react-demo/pull/179). Click the Header Dropdown and observe what is happening. Navigation should work and menu should close after clicking on one of its items.
3. Check the code in this PR if it makes sense.

- [ ] All existing tests pass

**Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

**Additional context**  
Hard to test in Storybook, therefor I made packed version of code and made PRs for comparison in React Demo page. 
If this works, we can add the same behaviour to the Popover Canvas and Popover Menu, maybe even Header Launcher.

[CDEP-2998]: https://tegel.atlassian.net/browse/CDEP-2998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


timrombergjakobsson:
Tested it out, seems to work nicely @theJohnnyMe, but what @adarean5 mentioned might be worth trying out!

nathalielindqvist (dismissed):
Nicely done! Works well 👏

packages/core/src/components/header/header-dropdown/header-dropdown.tsx
```diff
@@ -43,6 +43,33 @@ export class TdsHeaderDropdown {
     this.open = !this.open;
   }
 
+  connectedCallback() {
+    // Listen for click events on the default slot's assigned nodes
+    this.host.shadowRoot.addEventListener('slotchange', (event) => {
```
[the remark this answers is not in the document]
theJohnnyMe: Code updated! ;)

theJohnnyMe:
> Tested it out, seems to work nicely @theJohnnyMe, but what @adarean5 mentioned might be worth trying out!

Code updated to target only dropdown menu that holds all items ;)

adarean5:
> > Tested it out, seems to work nicely @theJohnnyMe, but what @adarean5 mentioned might be worth trying out!
> 
> Code updated to target only dropdown menu that holds all items ;)

Would it still work if you just added a default click binding to the slot? Like so (this is on line 118)
```
{this.open ? <slot onClick={() => {this.handleSlottedItemClick()}}></slot> : null}
```
My thought process is that it might be good enough to just have a click handler on for closing the wrapper on the wrapper itself, rather than add additional complexity by trying to add handlers to elements that might appear on the handler.
The downside of this approach is that the dialog would close no matter if the user clicks on an item within the wrapper or some empty space between the items.

theJohnnyMe:
> > > Tested it out, seems to work nicely @theJohnnyMe, but what @adarean5 mentioned might be worth trying out!
> > 
> > 
> > Code updated to target only dropdown menu that holds all items ;)
> 
> Would it still work if you just added a default click binding to the slot? Like so (this is on line 118)
> 
> ```
> {this.open ? <slot onClick={() => {this.handleSlottedItemClick()}}></slot> : null}
> ```
> 
> It might be good enough to have a click handler for closing the wrapper on the wrapper itself, rather than trying to add handlers to elements that might appear in the wrapper. The downside of this approach is that the dialog would close no matter if the user clicks on an item within the wrapper or some empty space between the items.

This is a good idea, but the more I think about it is that I would do some hybrid solution. So using what you say here, but understand if click is coming from tds-header-dropdown-list-item or even its children like anchor tag or button. I feel people might slot some section titles, dividers or other non interactive elements in it which would create confusion on what is happening and why it is closing.

timrombergjakobsson (dismissed):
Nice one @theJohnnyMe, next up maybe should be the same for header-launcher as well?

theJohnnyMe:
> Nice one @theJohnnyMe, next up maybe should be the same for header-launcher as well?

Yes, we would need to the same for that sub-component too but in another PR.

nathalielindqvist (approved):
Nice update! I think this might be more sustainable than using the slots, great job! 😄

nathalielindqvist (commented):
Might not be part of this PR, but the implementation in the React demo page seems to expect a <Link> element, so when feeding in just a title the styling is a bit off?

theJohnnyMe:
> Might not be part of this PR, but the implementation in the React demo page seems to expect a element, so when feeding in just a title the styling is a bit off?

There I have injected just a text to see how element will behave and will it close on it or not. It is just for test purposes, I do not think we have any subcomponent for titles in dropdown.

packages/core/src/components/header/header-dropdown/header-dropdown.tsx
```diff
@@ -43,6 +43,13 @@ export class TdsHeaderDropdown {
     this.open = !this.open;
   }
 
+  handleSlottedItemClick = (event: MouseEvent) => {
+    const eventSource = (event.target as HTMLElement).tagName.toLowerCase();
+    if (eventSource === 'a' || eventSource === 'button') {
```
timrombergjakobsson, line 48: We could rewrite it like this also:

```
if (['a', 'button'].includes(eventSource)) {
    this.open = false;
  }
```

Potentially more readable if we add more tag names going forward?

theJohnnyMe: Great idea @timrombergjakobsson ! Thanks for suggestion ;)

timrombergjakobsson (approved):
:shipit:

Issue #557, opened 2024-03-20
[Bug report]: missing padding in Table header in compact mode


jolalab:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

@scania/tegel: ^1.6.0

### Browser

Chrome

### Framework

Angular

### Version

Angular 16

### Reproduction steps

1. Create a Tegel Table with a header 
2. Set it to use compact mode
3. Write a long text in one of the columns
4. Shrink the window until the line breaks into multiple rows
5. Observe how there is no padding over or under the text

### Code example

```html
   <tds-table
        [responsive]="true"
        [compactDesign]="true"
        [noMinWidth]="true"
    >
                <tds-table-header>
                        <tds-header-cell
                            [cellValue]="A very very long text that keeps going and going and never ends and just keeeps going until the bitter end and never stops because it just keeps going and going and just continues through time and space into eternity and just continues after that too..."
                            sortable="true"
                       >
                </tds-table-header>

                <tds-table-body>
                     <tds-table-body-row></tds-table-body-row>
                 </tds-table-body>                 
</tds-table>
```

### Screenshots

![image](https://github.com/scania-digital-design-system/tegel/assets/164156972/0cd8b82c-4285-468e-8dcc-884ab915933c)

### Expected behaviour

There should be padding at the top and bottom even when the text breaks into multiple rows and when it is in compact mode

### Console errors

Add warning/error message

### Contact information

jonatan.lampa@scania.com

nathalielindqvist:
Hi @jolalab! Thank you for reporting this, a fix for the issue has now been merged and will be available in the next release! :)

Issue #469, opened 2023-12-13
[Bug report]: Enabling the compactDesign prop on tds-table overrides the padding from the disablePadding prop on the tds-body-cell.


adarean5:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

@scania/tegel: 1.3.3

### Browser

Chrome

### Framework

Angular

### Version

Angular 16

### Reproduction steps

1. Create a tegel table containing tds-body-cell elements.
2. Set the `compactDesign` prop on the `tds-table` component to `true`.
3. Set the `disablePadding` prop on the `tds-body-cell` to `true`.

Result: The padding is still applied to the cell.

### Code example

```html
<tds-table [compactDesign]="true">
    <tds-table-header>
        <tds-header-cell
            cellKey="description"
            cellValue="Description"
        ></tds-header-cell>
    </tds-table-header>
    <tds-table-body>
        <tds-table-body-row>
            <tds-body-cell
                cellKey="description"
                cellValue="Description"
                [disablePadding]="true"
            ></tds-body-cell>
        </tds-table-body-row>
    </tds-table-body>
</tds-table>
```

### Screenshots

In the dev console, it can be observed that the compact styling overrides the no-padding styling.
![MicrosoftTeams-image](https://github.com/scania-digital-design-system/tegel/assets/15961301/fad5a43b-8b35-4359-a734-9d20e3581e95)

### Expected behaviour

After applying the `disablePadding` prop to the cell I would expect the cell not to have padding.

### Console errors

None

### Contact information

jernej.strazisar@scania.com

nathalielindqvist:
Hi @adarean5! Thank you for reporting this, a fix for the issue has now been merged and will be available in the next release! :)

adarean5:
Awesome, thank you!

PR #594, opened 2024-04-09
fix(table): disable-padding with compact design

nathalielindqvist:
**Describe pull-request**  
This PR enables disable-padding to work with compact design instead of the latter overwriting the former. It also sets a top and bottom padding on the Table header text to ensure padding even if the text is very long and breaks into multiple rows.

**Solving issue** 
Fixes: [CDEP-3194](https://tegel.atlassian.net/browse/CDEP-3194) and [CDEP-3215](https://tegel.atlassian.net/browse/CDEP-3215)
https://github.com/scania-digital-design-system/tegel/issues/469
https://github.com/scania-digital-design-system/tegel/issues/557

**How to test**  
1. Go to Storybook link below.
2. Check in Table -> Basic.
3. Enable "Compact design" and "Disable cell padding".
4. Ensure the Table cell padding is disabled _and_ the Table header padding is reduced at the same time.
5. Manipulate the Table header text to be long enough to break into multiple rows.
6. Ensure there is still 8px top and bottom padding on the text.

- [x] All existing tests pass

**Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

[CDEP-3194]: https://tegel.atlassian.net/browse/CDEP-3194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CDEP-3215]: https://tegel.atlassian.net/browse/CDEP-3215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/table/table-body-cell/table-body-cell.scss
+++ b/packages/core/src/components/table/table-body-cell/table-body-cell.scss
@@ -16,7 +16,7 @@
 }
 
 :host(.tds-table__body-cell--no-padding) {
-  padding: 0;
+  padding: 0 !important;
 }
 
 :host(.tds-table__body-cell--hover) {
--- a/packages/core/src/components/table/table-body-cell/table-body-cell.scss
+++ b/packages/core/src/components/table/table-body-cell/table-body-cell.scss
@@ -15,10 +15,6 @@
   transition: background-color 200ms ease;
 }
 
-:host(.tds-table__body-cell--no-padding) {
-  padding: 0 !important;
-}
-
 :host(.tds-table__body-cell--hover) {
   background-color: var(--tds-table-body-cell-background-hover);
 }
--- a/packages/core/src/components/table/table-body-cell/table-body-cell.tsx
+++ b/packages/core/src/components/table/table-body-cell/table-body-cell.tsx
@@ -104,6 +104,19 @@ export class TdsTableBodyCell {
   }
 
   render() {
+    let paddingStyle = 'var(--tds-spacing-element-16)'; // Default padding
+
+    if (this.disablePadding) {
+      paddingStyle = '0';
+    } else if (this.compactDesign) {
+      paddingStyle = 'var(--tds-spacing-element-8) var(--tds-spacing-element-16)';
+    }
+
+    const dynamicStyles = {
+      textAlign: this.textAlignState,
+      // Conditionally set padding style
+      padding: paddingStyle,
+    };
     return (
       <Host
         class={{
@@ -112,9 +125,8 @@ export class TdsTableBodyCell {
           'tds-table__compact': this.compactDesign,
           'tds-table--divider': this.verticalDividers,
           'tds-table--no-min-width': this.noMinWidth,
-          'tds-table__body-cell--no-padding': this.disablePadding,
         }}
-        style={{ textAlign: this.textAlignState }}
+        style={dynamicStyles}
       >
         {this.cellValue}
         <slot />
--- a/packages/core/src/components/table/table-header-cell/table-header-cell.scss
+++ b/packages/core/src/components/table/table-header-cell/table-header-cell.scss
@@ -19,7 +19,7 @@
   transition: background-color 200ms ease;
 
   .tds-table__header-text {
-    padding: 0 16px;
+    padding: 8px 16px;
     margin: 0;
   }
 }


packages/core/src/components/table/table-body-cell/table-body-cell.scss
```diff
@@ -16,7 +16,7 @@
 }
 
 :host(.tds-table__body-cell--no-padding) {
-  padding: 0;
+  padding: 0 !important;
```
timrombergjakobsson, line 19: I'm not a super fan of using !important. Without checking a substitute here, is there another way?

nathalielindqvist: Me neither. The alternative could be to use if-statements of some kind to enable conditional rendering of the classes, but since the classes are established in different files I'm not sure how straight forward that would be. 'compact-design' is set in the header and then propagates down to the body cells while 'disable-padding' is set straight in the body cells. Maybe there's a way to connect the two files? In my mind Table header would have to check the Table body cell to see if 'disable-padding' is present and then conditionally decide if the 'compact-design' class should be added or not, but maybe I'm making it more complicated than it needs to be?

timrombergjakobsson: @nathalielindqvist maybe you can try and do this instead:

```
render() {
    const dynamicStyles = {
      textAlign: this.textAlignState,
      // Conditionally set padding style
      padding: this.disablePadding ? '0' : 'var(--tds-spacing-element-16)',
    };

    
    return (
      <Host
        class={{
          'tds-table__body-cell': true,
          'tds-table__body-cell--hover': this.activeSorting,
          'tds-table__compact': this.compactDesign,
          'tds-table--no-min-width': this.noMinWidth,
        }}
        style={dynamicStyles}
      >
        {this.cellValue}
        <slot />
      </Host>
    );
  }
```
I believe you can then remove the css for padding.

timrombergjakobsson:
"Manipulate the Table header text to be long enough to break into multiple rows.", maybe doing somethig wrong, but cant make it break into 2 lines?

packages/core/src/components/table/table-body-cell/table-body-cell.scss
nathalielindqvist: It absolutely works for just toggling the correct padding for disablePadding, but if compactDesign and disablePadding are enabled at the same time the inline style overwrites the rules for the compactDesign and sets 16px when top and bottom should be 8px. It gets a little tricky when the props need to play ball with each other.

nathalielindqvist:
> "Manipulate the Table header text to be long enough to break into multiple rows.", maybe doing somethig wrong, but cant make it break into 2 lines?

The text doesn't break into multiple lines unless there is a width maximum on the column, I added the custom-width attribute to one of the table-header-cells and set it to 150px, but I'm sure there are other ways to do it as well :)

packages/core/src/components/table/table-body-cell/table-body-cell.scss
theJohnnyMe: I agree with Tim, lets try to avoid !important and find another way to resolve this.

timrombergjakobsson: @nathalielindqvist ok, but if we do something like this:

```
let paddingStyle = 'var(--tds-spacing-element-16)'; // Default padding

    if (this.disablePadding) {
        paddingStyle = '0';
    } else if (this.compactDesign) {
        paddingStyle = 'var(--tds-spacing-element-8) var(--tds-spacing-element-16)';
    }
    
    const dynamicStyles = {
      textAlign: this.textAlignState,
      // Conditionally set padding style
      padding: paddingStyle
    };
```

nathalielindqvist: Thank you for the discussion! I removed the !important and implemented the conditional solution @timrombergjakobsson provided, thank you for that! Changes pushed :)

timrombergjakobsson: Glad to help :)!

timrombergjakobsson (approved):

theJohnnyMe (approved):

nathalielindqvist:
> Result I get and I think that is what fix was aiming at: Screenshot 2024-04-15 at 16 19 30
> 
> There is another Issue, that maybe your fix is fixing too, please check: #557 and if yes, do inform user about fix rolling out soon ;)
> 
> Nice work overall. One thing I think we should see if we can improve it rather using CSS classes to fix issue compared to using inline styling. I challenge you, in your learning time, to try to resolve this issue but using CSS classes instead.

Yes exactly! I'll make sure to ping the users from both issues and let them know the fixes are going out in the next release :)

I'll take a look at using CSS classes as well!

Issue #524, opened 2024-02-22
[Bug report]: Cannot set properties of undefined (setting 'value')


ZakirQamar:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

"@scania/tegel-angular": "^1.6.0"

### Browser

Chrome

### Framework

Angular

### Reproduction steps

1. Add tds dropdown to your page
2. use angular reactive form 
3. bind dropdown options dynamically using a list created/initialized in typescript
4: assign default value

### Code example

<tds-dropdown
            ngDefaultControl
            name="market"
            default-value="FR"
            placeholder=""            
            label="Market"
            label-position="outside">  
              <ng-container *ngFor="let market of configuratorService.markets; let idx = index">
                <tds-dropdown-option [value]="market.code">
                  {{ market.name }}
                </tds-dropdown-option>
              </ng-container>
          </tds-dropdown>

### Screenshots

1: set default value is not working..
![image](https://github.com/scania-digital-design-system/tegel/assets/16921625/de419fc8-4de5-4e8f-bfb0-c407341c79f6)

2: drop down select event is throwing error
![image](https://github.com/scania-digital-design-system/tegel/assets/16921625/81b83a9d-ff67-4de7-beff-b8b15e6a9b0d)

### Expected behaviour

1: The tds dropdown should work with the default value.
2: The tds dropdown shouldn't throw any errors on dropdown select

### Console errors

ERROR Error: Uncaught (in promise): TypeError: Cannot set properties of undefined (setting 'value')

mJarsater:
I think you would want to user `default-value` and not `value` for setting a default value. See: https://tds-storybook.tegel.scania.com/?path=/story/components-dropdown--default&args=defaultOption:Option+1&globals=backgrounds.value:!hex(F9FAFB)

ZakirQamar:
Sorry for the incomplete information. Yes, I am using the 'default-value' property, and it works when I use a static value(predefined options) in the dropdown options. However, it doesn't work when we load the options data using a loop.

e.g.
<tds-dropdown
            ngDefaultControl
            name="market"
            default-value="FR"
            placeholder=""            
            label="Market"
            label-position="outside">  
[...]

PR #599, opened 2024-04-12
fix(dropdown): monitor slot changes to assigning defaultValue on dropdown menu update

theJohnnyMe:
**Describe pull-request**  
Setting default value does not work properly when used in environment where slotted content is loading slow. 

**Solving issue**  
Fixes: 
- [CDEP-3192](https://tegel.atlassian.net/browse/CDEP-3192)
- https://github.com/scania-digital-design-system/tegel/issues/524

**How to test**  
1. Check if setting default value still works as intended.
2. Try to set defaultValue to the element that does not exists yet in the dropdown menu. It should return warning message.
3. Try to add option that is matching defaultValue in run time, observe if it sets it correctly. 

[CDEP-3192]: https://tegel.atlassian.net/browse/CDEP-3192?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -292,6 +292,13 @@ export class TdsDropdown {
     }
   }
 
+  /** Method to handle slot changes */
+  private handleSlotChange() {
+    if (this.defaultValue) {
+      this.setDefaultOption();
+    }
+  }
+
   /** Method to check if we should normalize text */
   private normalizeString(text: string): string {
     return this.normalizeText ? text.normalize('NFD').replace(/\p{Diacritic}/gu, '') : text;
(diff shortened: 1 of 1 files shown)


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -308,27 +311,46 @@ export class TdsDropdown {
...
             element.setSelected(false);
           }
         }
         this.setValueAttribute();
         return element;
       });
+
+      if (!matched) {
+        console.warn(
+          `TDS DROPDOWN: No matching option found for defaultValue "${this.defaultValue}"`,
```
nathalielindqvist, line 350: Really nice implementation, awesome that you found a way to handle this! 👏  When setting the defaultValue to a value that is not present in the Dropdown this warning shows up twice in the console, is the setDefaultOption function running twice some how?

theJohnnyMe: Worth checking. It might be that we would need to improve it like this (this.defaultValue !== '') { } as it seems it runs the !matched line anyway. Thanks for reporting, I will check it ;)

nathalielindqvist (approved):

timrombergjakobsson (approved):

PR #602, opened 2024-04-17
fix(datetime): arrow down fix for chromium

timrombergjakobsson:
**Describe pull-request**  
In chromium based browsers when handling the input[type="date"] element when the date value exceeds or drops below the bounds defined by the min and max attributes. The year 275760 is a known maximum date value in JavaScript's Date object in many browsers, reflecting the farthest date the built-in Date can handle correctly. So when you for example have a min date set and you press key down and passes that, 275760 is shown. In Firefox and Safari it does not. 

**Solving issue**  
Fixes: Problem reported here in [support channel](https://teams.microsoft.com/l/message/19:5e33f67fe502441f914fbcdc6e2548f5@thread.skype/1712846149027?tenantId=3bc062e4-ac9d-4c17-b4dd-3aad637ff1ac&groupId=79f9bfeb-73e2-424d-9477-b236191ece5e&parentMessageId=1712846149027&teamName=Tegel%20Design%20System&channelName=Development%20support%20-%20Tegel&createdTime=1712846149027)

**How to test**  
The problem is a bit tricky to test, I recreated it by creating a small react app and npm packing and testing it there. Also in React demo pages and Angular ones.


nathalielindqvist (approved):
The fix in the demo page seems to work well, LGTM! 🦄

theJohnnyMe (changes requested):
This approach unfortunately does not work. In Angular demo page, both min and max are set, so problem does not occur. Try to remove max one for example and test it, it would be possible to recreate issue use had. Same behaviour you can get in Storybook too.

I do not know if there is proper way to fix it as this from code side. I would rather educate users to use pairs, so both min and max. Maybe in their project they might not know the max date, but they can always do something like 100 years from today's date and vice versa for min.

So issue here is using just min or just max. When none is set or both are set, component works as intended.

timrombergjakobsson:
> This approach unfortunately does not work. In Angular demo page, both min and max are set, so problem does not occur. Try to remove max one for example and test it, it would be possible to recreate issue use had. Same behaviour you can get in Storybook too.
> 
> I do not know if there is proper way to fix it as this from code side. I would rather educate users to use pairs, so both min and max. Maybe in their project they might not know the max date, but they can always do something like 100 years from today's date and vice versa for min.
> 
> So issue here is using just min or just max. When none is set or both are set, component works as intended.

I see what you mean, however I still feel that maybe something can be handled on the code side. In addition to informing them on how to use it?

theJohnnyMe:
> > This approach unfortunately does not work. In Angular demo page, both min and max are set, so problem does not occur. Try to remove max one for example and test it, it would be possible to recreate issue use had. Same behaviour you can get in Storybook too.
> > I do not know if there is proper way to fix it as this from code side. I would rather educate users to use pairs, so both min and max. Maybe in their project they might not know the max date, but they can always do something like 100 years from today's date and vice versa for min.
> > So issue here is using just min or just max. When none is set or both are set, component works as intended.
> 
> I see what you mean, however I still feel that maybe something can be handled on the code side. In addition to informing them on how to use it?

To have a response here besides Teams: If possible find another solution that would work for when the user sets only min or only max to help user from accidentally exceeding boundary set by browser.

theJohnnyMe:
Converted to draft as a fix does affect the issue. Leaving it for future investigation.

theJohnnyMe:
I am closing it due to inactivity.

PR #607, opened 2024-04-19
fix(vars): removing duplicate vars imports

timrombergjakobsson:
**Describe pull-request**  
Vars imports should only be in the global.scss file. 

**Solving issue**  
Fixes: There were some duplicate imports.


theJohnnyMe (approved):
Nice finding and clean up Tim!

PR #619, opened 2024-05-20
feat(modal): Expose a closable property to be able to hide x icon

mistermalm:
## **Describe pull-request**  
Expose a collapsible property to be able to hide x icon

## **Issue Linking:**  
- **Jira:** [CDEP-3294](https://tegel.atlassian.net/browse/CDEP-3294)    

## **How to test**  
1. Set closable property to false.
2. See that the close button is no more.

- [x] All existing tests pass
- [] I have updated the documentation (if applicable)

## **Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

## **Additional context**  
Unit tests will be added if requested.

[CDEP-3294]: https://tegel.atlassian.net/browse/CDEP-3294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


timrombergjakobsson (commented):
Nice man! Maybe we should add a test case for covering this also?

mJarsater (commented):
I think this has been requested before and if I remember correctly this is a design decision. That the modal should  **always** have a close button?

theJohnnyMe:
> I think this has been requested before and if I remember correctly this is a design decision. That the modal should **always** have a close button?

I had the same memory so I asked yesterday about it but designers approved it. I am not sure if was I confusing it with the banner component thou.

timrombergjakobsson:
> > I think this has been requested before and if I remember correctly this is a design decision. That the modal should **always** have a close button?
> 
> I had the same memory so I asked yesterday about it but designers approved it. I am not sure if was I confusing it with the banner component thou.

Yes, I think memory serves you correct. Although as I believe you can hide it in Figma, it made sense to have this flexibility in the component also. But please, @charlottelorensson or Laurens, you could also give your view on it.

nathalielindqvist (commented):
Works as expected, nicely done! Maybe introduce a test case for it to cover all bases? 😄

mistermalm:
I agree that it's not best practice UX to disable the close button, sometimes users do not want to make a decision, even if it has a cancel button, they just want to close.

Will add tests as well to ensure quality

packages/core/src/components/modal/modal.stories.tsx
```diff
@@ -72,6 +72,13 @@ export default {
         type: 'boolean',
       },
     },
+    closable: {
+      name: 'Closable',
+      description: 'Controls visibility of the close button.',
+      control: {
+        type: 'boolean',
+      },
```
theJohnnyMe, line 80: Add here that defaultValue is "true" so it reflects defaultValue from modal.tsx

packages/core/src/components/modal/modal.tsx
```diff
@@ -46,6 +46,9 @@ export class TdsModal {
    * will be decided by this prop and will need to be controlled from the outside. */
   @Prop() show: boolean;
 
+  /** Enables the close button. */
```
theJohnnyMe, line 49:
```suggestion
  /** Shows or hides the close [X] button. */
```
Just to be more specific. Enable might mean that button is there but we do something with preventDefault etc...

mistermalm:
I have now pushed tests for close button aswell

nathalielindqvist (approved):
LGTM, great work! 🚀

theJohnnyMe (approved):
LGTM

Issue #614, opened 2024-05-16
[Bug report]: TdsModal does not respect the "show" property


quaspar:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

"@scania/tegel-react": "1.8.1"

### Browser

Chrome

### Framework

React

### Version

React 18.3.1

### Reproduction steps

1. Add TdsModal with property "show" set to true.
2. Click close button or backdrop

### Code example

Demo:
https://quaspar.github.io/tegel-modal-bug-report/

Code:
https://github.com/quaspar/tegel-modal-bug-report/blob/main/src/App.js

### Screenshots

Add/Copy screenshot here

### Expected behaviour

Modal should not close as long as the show prop is true.

### Console errors

Add warning/error message

### Contact information

caspar.behrendt@scania.com

Lunkan89:
@quaspar I believe that you can reach this behavior by not using a `selector` and adding `onTdsClose={e => e.preventDefault()}`
Then showing and hiding the modal should only rely on the toggling of the show prop.
Let me know if this solution works for you.

theJohnnyMe:
Not considered an issue. Seems like lack of information. Closing it. Reopen if provided solution does not work.

quaspar:
Hello, thank you for your replies. Where am I supposed to add the onTdsClose property? I've added it to the modal component, and also added a console log, but no event seems to be caught, and modal closing is not prevented. See links to code and demo above.

PR #648, opened 2024-06-07
fix(tdsmodal): dont close when show prop is provided

Lunkan89:
## **Describe pull-request**  

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-`: [CDEP-](https://tegel.atlassian.net/browse/CDEP-)    

## **How to test**  
1. Go to...
2. Check in...
3. Run ...

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  

### DIFF
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -123,7 +123,10 @@ export class TdsModal {
   handleClose = (event) => {
     const closeEvent = this.tdsClose.emit(event);
     if (!closeEvent.defaultPrevented) {
-      this.isShown = false;
+      // respecting passed show prop
+      if (this.show === undefined) {
+        this.isShown = false;
+      }
     }
   };
 


Lunkan89:
Using this issue to learn the work flow. I dont have access to Jira yet so cant reference connected jira issue.
Assign myself to related issue: https://github.com/scania-digital-design-system/tegel/issues/614

nathalielindqvist (commented):
Nicely done! We might need to reconsider the Storybook example implementation, in this version the Modal can't be closed at all as long as the 'show' prop is true. And if the 'show' prop is false you can still show it by clicking the "Open modal" button, which adds the "show" class but not the prop. Perhaps we can find a good way to work around this?

Lunkan89:
@nathalielindqvist Maybe we should ignore the interaction with the reference element altogether if the show prop is provided, giving full control of show/hide to implementation? or am i miss understanding?

nathalielindqvist:
> @nathalielindqvist Maybe we should ignore the interaction with the reference element altogether if the show prop is provided, giving full control of show/hide to implementation? or am i miss understanding?

@Lunkan89 I think we have used the reference element to make it super clear how the component can be implemented in an actual application and not just rely on the Storybook controls. Since we have our demo pages in both Reach and Angular we could probably remove the reference element from our Storybook example. However, I still think the close button in the Modal itself should change the 'show' prop and the Modal should be hidden. Otherwise the Modals behavior might be misinterpreted, maybe?

Lunkan89:
This maybe a non-issue and the answer to https://github.com/scania-digital-design-system/tegel/issues/614 should just be to use `closeable = false` if desired behavior is to not be able to close modal on `show = true`.
This PR maybe obsolete?

timrombergjakobsson:
> @nathalielindqvist This maybe a non-issue and the answer to #614 should just be to use `closeable = false` if desired behavior is to not be able to close modal on `show = true`. This PR maybe obsolete?

Well the `closeable` prop controls whether the close button is displayed. However, it does not inherently prevent the modal from closing through other interactions, like clicking the overlay (unless the prevent prop is set to true).

packages/core/src/components/modal/modal.tsx
```diff
@@ -123,7 +123,10 @@ export class TdsModal {
   handleClose = (event) => {
     const closeEvent = this.tdsClose.emit(event);
     if (!closeEvent.defaultPrevented) {
-      this.isShown = false;
+      // respecting passed show prop
+      if (this.show === undefined) {
+        this.isShown = false;
+      }
```
theJohnnyMe, line 129: I think this makes sense as this is what we say for the show prop:
```
  /** Controls whether the Modal is shown or not. If this is set hiding and showing
   * will be decided by this prop and will need to be controlled from the outside. */
  @Prop() show: boolean;
```

My concern is should we then block the` initializeModal()` method too? In a way, both opening and closing then would be fully under control of a user by toggling state of show prop.

Issue #643, opened 2024-06-03
[Feature Request]: tds-tooltip showing on click


minghualu:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [X] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

Add functionality to show tooltip when clicking the element. Currently, the tooltip can only be shown when hovering over it.

### Background

Our application needs to be useable on tablets where hovering is not possible.

PR #653, opened 2024-06-12
feat(tds-tooltip): show on click

Lunkan89:
Added click as trigger event for showing tooltip

## **Describe pull-request**  

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-`: [CDEP-](https://tegel.atlassian.net/browse/CDEP-)    
- **GitHub:** https://github.com/scania-digital-design-system/tegel/issues/643

## **How to test**  
1. Go to...
2. Check in...
3. Run ...

- [X] I have added unit tests for my changes (if applicable)
- [X] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  


packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -82,7 +85,13 @@ export class TdsTooltip {
           }}
           selector={this.selector}
           referenceEl={this.referenceEl}
-          trigger={this.mouseOverTooltip ? 'hover-popover' : 'hover'}
+          trigger={
```
mJarsater, line 88: Let's try and avoid double turnaries

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -27,8 +27,11 @@ export class TdsTooltip {
   /** Allow mouse over Tooltip. Useful when Tooltip contains clickable elements like link or button. */
   @Prop() mouseOverTooltip: boolean = false;
 
+  /** What triggers the popover to show */
+  @Prop() trigger: 'click' | 'hover' = 'hover';
```
mJarsater, line 31: Could there not be an instance where it should be shown on both click and hover? Maybe even on focus?

timrombergjakobsson: Agree! Maybe a method for handling it instead? Something like:

```
determineTrigger() {
    if (this.trigger === 'hover') {
      return this.mouseOverTooltip ? 'hover-popover' : 'hover';
    }
    return this.trigger;
  }
```
And then use it:
`trigger={this.determineTrigger()}`

packages/core/src/components/tooltip/tooltip.tsx
timrombergjakobsson: Agree! Maybe a method for handling it instead? Something like:

```
determineTrigger() {
    if (this.trigger === 'hover') {
      return this.mouseOverTooltip ? 'hover-popover' : 'hover';
    }
    return this.trigger;
  }
```
And then use it:
`trigger={this.determineTrigger()}`

packages/core/src/components/tooltip/tooltip.tsx
timrombergjakobsson: Yes maybe, so in that case, I guess we could change the type of the trigger prop from 'click' | 'hover' to a string that allows combinations?
`  @Prop() trigger: string = 'hover'; // This can now be something like 'click,hover,focus'
`

timrombergjakobsson: Also, we might be actually mixing up the usages of a Tooltip and a Popover. So maybe a Tooltip that has a trigger for "click" is actually a Popover, it's own component.

mJarsater: I think a tooltip might still need to be shown on click, on phone for instance where a users can't hover an element. I might even suggest introducing more props over adding options to the existing `trigger`. Something like: 

```jsx
@Prop() showOnHover: boolean = true // true by default
@Prop() showOnClick: boolean = false
@Prop() showOnFocus: boolean = false
```

Or it could be a string array? But that would go against recommendations of props never being rich objects. 
```jsx
@Prop() trigger: string[] = ['hover']


<tds-tooltip trigger=['hover', 'click', 'focus']> 
```

Would really have to be validated and tested though.

mJarsater: Ah my bad, the trigger prop was introduced in this PR? Then I'd say you have every opportunity to create a more scalable and non-breaking API. My suggestion would be to just introduce two new props. 

```jsx
@Prop() shouldShowOnClick: boolean = false
@Prop() shouldShowOnFocus: boolean = false
```

And that it should **always** show on hover (if design agrees).

mJarsater: I think the `trigger` prop name is a bit misleading. To me that points to a trigger element, not a user action.

Lunkan89: super duper 👍 
I don't think it should be just a string since it correlates directly to what type of event is setup.
The naming of the prop `trigger` was because it is used in `popover-core` which tooltip uses.
Also from what i can se in `popover-core` if the trigger is `hover` or `hover-popover` a focus event is setup as well. https://github.com/scania-digital-design-system/tegel/blob/8d6cb56ac4db7102d4906a4eca3941ed5a04b5e9/packages/core/src/components/popover-core/popover-core.tsx#L195-L196

mJarsater: Ah alright, yeah then it's great that you used the same prop name! Let's try and figure out a way users could add both click and hover?

Lunkan89: Is that not redundant? since you need to hover to click?
What situation are you thinking of?

mJarsater: I'm thinking that you are building an app that is meant to be used both on desktop and mobile. For desktop you want the users to just hover, but on mobile they can't so you want it to open on click. Therefor I think a scenario where both click and hover, and like I mentioned before maybe even just focus should all show the tooltip.

@theJohnnyMe @timrombergjakobsson

Lunkan89: true, i was thinking of that the identification of specific platforms needed to be implemented outside and thus they have the ability to change the prop, to avoid unused eventlistners.
But the focus should be easy usage!

Question: Since we are using `tds-popover-core` component and that uses specific strings as type for prop `trigger` the prop i need to pass needs to be a mutual exclution of your suggestion:
```
@Prop() showOnHover: boolean = true // true by default
@Prop() showOnClick: boolean = false
@Prop() showOnFocus: boolean = false
```
If not, that would impact a larger rewrite of `tds-popover-core` would it not?
Should we leave it as is or shoot for some popover-core rewrite?
maybe separate out the addition of focus eventlistners from when 'hover' is passed as trigger to it's own prop in `tds-popover-core`?

Thanks you for the feedback!

timrombergjakobsson: Rewrite of `tds-popover-core` will most likely happen further along, would be nice to instead of relying on @popperjs/core, use the Popover attribute https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover

Also, I believe this topic might need some further refinement and discussion, I agree that we need to have some mobile adaption, as @mJarsater mentioned there is no hover on touch, so how do we adapt to that, not only code wise but also design wise. Saxed quote to the mobile discussion:

- Automatic trigger — product controls when tooltip is shown, and/or hidden
- Tap triggered — Requires action from user, but also requires affordance to know how to trigger it, such as an information icon or similar.

So then som visual feedback needs to be added on touch/mobile, possibly using media query:

```
@media (any-hover: none) and (pointer: coarse) {
  .icon-to-show {
    display: block; /* Show only on touch devices */
    position: relative; /* Position it relative to the parent element */
    z-index: 100; /* Ensure it is clickable */
    cursor: pointer; /* Indicates it's interactive */
    padding: 10px; /* Add padding for easier touch interaction */
    border-radius: 50%; /* Optional: style as a circular button */
    background-color: #eee; /* Optional: background color */
    box-shadow: 0 2px 6px rgba(0,0,0,0.15); /* Optional: drop shadow for better visibility */
  }
}
```

theJohnnyMe: Lots of good discussions guys ;)

1. On Focus: I was checking on demo pages how current Tooltip works. Once I focus button on which tooltip is attached too, it does come up. So I think we do not have to do anything about it.
2. Replacing tds-popover-core with popover attribute is a great idea @timrombergjakobsson. However, I would leave it as a refactoring ticket. The reason is that we would have to test the other 3 components (Popover Menu, Popover Canvas and Tooltip) to ensure we have the same functionality. Also, positioning of popover element about the target that triggers it we might need to develop on our own. 
3. Adding additional prop might be confusing for us after some period. I checked the code of tds-popover-core and as `trigger` prop is defined there with @watch decorator too, I would propose we stick to that prop. User can change its value once the device is with touch instead of a mouse. So I would not change @Lunkan89 idea of implementation. I do understand that user might want, in some specific cases, to ask for both click and hover functionality, if we do not refactor the component to enable it, user can always take control over the `show` prop and show it on whatever action he/she wants.

packages/core/src/components/tooltip/tooltip.tsx
```diff
@@ -27,8 +27,11 @@ export class TdsTooltip {
...
+
   /** Prop in control of showing and hiding prop */
-  @Prop({ mutable: true }) show: boolean = false;
+  @Prop({ mutable: true }) show: boolean = null;
```
theJohnnyMe, line 34:
```suggestion
  @Prop({ mutable: true }) show: boolean = false;
```
As prop is defined as a boolean, I propose we keep using false here unless there is a specific reason to keep it null.

Lunkan89: I agree, although when i change all the flags to false `popover-canvas` & `popover-menu` are effected since we rely on default value to be null when flag is not provided at all.
My suggestion is to merge this or to totally revisit testing and logic around show flags in all components using `popover-core`.

Note: I agree, but it works for now :)

mJarsater: I agree with @Lunkan89, we need to keep it undefined/null if the users hasn't used it. Need to make sure to differentiate between controlled vs uncontrolled state of components. Same should go for modal/popover/tooltip

theJohnnyMe: Then all good, I missed the part on dependency in popover-canvas and popover-menu.

theJohnnyMe (approved):
LGTM ;)

timrombergjakobsson:
I think it looks good, however, visited on mobile and regardless of setting `hover` or `click` it always shows on click. I guess as intended, but thinking that we should revisit this further down the line and provide more visual indication that we are on mobile.

PR #663, opened 2024-06-19
feat(grid): new grid

mistermalm:
## **Describe pull-request**  
Implementation of new grid

## **Issue Linking:**  
- **Jira:** [feat/CDEP-3314-new-grid](https://tegel.atlassian.net/browse/CDEP-3314)    

## **How to test**  
1. Go to Storybook
2. Check in Grid New
3. Resize window and/or toggle fluid.

## **Suggested test steps**
- [ ] Storybook controls

## **Additional context**  
No tests or component. We only ship CSS for new grid system.

[CDEP-3314]: https://tegel.atlassian.net/browse/CDEP-3314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nathalielindqvist (commented):
Nice work! Wondering if it could be beneficial to separate the fluid and fixed examples in two different sub-stories?

Also thinking about the naming. It doesn't seem like the classnames collide with our old grid so that's great! Would it be feasible to simply rename this story to "Grid" and rename the old grid story to "DEPRECATED Grid"? Another alternative could be to name this story something like "TDS Grid" to differentiate them, but then we have an issue of inconsistencies between stories on a global level since no other story have that suffix. What do you think @timrombergjakobsson @theJohnnyMe?

packages/core/src/global/grid.scss
```diff
@@ -0,0 +1,70 @@
...
+
+.tds-grid-fixed,
+.tds-grid-fluid {
+  display: grid;
+  width: calc(100% - 32px);
+  padding: 0 16px;
+  grid-template-columns: repeat(6, 1fr);
+  grid-template-rows: repeat(2, 1fr);
+  grid-gap: 16px;
+}
```
theJohnnyMe, line 14:
```suggestion
.tds-grid-fixed,
.tds-grid-fluid {
  display: grid;
  width: calc(100% - 32px);
  padding: 0 16px;
  grid-template-columns: repeat(6, 1fr);
  grid-gap: 16px;
}
```

Line related to the number of the rows should not be part of code Tegel ships. It is up to user to specify on their side. However, you will need it for demo purposes in Storybook, so please add it under story demo code:
`  grid-template-rows: repeat(2, 1fr);`

packages/core/src/global/grid.scss
```diff
@@ -0,0 +1,70 @@
...
+
+@media screen and (min-width: 416px) {
+  .tds-grid-fixed {
+    width: calc(416px - 32px);
+  }
+
+  .tds-grid-fluid {
+    grid-template-columns: repeat(12, 1fr);
+    grid-template-rows: 1fr;
+  }
```
theJohnnyMe, line 24:
```suggestion
  .tds-grid-fluid {
    grid-template-columns: repeat(12, 1fr);
  }
```

Same as upper one, use it only for demo purposes: 
` grid-template-rows: 1fr;`

packages/core/src/global/grid.scss
```diff
@@ -0,0 +1,70 @@
...
+}
+
+@media screen and (min-width: 604px) {
+  .tds-grid-fixed {
+    width: calc(604px - 32px);
+    grid-template-columns: repeat(12, 1fr);
+    grid-template-rows: 1fr;
+  }
+}
```
theJohnnyMe, line 33:
```suggestion
@media screen and (min-width: 604px) {
  .tds-grid-fixed {
    width: calc(604px - 32px);
    grid-template-columns: repeat(12, 1fr);    
  }
}
```

Same here. 
`grid-template-rows: 1fr;`

theJohnnyMe:
> Nice work! Wondering if it could be beneficial to separate the fluid and fixed examples in two different sub-stories?
> 
> Also thinking about the naming. It doesn't seem like the classnames collide with our old grid so that's great! Would it be feasible to simply rename this story to "Grid" and rename the old grid story to "DEPRECATED Grid"? Another alternative could be to name this story something like "TDS Grid" to differentiate them, but then we have an issue of inconsistencies between stories on a global level since no other story have that suffix. What do you think @timrombergjakobsson @theJohnnyMe?

I helped by pushing a new section "Deprecated", so we can push deprecated components and foundations there. 

@mistermalm  - I agree with Nathalie, maybe it is good to make separate stories for fixed and fluid grid. That also might mean removing `args `so controls panel will be gone aka making more space on screen to test grid.

mistermalm:
@theJohnnyMe  @nathalielindqvist 
Sure could separate them, will double check with Laurens since it was his to combine them.

@theJohnnyMe Thanks for the deprecation push, I really liked it how the import ended up!
Also I will double check if I have to add new or delete those css properties you suggested, based on how it's now working when I am using the grid in my test project

theJohnnyMe (approved):
Great work @mistermalm ! LGTM

nathalielindqvist (approved):
Nicely done! 🚀

packages/core/src/global/global.scss
```diff
@@ -5,8 +5,14 @@
...
-// Grid
-@import '../../../../grid/grid';
+// Grid (Deprecated - will be removed in version 2.0)
+@import '../../../../grid/grid-deprecated';
+
+// Grid new
+@import './grid';
+
+// Grid new
+@import './grid';
```
theJohnnyMe, line 16:
```suggestion

// Grid new
@import './grid';

```
Duplicate import.

packages/core/src/global/global.scss
```diff
@@ -5,8 +5,14 @@
...
```
theJohnnyMe, line 15:
```suggestion
```
Duplicate import.

PR #676, opened 2024-07-04
fix(dropdown): reset scroll

mistermalm:
## **Describe pull-request**  
Reset the scroll height of dropdown component when triggering the reset method.

## **Issue Linking:**  
- **Jira:** [CDEP-3371-reset-dropdown-scroll](https://tegel.atlassian.net/browse/CDEP-3371)    

## **How to test**  
1. Go to story file for Dropdown
2. Add a reset button  `<button onclick="dropdown.reset()">Reset</button>`
3. Scroll in the dropdown menu and select an option
4. Press reset button and see that the dropdown is scrolled back up/down.
5. Change `Open Direction` and repeat step 3 & 4.

## **Suggested test steps**
- [x] Storybook controls


nathalielindqvist (approved):
Works well, nicely done! ⭐

theJohnnyMe (approved):
I think this issue we have fixed before, as in current code change there is no changes in logic that resets dropdown menu items. It only adds the feature to scroll dropdown menu up once reset method is called. I tested reset in Storybook preview and it seems to be working. Once this goes out, @mistermalm please check with user if problem still exists. 

LGTM.

PR #678, opened 2024-07-05
ci(release-scripts): no-verify flag and release-all script update

theJohnnyMe:
## **Describe pull-request**  
 - adding --no-verify flag to all push commands in release scripts
 - remaking release all script, now with merging develop to main as part of the script

## **Issue Linking:**  
- **No issue:** These issues we encountered last couple of times during the release

## **How to test**  
1. Please check code and see if ti makes sense. If you have doubts, better flag it so we can check together.
2. I don't know if there is a proper way to test it, we may need to run it on the next release.


nathalielindqvist (approved):
I think it looks good, would be great to try this out! 👌

mistermalm (approved):

PR #689, opened 2024-07-17
feat(tegel-styles): migration in monorepo, connection to foundations

theJohnnyMe:
## **Describe pull-request**  
Migrating tegel-styles repo as part of tegel monorepo.

What was done:
 - all foundation files are (colors, typography, spacing, grid, logotype) now on top level of monorepo
 - tegel-styles are renamed to styles and moved under packages
 - styles output different CSS files depending on user needs. User can choose to import all foundation file in one file or add the one that they want to use.
 - ag-grid overwrites is also a separate file so user can import only that
 - There is a node script that converts all scss to CSS.
 - Some files are libraries are removed as they are not used or needed anymore
 
My thinking was to enable modularity and flexibility as some users will consume @scania/tegel that has all styles so they just need ag-grid. Others might use just ag-grid and foundation files in case they do not use our @scania/tegel package. Also, users that would like to use just foundations, now they can do it.

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-`: [CDEP-3427](https://tegel.atlassian.net/browse/CDEP-3427)    

## **How to test**  
1. Run branch on your local
2. Check if build command on packages/styles works for you
3. Check if main repo build as usual

[CDEP-3427]: https://tegel.atlassian.net/browse/CDEP-3427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tegel-styles/LICENSE
mJarsater, line 1: I obviously don't have the full scope here. But I'm just thinking why we would need to prefix a style folder in the root with "tegel-". Can't it be called just styles? 

Otherwise we'll end up with a repo structure like so:
tegel/tegel-styles
tegel/core
tegel/react
.... and so on.

Instead of 
tegel/styles
tegel/core
tegel/react
.... and so on.

theJohnnyMe: You are right, Martin. I have just taken the folder and migrated it here, but I will do some cleaning and connecting to the mono repo core for colors and typography.

theJohnnyMe: @mJarsater  - I made many changes now so please take a look and let me know what you think. There is a bit more description added in the PR description so please check it out. I kept "tegel-" as a prefix everywhere so it is clear to end user what is used. Possibly we could change that as folder import will be "@scania/tegel-styles/.../tegel-foundations.css" so maybe I could remove it actually, I am realizing it as I am writing these lines... 

Anyway, think about this approach with connection to foundations and offer to just import specific one or all...let me know you think ;)

tegel-styles/scss/ag-grid/tegel-ag-grid-styles.scss
```diff
@@ -0,0 +1,166 @@
+/* AG-Grid General Overrides using Tegel Design System Variables */
```
timrombergjakobsson, line 1: I'm thinking the "ag-grid"-prefix is quite redundant here? Maybe even the "ag-grid" one? Could also just be:
`/ag-grid/styles.scss`?
Or `overrides.scss`?

tegel-styles/scss/foundations/tegel-foundations.scss
```diff
@@ -0,0 +1,5 @@
+@import './tegel-colors';
```
timrombergjakobsson, line 1: Same here as comment above.

tegel-styles/scss/ag-grid/tegel-ag-grid-styles.scss
theJohnnyMe: Update done!

tegel-styles/scss/foundations/tegel-foundations.scss
theJohnnyMe: Update done!

PR #696, opened 2024-07-24
fix(dropdown): duplicate value in multiselect

timrombergjakobsson:
## **Describe pull-request**  
When multiselect is set to true, choosing default values creates duplicates in placeholder. This PR enforces value uniqueness in 2 methods. Bug is visual in latest Storybook. 

## **Issue Linking:**  
- **Jira:** : [CDEP-3396](https://tegel.atlassian.net/browse/CDEP-3396)    

## **How to test**  
1. Go to Storybook
2. Go to Dropdown component
3. Set multiselect bool to true and select multiple values from default options. Should not yield duplicates. 

## **Screenshots**  

Before:
Skärmavbild 2024-07-24 kl  09 29 48

After:
Skärmavbild 2024-07-24 kl  09 31 00

[CDEP-3396]: https://tegel.atlassian.net/browse/CDEP-3396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -614,6 +607,7 @@ export class TdsDropdown {
                   </div>
                 )}
                 <div class={`placeholder ${this.size}`}>
+                  {console.log(this.getValue())}
```
theJohnnyMe, line 610: I guess we do not need this :)

theJohnnyMe:
Nice fix @timrombergjakobsson, works as expected. Just remove console.log and I will approve it .

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -312,45 +314,36 @@ export class TdsDropdown {
...
-              this.value = this.value ? [...this.value, element.value] : [element.value];
-              matched = true;
-            }
-          });
-        } else {
-          if (this.defaultValue === element.value) {
+      const defaultValues = this.multiselect ? this.defaultValue.split(',') : [this.defaultValue];
+
+      // Ensure unique default values
+      const uniqueDefaultValues = [...new Set(defaultValues)];
```
mJarsater, line 326:
```suggestion
const defaultValues = this.multiselect ? new Set(this.defaultValue.split(',')) : new Set([this.defaultValue]);
```

Might be worth looking through this implementation and see if we are instatiating arrays that should be Sets anywhere else

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -312,45 +314,36 @@ export class TdsDropdown {
...
+
+      uniqueDefaultValues.forEach((value) => {
+        children.forEach((element: HTMLTdsDropdownOptionElement) => {
+          if (value === element.value) {
             element.setSelected(true);
-            this.value = [element.value];
+            this.value = this.value
+              ? [...new Set([...this.value, element.value])]
+              : [element.value];
```
mJarsater, line 334:
```suggestion
            this.value = this.value
              ? new Set([...this.value, element.value])
              : [element.value];
              
```

packages/core/src/components/dropdown/dropdown.tsx
timrombergjakobsson: Oops, good catch!

theJohnnyMe (approved):
LGTM

PR #705, opened 2024-08-01
fix(table): removing unused files and slots

theJohnnyMe:
## **Describe pull-request**  
There was a slot and file not used in the table so I have removed it.
The "no-result" slot was removed from the component back in May 2023, but its description remained in the file.
dummy-data.json was not used at all, and probably remained during migration from sdds.

## **Issue Linking:**  
- **No issue:** Found after support question relating to the no-result message.

## **How to test**  
1. Open preview link.
2. Check that "no-result" slot is not present under "Notes" of tds-table-body component

- [x] All existing tests pass


timrombergjakobsson:
Should we not keep it and expand on it instead?

theJohnnyMe:
> Should we not keep it and expand on it instead?

It has not been in use since May last year when the slot was deleted from the HTML piece of code. 
Even if we put an additional slot back, it would be nested just next to the default one so I do not use the use of it.

timrombergjakobsson (approved):

PR #721, opened 2024-08-15
fix(table): scroll instead of expanding parent

mistermalm:
## **Describe pull-request**  
Any elements, that are wider than the available width inside the expand-row slot, should overflow instead of expanding the entire table.

## **Issue Linking:**  
- **Jira:** [CDEP-3353: The expand-row slot expands the entire table instead of overflowing](https://tegel.atlassian.net/browse/CDEP-3353)     

## **How to test**  
1. Go to: `table-component-expandable-rows.stories.tsx`
2. Replace `<div slot="expand-row">Hello world 1</div>` with some wider content for example:  
```
<div slot="expand-row">
  <tds-body-cell cell-value="Test value 1" cell-key="truck"></tds-body-cell>
  <tds-body-cell cell-value="Test value 2" cell-key="driver"></tds-body-cell>
  <tds-body-cell cell-value="Test value 3" cell-key="country"></tds-body-cell>
  <tds-body-cell cell-value="Test value 4" cell-key="mileage"></tds-body-cell>
  <tds-body-cell cell-value="Test value 1" cell-key="truck"></tds-body-cell>
</div>
```
3. Go to the storybook Table -> Expandable Rows
4. Expand the row you edited and see that it's now scrollable instead of expanding the whole table.

## **Additional context**  
Does anyone have a suggestion for the same behaviour using other CSS properties? The max-width: 1px is working but it's not super clear what it does for anyone who just looking briefly at the styling.

### DIFF
--- a/packages/core/src/components/table/table-body-row-expandable/readme.md
+++ b/packages/core/src/components/table/table-body-row-expandable/readme.md
@@ -31,6 +31,20 @@ tds-table-body-row-expandable::part(expand-row) {
 
 This CSS will apply a background color and border to the expandable row part of the `tds-table-body-row-expandable` component.
 
+### Overflow solution for Expanded Rows (Use with Caution)
+
+If you need to control overflow in a table, you can wrap the table content using a div with the following styles:
+
+```html
+<div style="overflow:auto; width:100%; display:table-caption;">
+    // Your table content here
+</div>
+```
+
+This approach forces the content to scroll within the table. However, it leverages unconventional CSS (display: table-caption) to work across Firefox, Chrome, and Safari. This might lead to unexpected issues in complex layouts or future browser updates. While effective, use this method with caution and consider other layout strategies.
+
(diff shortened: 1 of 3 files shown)


Lunkan89:
I think we should just check in with a designer on this one as well.

theJohnnyMe (changes requested):
Fix does work in Chrome and Safari, but not in Firefox:

Screenshot 2024-08-22 at 14 08 44

I will think if I can come up with some CSS trick or we might need to use JS for it.

theJohnnyMe:
As it seems, what we propose here should not endanger any current implementation of an expanded row. It seems to work good in Firefox, Chrome and Safari, therefore I approve it.

NIce work @mistermalm ;)

theJohnnyMe (approved):

timrombergjakobsson (approved):

Lunkan89 (approved):

Issue #709, opened 2024-08-05
[Feature Request]: Close popover menu programmatically


ExeAy:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [X] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

Be able to close the pop over menu on for example a button click.

### Background

Not all actions on a menu are links. We have for example a modal that is triggered or a string that gets copied. In those cases the menu remains open.

### Idea

Expose a close method that can be called. This will close the dialog.

Another solution is to make it automatically close on any click but that will be a breaking change and I'm not sure if it's something we want. Although that solution is preferred for me.

### Goals and objectives

Possible to close the menu programmatically.

### Contact information

hannes.falk@scania.com

PR #726, opened 2024-08-23
feat(popover): close menu programmatically

ExeAy:
## **Describe pull-request**  
Added a `close()` method to the popover menu. This will make it possible to programmatically close the menu which is for example needed when having some button as a menu item that doesn't act as a link. 

## **Issue Linking:**  
- [**GitHub**](https://github.com/scania-digital-design-system/tegel/issues/709)

## **How to test**  
1. Implement for example a button that calls the `close()` method provided by the popover menu.
2. Click the button and make sure that the menu closes.

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
    - Not exactly, I wanted an `open` prop instead to align with for example side menu dropdown. But the popover menu has a conflicting variable called `show` that has a slightly other purpose here.
- [x] Events


packages/core/src/components/popover-menu/popover-menu.stories.tsx
```diff
@@ -117,24 +117,20 @@ const Template = ({ menuPosition, icons, fluidWidth }) => {
           <a href="#"> ${icons ? '<tds-icon name="share"></tds-icon>' : ''} Action </a>
         </tds-popover-menu-item>
         <tds-popover-menu-item disabled>
-          <button> ${icons ? '<tds-icon name="share"></tds-icon>' : ''} Action </button>
+          <button> ${icons ? '<tds-icon name="share"></tds-icon>' : ''} Disabled action </button>
```
theJohnnyMe, line 120: Nice catch!

theJohnnyMe (approved):
Thank you for a new feature @ExeAy !
Tested and it works nice. One thing we could do is to add the same feature to popover-canvas too.

ExeAy:
> Thank you for a new feature @ExeAy ! Tested and it works nice. One thing we could do is to add the same feature to popover-canvas too.

Oh, for sure. I'll get to it later today or this week when I find time.

timrombergjakobsson (approved):
LGTM!

Lunkan89 (approved):
Nice work!

PR #733, opened 2024-08-27
feat(tabs): allow custom left padding

timrombergjakobsson:
## **Describe pull-request**  
Enable ability for users to pass a custom left padding value through a prop. 

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-3454`: [CDEP-](https://tegel.atlassian.net/browse/CDEP-3454)    

## **How to test**  
1. Go to Storybook
2. Check in Inline Tabs and Navigation Tabs
3. Set the left padding prop, verify that it only affects the first element


theJohnnyMe:
Reminder on this one:

Screenshot 2024-08-27 at 16 37 50

theJohnnyMe:
@timrombergjakobsson  - Please check this with Laurens:
Do we want to have the possibility to control only the left padding/margin of the first element so users can do something like this:

Screenshot 2024-08-27 at 16 47 05

I am a bit concerned that might be the case. Let's see what he says.

timrombergjakobsson:
> Reminder on this one:
> 
> Screenshot 2024-08-27 at 16 37 50

Of course, the PR is not done so will add.

lau-vt:
It's not quite the intended result. 
The idea is for only the left padding on the parent element to change, meaning 1 padding value controlled, and no left/right padding on the child elements, this would preferably be controlled with a gap setting or something like that. See image as reference. 

Screenshot 2024-08-29 at 14 55 28

theJohnnyMe:
We could drop the margin from the first element and instead set padding-left with the prop you have on parent wrapper, I think it might be the way to go: 
Screenshot 2024-09-04 at 08 59 01

timrombergjakobsson:
> We could drop the margin from the first element and instead set padding-left with the prop you have on parent wrapper, I think it might be the way to go: Screenshot 2024-09-04 at 08 59 01

Good catch @theJohnnyMe! Ill amend!

mistermalm (approved):

packages/core/src/components/tabs/inline-tabs/inline-tabs.tsx
```diff
@@ -32,6 +32,9 @@ export class TdsInlineTabs {
    * If this is set, all Tab changes need to be handled by the user. */
   @Prop({ reflect: true }) selectedIndex: number;
 
+  /** Custom left padding value for the wrapper element. */
+  @Prop({ reflect: true }) leftPadding: number = 0;
+
```
theJohnnyMe, line 37:
```suggestion
  /** Custom left padding value for the wrapper element. */
  @Prop({ reflect: true }) leftPadding: number = 32;

```

To avoid braking change: 
We need to keep the visual component unchanged for current users. 
User that goes from version 1.13.0 to a newer one, should not see any visual difference between therefor this small change.
Please update that and the screenshots for the unit test as they might fail.

theJohnnyMe (changes requested):
A small change to add, otherwise looks good to me.

timrombergjakobsson:
> A small change to add, otherwise looks good to me.

Done!

theJohnnyMe:
@timrombergjakobsson  - Please double-check why Amplify and Playwright are failing. Probably new build and tests update is needed.

timrombergjakobsson:
> @timrombergjakobsson - Please double-check why Amplify and Playwright are failing. Probably new build and tests update is needed.

Strange, new screenshot were generated, it passed locally. So maybe something intermittent. Rerunning the build. Also, Amplify seems to be passing?

theJohnnyMe:
True, then just rerun Playwright on this PR, probably it get confused.

packages/core/src/components/tabs/inline-tabs/inline-tabs.stories.tsx
```diff
@@ -51,27 +51,40 @@ export default {
...
       if: { arg: 'defaultSelectedIndex', eq: 'None' },
     },
+    leftPadding: {
+      name: 'Left Padding',
+      description: 'Sets the custom left padding for the wrapper element. Accepts a numeric value.',
+      control: {
+        type: 'number',
+      },
+      table: {
+        defaultValue: { summary: 0 },
```
theJohnnyMe, line 68:
```suggestion
        defaultValue: { summary: 32 },
```

packages/core/src/components/tabs/inline-tabs/inline-tabs.stories.tsx
```diff
@@ -51,27 +51,40 @@ export default {
...
+      },
+    },
   },
   args: {
     modeVariant: 'Inherit from parent',
     defaultSelectedIndex: 'None',
     selectedIndex: 'None',
+    leftPadding: 0, // Default value for left padding
```
theJohnnyMe, line 76:
```suggestion
    leftPadding: 32, // Default value for left padding
```

packages/core/src/components/tabs/navigation-tabs/navigation-tabs.stories.tsx
```diff
@@ -51,27 +51,39 @@ export default {
...
       if: { arg: 'defaultSelectedIndex', eq: 'None' },
     },
+    leftPadding: {
+      name: 'Left Padding',
+      description: 'Sets the custom left padding for the wrapper element. Accepts a numeric value.',
+      control: {
+        type: 'number',
+      },
+      table: {
+        defaultValue: { summary: 0 },
```
theJohnnyMe, line 68:
```suggestion
        defaultValue: { summary: 32 },
```

packages/core/src/components/tabs/navigation-tabs/navigation-tabs.tsx
```diff
@@ -32,6 +32,9 @@ export class TdsNavigationTabs {
    * If this is set, all Tab changes need to be handled by the user. */
   @Prop({ reflect: true }) selectedIndex: number;
 
+  /** Custom left padding value for the wrapper element. */
+  @Prop({ reflect: true }) leftPadding: number = 0;
+
```
theJohnnyMe, line 37:
```suggestion
  /** Custom left padding value for the wrapper element. */
  @Prop({ reflect: true }) leftPadding: number = 32;

```

mistermalm (approved):

theJohnnyMe (approved):
Nice work @timrombergjakobsson ! 
LGTM

PR #737, opened 2024-09-02
build: release script and wrappers improvement

theJohnnyMe:
## **Describe pull-request**  

- Fixing issues with wrapper packages after the addition of a new component 
- Renaming build commands 
- Updating Angular-17 wrapper to export components project directly in dist, instead of dist/components it used to be. 

## **Issue Linking:**  
- **No issue:** There was an issue with the beta release of tables. The core problem was that wrapper packages did not have the dependency on @scania/tegel package.
- Also, Angular-17 got .npmignore file in build process which removes package.json of component when shipping the code so apps consuming wrappers do not know where to load the module from.

## **How to test**  
1. Create a packed file with `npm pack` of `packages/core`
2. Install that file in one of the wrapper packages. Note: Angular needs it on `packages/angular` and `packages/angular/projects/components` levels
3. Try to run in one of the frameworks. 

## **Additional context**  
Double-check build commands on Amplify before a release.

### DIFF
--- a/.github/workflows/release-all.yml
+++ b/.github/workflows/release-all.yml
@@ -120,8 +120,12 @@ jobs:
         working-directory: packages/angular
         run: npm install @scania/tegel@${{ steps.version.outputs.PACKAGE_VERSION }}
 
+      - name: Angular - Install latest tegel package in components project
+        working-directory: packages/angular/projects/components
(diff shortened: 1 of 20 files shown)


mistermalm (approved):

timrombergjakobsson:
Playwright failed I can see, maybe try and manually envoke it?

Issue #672, opened 2024-06-28
[Feature Request]: Implement a row click callback for table rows.


ExeAy:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [X] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

I need a row click handler to handle events on a row click. If a row click is enabled we might want to style each rows so that the cursor is a pointer.

### Background

In my case I want to route the user to a details page when the user clicks on a row.

The benefits is that it makes some actions easier to do for the users. Instead of locating a button or a menu item they can just click the row directly.

### Idea

E.g. add a tdsClick event on table-body-row. Also change style on the cursor if tdsClick is defined.

### Goals and objectives

A user can click on a row and a desired action is taken.

nathalielindqvist:
Closing this issue since this feature is now available in the latest version of our packages 🙂

PR #760, opened 2024-09-17
feat: add tdsClick event to table rows

ckrook:
## **Describe pull-request**  
This pull-request is regarding adding an tdsClick event to the table rows handle events on a row click.

## **Issue Linking:**  
- **GitHub:** [Include issue link  ](https://github.com/scania-digital-design-system/tegel/issues/672)
- **No issue:** Adding a tdsClick event to the table rows

## **How to test**  
1. Go to storybook and navigate to the table component > basic row.
2. Check in the console that the event is being triggered upon clicking a row.

- [x] All existing tests pass

## **Suggested test steps**
- [x] Events


nathalielindqvist (commented):
Great job, it works well! 🥳  Reading the issue our user submitted they suggest to add a pointer styling to the indicator if the tdsClick event is defined. Is that something we would like to add in this PR?

mJarsater:
> Great job, it works well! 🥳 Reading the issue our user submitted they suggest to add a pointer styling to the indicator if the tdsClick event is defined. Is that something we would like to add in this PR?

I don't think that is gonna work since the tdsClick event is created inside of the component by us and dispatch whether the users adds an event listener or not. I don't know if you wanna consider having the full row as a tabable element though, mostly for accessability. As far as I can tell there is no way for a user to open a expandable row if they can't use their mouse?

timrombergjakobsson:
> > Great job, it works well! 🥳 Reading the issue our user submitted they suggest to add a pointer styling to the indicator if the tdsClick event is defined. Is that something we would like to add in this PR?
> 
> I don't think that is gonna work since the tdsClick event is created inside of the component by us and dispatch whether the users adds an event listener or not.
> 
> I don't know if you wanna consider making the while row tabable though? If a user is expected to press a row to perform an action I think they should also be able to tab to it, for a11y.
> 
> Edit: Making it tabable would create the same problem as abovve though. There is not way (afaik) to know whether an element has an eventlistener attach to it or not. So if you want to make it tablable and use `cursor:pointer` if there is a click event added I think the only option is to create another prop. Having something like `clickable` that the use can add **if** they also attach an eventlistener.
> 
> My suspicion is that the user i coming from React or Angular where the event listener would be passed as a prop and we could then simply check if the prop !== undefined. But I don't think that's gonna work with stencil.

@mJarsater good catch! Thanks a lot for the feedback. I believe that browsers don’t expose APIs to inspect or retrieve a list of event listeners associated with a particular DOM element? So like you said, it won't work. So I guess we would need to manually ask the user to declare intent (like through a clickable prop). The prop tells the component: "Yes, I want this row to be clickable," without relying on detecting event listeners.

ckrook:
https://github.com/user-attachments/assets/75e42142-bf5f-4658-a3a9-e1e9d6eb0004



Is this the look we want to have for the table if the clickable is true? (I'm thinking about the outline style on click / Enter keypress)

packages/core/src/components/table/table-body-row/table-body-row.scss
```diff
@@ -60,3 +60,11 @@
...
   }
 }
+
+:host(.tds-table__row--clickable) {
+  cursor: pointer;
+}
+
+:host(.tds-table__row--clickable:focus-visible) {
+  outline: var(--focus-outline, 2px solid blue);
+}
```
theJohnnyMe, line 70: To answer your question about style of focus state:
Instead, you can use mixin from `packages/core/src/mixins/_focus-state.scss`.

theJohnnyMe:
I can not get tdsClick working either in preview link or in my local after pulling branch. Not sure what is off, please investigate.
I get this error in console, not sure if it is related to tdsClick not emitting on click:

packages/core/src/components/table/table-component-basic.stories.tsx
```diff
@@ -233,91 +245,49 @@ const BasicTemplate = ({
...
+            )
+            .join('')}
       </tds-table-body>
-  </tds-table>`);
+  </tds-table>
+
+  <!-- The script below is just for demo purposes -->
+  <script>
+    // Select all rows
+    const rows = document.querySelectorAll('tds-table-body-row');
```
mJarsater, line 282: Should be a let since it need to reset on each screen load. I think this is where your error is coming from @theJohnnyMe

nathalielindqvist (commented):
Has something happened in this PR? There seems to be 67 files changed, was there a rebase that went out of bounce?

nathalielindqvist (commented):
The tdsClick event returns the tableIds as undefined, is this expected behavior?
Screenshot 2024-10-10 at 13 11 33

Issue #738, opened 2024-09-03
[Bug report]: Tds-dropdown ( multiselect ) component with filter shows no results when at least one item selected


sapostol:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

"@scania/tegel-react": "1.13.0"

### Browser

Chrome

### Framework

React

### Version

"react": "18.2.0"

### Reproduction steps

1. Add component Tds-Dropdown with multiselect and filter (list op options: option 1, option 2, option 3, option 4)
2. Click on the dropdown and select option 1 and option 2
3. Click again on the dropdown label and type "op" 
4. You will see now the "No results" message as I think it tries to search for the entire value of the multiselect at this moment, meaning that instead of just searching for "op" through the options, it searches for "option 1, option 2, op"

### Code example

<TdsDropdown 
                  name='connectedEquipments'
                  multiselect
                  filter
                  modeVariant='secondary'
                  size='md'
                  placeholder='Select equipments'
                  label='Associate equipments'
                  label-position='outside'
                  noResultText='No results'
                  openDirection='up'
                  default-value={['option4']}
>
           <TdsDropdownOption value='option1'>
                Option 1
            </TdsDropdownOption>
           <TdsDropdownOption value='option2'>
                Option 2
            </TdsDropdownOption>
           <TdsDropdownOption value='option3'>
                Option 3
            </TdsDropdownOption>
           <TdsDropdownOption value='option4'>
                Option 4
            </TdsDropdownOption>      
 </TdsDropdown> 
                  

### Screenshots

Add/Copy screenshot here
![tdsDropdownIssue](https://github.com/user-attachments/assets/17b1deb2-5e28-45d4-912a-107883cfcd01)

### Expected behaviour

Component Tds dropdown with multiselect and filter should only search for the last string typed after the last comma, if any option selected, not for the entire value

### Console errors

Add warning/error message

### Contact information

simona.apostol@scania.com

theJohnnyMe:
Fix release in 1.17.0 tegel version.

PR #779, opened 2024-10-01
feat(dropdown): filtering and multi select improvements in dropdown 

ckrook:
## **Describe pull-request**  
We previosly had an issue with the dropdown component when filter:true & multiselect:true.
To address this issue I clear the input when its focused. This makes better UX since the user can search on whatever they want. When the user blur the input they can see which selections have been done.

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-2758`: [CDEP-](https://tegel.atlassian.net/browse/CDEP-2758)    
- **GitHub:** [Issue 738](https://github.com/scania-digital-design-system/tegel/issues/738)  

## **How to test**  
1. Go to Storybook,
2. Navigate to dropdown component
3. In the storybook controls, choose filter: true & multiselect true

- [x] All existing tests pass

## **Suggested test steps**
- [x] Keyboard operability


packages/core/src/components/dropdown/test/multiselect-filter/dropdown.e2e.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    await dropdownListElementOneButton.click();
+    await expect(dropdownListElementOneButton).toHaveAttribute('aria-selected');
+
+    const inputText = page.getByRole('button', { name: /Option 1/ });
+    await expect(inputText).toHaveCount(1);
+
+    await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });
+  });
+});
```
nathalielindqvist, line 28: Should we add a test that checks that the input field has been cleared once focused but still displays the selected option as well? Maybe add another screenshot for that too? :)

timrombergjakobsson: Also, we should start with adding a test for dark mode everytime!

ckrook: I agree with @nathalielindqvist . A test which assure that the input field has been cleared once focus sound good.

ckrook: @timrombergjakobsson, could you clarify why adding a test for dark mode is necessary? From my understanding, the dark mode logic is handled outside the component itself and doesn't directly impact the component's internal logic.

theJohnnyMe:
As discussed, behaviour of using dropdown with just filter has changed. We need to find a way to introduce this fix without altering production.

packages/core/src/components/dropdown/test/multiselect-filter/dropdown.e2e.ts
timrombergjakobsson: @ckrook this was something that we started discussing that we should start adding screenshots for dark mode to verify that it looks correct. Although we have not really been doing it in PR's  lately, so maybe we should have a new discussion regarding if we should do it or not. What do you say @theJohnnyMe?

nathalielindqvist:
> As discussed, behaviour of using dropdown with just filter has changed. We need to find a way to introduce this fix without altering production.

Has the discussion referenced in this comment been addressed @ckrook? I wasn't part of that so I'm not sure which changes are expected, but I can't see that any changes have been pushed after the comment was made? 🙂

ckrook:
> > As discussed, behaviour of using dropdown with just filter has changed. We need to find a way to introduce this fix without altering production.
> 
> Has the discussion referenced in this comment been addressed @ckrook? I wasn't part of that so I'm not sure which changes are expected, but I can't see that any changes have been pushed after the comment was made? 🙂

I just implemented a fix for this :)

nathalielindqvist (commented):
There seems to be an extra screenshot in this PR, the "darwin" one can be deleted :)

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -528,18 +528,25 @@ export class TdsDropdown {
                   ref={(inputEl) => (this.inputElement = inputEl as HTMLInputElement)}
                   class={`${this.labelPosition === 'inside' ? 'placeholder' : ''}`}
                   type="text"
-                  placeholder={this.placeholder}
-                  value={this.getValue()}
+                  placeholder={this.filterFocus ? '' : this.placeholder}
+                  value={this.multiselect && this.filterFocus ? '' : this.getValue()}
                   disabled={this.disabled}
                   onInput={(event) => this.handleFilter(event)}
                   onBlur={(event) => {
```
nathalielindqvist, line 535: When putting filter to true and multiselect to false, choosing an option in the Dropdown and then blurring the input the chosen value is not displayed, only the placeholder. But when opening the Dropdown list again the value is marked as selected. Could it be a conditional rendering of the selected value that requires both filter and multiselect to be true?

ckrook: I'm not able to recreate this as seen in the recording
dropdown

ckrook: Thank you for your explanation regarding this issue. Before proceeding with any further changes, I want to emphasize that the behavior you've described is currently what we have in production.

You can see it demonstrated here: https://tds-storybook.tegel.scania.com/?path=/story/components-dropdown--default&args=filter:true&globals=backgrounds.value:!hex(F9FAFB)

If this is not the desired behavior, I believe it would be better to address this in a new feature request rather than changing it here.

theJohnnyMe: On this one I agree with @ckrook , it is in prod too so lets create another ticket for it. I am afraid that adding new fixes under this one is out of scope and will make a review and test process more complex.

ckrook:
> There seems to be an extra screenshot in this PR, the "darwin" one can be deleted :)
---
I have removed this printscreen!

ckrook:
Added a test which should have existed for the dropdown component with filter set to true. 
This way we have a bigger test coverage on this component.

theJohnnyMe (approved):
Fix works, nice work @ckrook ! 
LGTM.

PS: I did rebase PR with the latest develop when testing on my local.

nathalielindqvist (approved):
Nice!

PR #794, opened 2024-10-10
feat(table): adding tdsClick event to table rows

ckrook:
## **Describe pull-request**  
This pull-request is regarding adding an tdsClick event to the table rows handle events on a row click.

## **Issue Linking:**  
- **GitHub:** [Include issue link  ](https://github.com/scania-digital-design-system/tegel/issues/672)
- **No issue:** Adding a tdsClick event to the table rows

## **How to test**  
1. Go to storybook and navigate to the table component > basic row.
2. Check in the console that the event is being triggered upon clicking a row.

- [x] All existing tests pass

## **Suggested test steps**
- [x] Events


packages/core/src/components/table/table-body-row/table-body-row.tsx
```diff
@@ -61,6 +64,18 @@ export class TdsTableBodyRow {
...
+  /** Event emitted when a row is clicked. */
+  @Event({
+    eventName: 'tdsClick',
+    composed: true,
+    cancelable: false,
+    bubbles: true,
+  })
+  tdsClick: EventEmitter<{
+    tableId: string;
+    rowIndex: number;
```
mJarsater, line 76: might wanna return the click event here aswell.

packages/core/src/components/table/table-body-row/table-body-row.tsx
```diff
@@ -98,12 +129,16 @@ export class TdsTableBodyRow {
...
       <Host
+        tabindex={this.clickable ? '0' : null}
         class={{
           'tds-table__row': true,
           'tds-table__row--selected': this.selected,
           'tds-table__compact': this.compactDesign,
           'tds-table--divider': this.verticalDividers,
+          'tds-table__row--clickable': this.clickable,
         }}
+        onClick={() => this.handleRowClick()}
```
mJarsater, line 140:
```suggestion
        onClick={(e) => this.handleRowClick(e)}
```

mJarsater: and them pass the event to the emitter

nathalielindqvist (commented):
tdsClick returns tableId's as undefined, not sure if this is because there is no id attached to the Table in our example or because of a bug?
Screenshot 2024-10-10 at 13 11 33

mJarsater:
> tdsClick returns tableId's as undefined, not sure if this is because there is no id attached to the Table in our example or because of a bug? Screenshot 2024-10-10 at 13 11 33

The table should be instaniated with a default id:

from table.tsx
```jsx
tableId: string = generateUniqueId();
```

So it seems like this is some sort of bug.

packages/core/src/components/table/table-body-row/table-body-row.tsx
ckrook: Thank you for the feedback @mJarsater  👍

packages/core/src/components/table/table-body-row/table-body-row.tsx
ckrook: Thanks! 👍

ckrook:
I fixed the things that @mJarsater suggested. I do get the id now when clicking the rows. @nathalielindqvist do you still experience this issue?

tdsClick-event

nathalielindqvist (commented):
Now the tableId's return an id! But the rows are clickable when "Clickable rows" is set to false, but when it's set to true I get an error and the event is not fired:
Screenshot 2024-10-10 at 16 50 44

ckrook:
@nathalielindqvist I fixed the script so this error is not getting triggered. :)

packages/core/src/components/table/table-body-row/table-body-row.tsx
```diff
@@ -70,6 +87,24 @@ export class TdsTableBodyRow {
...
+    this.tdsClick.emit({
+      event: e,
+      tableId: this.tableId,
+      rowIndex: rowIndex,
+    });
+  }
+
+  handleKeyDown(e) {
+    if (this.clickable && (e.key === 'Enter' || e.key === ' ')) {
+      e.preventDefault();
```
mJarsater, line 103: Why is this needed?

timrombergjakobsson: Good point @ckrook!

mJarsater: I was more thinking of the `e.preventDefault()`, why is that part needed?

ckrook: The purpose of e.preventDefault() here is to prevent any default behavior associated with the Enter key or space, which could potentially trigger unintended actions in the browser (like submitting a form or scroll the page)

theJohnnyMe (approved):
Works and looks good to me.

timrombergjakobsson (approved):
Nice work @ckrook!

packages/core/src/components/table/table-body-row/table-body-row.tsx
mJarsater: Smart! I'd maybe just leave a comment stating this, just for the future. Feels like a thing that could easily be wrongfully removed in a refactor 🚀

nathalielindqvist (approved):
Awesome work! 💯

Issue #766, opened 2024-09-23
[Bug report]: Tds-dropdown reset() method does not clear the filter value.


furkick:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

@scania/tegel-angular-17: 1.13.0

### Browser

Chrome

### Framework

Angular

### Version

Angular 18

### Reproduction steps

1. Use code example
2. Type 'Option 2' in select to filter
3. Run the reset() method on the select.
4. The filter is not cleared.

### Code example

```
<tds-dropdown 
   name="dropdown" 
   label="Label text" 
   label-position="outside" 
   placeholder="Placeholder" 
   helper="Helper text" 
   size="lg" 
   open-direction="auto" 
   filter
   >
   <tds-dropdown-option value="option-1">Option 1</tds-dropdown-option>
   <tds-dropdown-option value="option-2">Option 2</tds-dropdown-option>
   <tds-dropdown-option value="option-3">Option 3</tds-dropdown-option>
</tds-dropdown>
```

### Expected behaviour

The reset() method should also clear the filter.

### Contact information

christopher.halfpenny@scania.com

theJohnnyMe:
Fixed in tegel 1.17.0

timrombergjakobsson:
[Merged here](https://github.com/scania-digital-design-system/tegel/pull/795)

PR #795, opened 2024-10-14
fix(dropdown): tds dropdown reset method

ckrook:
## **Describe pull-request**  
This pull request addresses issues with the tds-dropdown component's reset() method. The following enhancements and fixes are included:

- Fixed the reset() method to ensure it properly resets the dropdown component, including clearing the filter input when the filter functionality is enabled.

- Modified the setValueAttribute() method so that it removes the value attribute when no value is selected, instead of setting it to the string "null"

- Added unit tests to verify that the reset() method works

- Updated the Storybook example to include a "Reset Dropdown" button

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-3518`:  
https://tegel.atlassian.net/jira/software/projects/CDEP/boards/1/backlog?selectedIssue=CDEP-3518&text=reset

- **GitHub:** Issue 768: 
https://github.com/scania-digital-design-system/tegel/issues/766

## **How to test**  
1. Go to Storybook.
2. Navigate to dropdown component
3. Select an Option.
4. You will then be able to reset the Option by clicking the "Reset Dropdown" button
5. The expected behaviour from clicking this button is to reset the selected options.  

- [x] All existing tests pass
- [x] Behavior available in storybook with documented descriptions (if applicable)

## **Suggested test steps**
- [x] Keyboard operability
- [x] Interactive elements have labels.
- [x] Storybook controls

## **Additional context**  
I can remove the "reset dropdown" button. But I think its good to have on the storybook page just to illustrate the reset functionality. What do you think?


packages/core/src/components/dropdown/dropdown.stories.tsx
```diff
@@ -294,15 +294,19 @@ const Template = ({
               îles Åland
             </tds-dropdown-option>
         </tds-dropdown>
+        <button onClick="resetDropdown()">Reset Dropdown</button>
     </div>
 
     <script>
-      dropdown = document.querySelector('tds-dropdown')
+      const dropdown = document.querySelector('tds-dropdown');
```
mJarsater, line 301:
```suggestion
      let dropdown = document.querySelector('tds-dropdown');
```

Just so storybook can reassign it when rerendering

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -429,7 +429,11 @@ export class TdsDropdown {
 
   private setValueAttribute = () => {
     if (this.value?.toString() === '') this.value = null;
-    this.host.setAttribute('value', this.value?.map((val) => val).toString() ?? null);
+    if (this.value === null) {
+      this.host.removeAttribute('value');
+    } else {
+      this.host.setAttribute('value', this.value?.map((val) => val).toString());
+    }
   };
```
mJarsater, line 437: Should probably check if it null first, if not null the the guards "?" aren't needed.

Example: (haven't tried this, just psuedo code ish)
```jsx
  private setValueAttribute = () => {
    if(!value || this.value?.toString() === ''){
         this.host.removeAttribute('value');
         this.value = null;
    }  else {
      this.host.setAttribute('value', this.value.map((val) => val).toString());
      this.value = value;
    }
  };
```

mJarsater: How come this `setValueAttribute ` method doesn't have any args? Is it used by us internally or by users?

ckrook: I’m a bit unsure about your comment regarding internal usage, as a private function is not publicly accessible. Additionally, regarding the arguments, the function was created 12 months ago without any parameters.

mJarsater: I get that the `private` methods are exposed via the API to our users. But the `reset` method uses a `internatlReset` which then uses `setValueAttribute` so it is sort of exposed. But the `setValueAttribute` then just updates the attribute on the element. If that is the case I think the code I suggested above should be something more like:


```jsx
  private setValueAttribute = () => {
    if(!value || this.value?.toString() === ''){
         this.host.removeAttribute('value');
    }  else {
      this.host.setAttribute('value', this.value.map((val) => val).toString());
    }
  };
```

Since it should **only** update the value attribute and not property.

mJarsater: If the value should also be "reset", I'd suggested either renaming this private method and remove the value here aswell. Or to create a setValue method (if it doesn't already exist) and call that after this method has been called, with a null or empty value.

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -429,7 +429,11 @@ export class TdsDropdown {
 
   private setValueAttribute = () => {
     if (this.value?.toString() === '') this.value = null;
-    this.host.setAttribute('value', this.value?.map((val) => val).toString() ?? null);
+    if (this.value === null) {
+      this.host.removeAttribute('value');
+    } else {
+      this.host.setAttribute('value', this.value.map((val) => val).toString());
+    }
   };
```
mJarsater, line 437: Should probably be: 

```jsx
  private setValueAttribute = () => {
    if (this.value === null || this.value?.toString() === '') {
      this.host.removeAttribute('value');
    } else {
      this.host.setAttribute('value', this.value.map((val) => val).toString());
    }
  };
```
or maybe even:
```jsx
  private setValueAttribute = () => {
    if (!this.value) {
      this.host.removeAttribute('value');
    } else {
      this.host.setAttribute('value', this.value.map((val) => val).toString());
    }
  };
```

mJarsater, line 437: Should probably be: 

```jsx
  private setValueAttribute = () => {
    if (this.value === null || this.value?.toString() === '') {
      this.host.removeAttribute('value');
    } else {
      this.host.setAttribute('value', this.value.map((val) => val).toString());
    }
  };
```
or maybe even:
```jsx
  private setValueAttribute = () => {
    if (!this.value) {
      this.host.removeAttribute('value');
    } else {
      this.host.setAttribute('value', this.value.map((val) => val).toString());
    }
  };
```

mJarsater: And `this.value = null;` that is currently being done on line 431 should not be done in this function. Since it's called `setValueAttribute` I think that is all it should do, not update component state.

packages/core/src/components/dropdown/dropdown.tsx
ckrook: Thank you for the explanation @mJarsater , grateful for the input! 🙌

packages/core/src/components/dropdown/test/basic/dropdown.e2e.ts
```diff
@@ -47,4 +47,35 @@ test.describe.parallel('tds-dropdown-basic', () => {
...
+    // Call the reset() method on the dropdown component
+    await page.evaluate(() => {
+      const dropdownnew = document.querySelector('tds-dropdown');
+      dropdownnew.reset();
+    });
+
+    // Verify that the dropdown has been reset to its initial state
+    await expect(placeholder).toHaveText(initialPlaceholderText);
+    await expect(dropdown).not.toHaveAttribute('value');
+  });
```
mJarsater, line 80: I think you also need to check the value property here

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -428,8 +428,11 @@ export class TdsDropdown {
...
 
   private setValueAttribute = () => {
-    if (this.value?.toString() === '') this.value = null;
```
mJarsater, line 431: Before we were doing `this.value = null;`, just need to make sure this is being done somewhere else even after this change.

ckrook: This is happening inside internalReset function right? @mJarsater I do not think we need to add it elsewhere. Am I wrong?

packages/core/src/components/dropdown/test/basic/dropdown.e2e.ts
ckrook: I'm encountering an issue where checking for the value breaks the test. I suspect it's because if the attribute doesn't exist, it can't retrieve any value. Do you have any suggestions on how to handle this in the code? I'm open to ideas!

mJarsater: Could something like: `await expect(dropdown).toHaveJSProperty('value', null);` work? 

First check that the attribute isn't present on the element and then check that it doesn't have the jsproperty?

ckrook: Unfortunately this do not work, I do encounter a timeout with this.

mJarsater: Alright, what is the stacktrace?

timrombergjakobsson:
@ckrook noticed the branch is not in the `fix/`-namespace, could we move it correctly? You could do `git branch -m fix/branch-name` locally and push it. Would correct it!

packages/core/src/components/dropdown/test/basic/dropdown.e2e.ts
timrombergjakobsson: Maybe verify for undefined instead?

`await expect(dropdown).toHaveJSProperty('value', undefined);`

timrombergjakobsson: @mJarsater true, but as we are in:
```
private setValueAttribute = () => {
  if (!this.value || this.value?.toString() === '') {
    this.host.removeAttribute('value');
  } else {
    this.host.setAttribute('value', this.value.map((val) => val).toString());
  }
};
```
when this.value is null, it's removing the value attribute from the host element entirely, rather than setting it to "null" or undefined. Or? The value property doesn't exist on the element after reset?

ckrook: Seems like 
`await expect(dropdown).toHaveJSProperty('value', undefined);`
Works. so yes. It looks like it is set to undefined.

mJarsater: @timrombergjakobsson ah yes, nice!

mistermalm (approved):

timrombergjakobsson (approved):

PR #799, opened 2024-10-17
fix(table-body-row): moved comment that caused stencil build error

Lunkan89:
## **Describe pull-request**  
moved comment that caused stencil build error.

## **How to test**  
1. fetch and ceckout branch
2. `npm run build-all` in root

### DIFF
--- a/packages/core/src/components/table/table-body-row/table-body-row.tsx
+++ b/packages/core/src/components/table/table-body-row/table-body-row.tsx
@@ -71,8 +71,8 @@ export class TdsTableBodyRow {
     cancelable: false,
     bubbles: true,
   })
+  // return clickevent
   tdsClick: EventEmitter<{
-    // return clickevent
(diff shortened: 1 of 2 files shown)


packages/core/src/components/table/table-body-row/table-body-row.tsx
```diff
@@ -71,8 +71,8 @@ export class TdsTableBodyRow {
     cancelable: false,
     bubbles: true,
   })
+  // return clickevent
   tdsClick: EventEmitter<{
-    // return clickevent
     event: MouseEvent;
```
mJarsater, line 76: Just an FYI, this could also be a KeyboardEvent. Since it can be "pressed" by space or enter while tabbing.

timrombergjakobsson (approved):

nathalielindqvist (approved):

PR #803, opened 2024-10-18
fix: install core in all wrappers

mistermalm:
Install Core in all wrappers to make sure all dependencies are there


ckrook (approved):

theJohnnyMe (approved):
NIce, lets give it a shot ;)

PR #828, opened 2024-11-04
ci(holy-grail): improvements

theJohnnyMe:
## **Describe pull-request**  

- Separating version bump in all wrappers to separate job 
- Removing committing and push from wrappers 
- The branch dept is set to 1 for wrappers as they do not need other branches but just release branch

## **Issue Linking:**  
- **No issue:** In last release we had issue where not every wrapper committed changes done in`package.json` and `package-lock.json` files. Now, as bumping is extracted from all of the wrappers and made into separate step, I think we should have all the versions in place after successful release.

## **How to test**  
1. Check if code makes sense.

## **Additional context**  
Partial jobs might be tried with ACT. So feel free to test it.

### DIFF
--- a/.github/workflows/holy-grail.yml
+++ b/.github/workflows/holy-grail.yml
@@ -142,9 +142,55 @@ jobs:
             echo "No changes to commit"
           fi
 
+  #BUMP WRAPPERS TO NEW VERSION
+  bump-wrappers-to-new-version:
+    needs: release-core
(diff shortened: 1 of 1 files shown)


mistermalm (approved):

Lunkan89 (commented):
I will approve this one so we can hopefully give it a try this upcoming release.

Lunkan89 (approved):

Issue #790, opened 2024-10-08
[Bug report]: Popover menu throws error on creation. Cannot read properties of undefined ('update')


joakim-malmstrom-scania:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

@scania/tegel-angular: ^1.15.0

### Browser

Chrome

### Framework

Angular

### Version

15

### Reproduction steps

1. See code below for steps on creating the popover
2. A directive is created
3. The directive is injected in a .html file
4. The popover component is created on right-clicking an element

### Code example

```
@Directive({
  selector: '[popover-ref]',
})

export class PopoverRefDirective {
  constructor(public viewContainerRef: ViewContainerRef) { }
}
```
```
// component.html
<!-- Popover Menus -->
<ng-template popover-ref></ng-template>
```
```
// component.ts
  @ViewChild(PopoverRefDirective) popoverRef: PopoverRefDirective;

  ngAfterViewInit() {
    this.popoverMenuLoaderService.setViewContainer(this.popoverRef);
  }
```
```
// openMenu is called on right click
  openMenu<T, Y>(popoverMenuComponent: Type<PopoverMenuComponent<T, Y>>, data?: T): EventEmitter<Y> {
    const component = this.createPopoverMenuComponent(popoverMenuComponent, data);

    return component.instance.show();
  }

  createPopoverMenuComponent<T, Y>(popoverMenuComponent: Type<PopoverMenuComponent<T, Y>>, data?: T): ComponentRef<PopoverMenuComponent<T, Y>> {
    const component: ComponentRef<PopoverMenuComponent<T, Y>> = this.container.viewContainerRef.createComponent(popoverMenuComponent);

    component.instance.data = data;

    return component;
  }
```
```
// created component.ts
  show(): EventEmitter<any> {
    this.popoverRef.nativeElement.show = true;
    return this.popoverClosed;
  }
```

### Screenshots

![image](https://github.com/user-attachments/assets/5ac180ea-fda9-4b8c-a6ce-5f4ca2a80fc5)

### Expected behaviour

Upon creation, popover component should not throw: TypeError: Cannot read properties of undefined (reading 'update'). 

### Console errors

[...]

PR #839, opened 2024-11-08
feat(popover-core):  prevent internal logic

theJohnnyMe:
## **Describe pull-request**  
The user has initially reported this as an error, but digging further, I think this is a feature with a small fix.

Small fix:
 - Currently, we have an issue with the component's lifecycle methods. The component is to be updated before it is initialized, resulting in console.log errors. It can be seen in demo pages too on the form page. 
  - It was fixed by moving the initialization to the connectedCallback.

Feature:
Digging into the users issue recreated on [Stackblitz example](https://stackblitz.com/edit/stackblitz-starters-kvkg2u?file=src%2Fcomponents%2Fbict-popover-menu.component.ts) it was clear that user is taking full control over when a component should appear and there is no real use of `selector`or `referenceEl` however, component expect those to position itself. I have changed the core so that if user has not passed any of those props component will disable internal logic.

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-`: [CDEP-3532](https://tegel.atlassian.net/browse/CDEP-3532)    
- **GitHub:** https://github.com/scania-digital-design-system/tegel/issues/790

## **How to test**  
1. Try example that user has send, observe if you get errors in console. 
2. Try installing this [beta version](https://www.npmjs.com/package/@scania/tegel-angular-17/v/1.18.0-popover-menu-beta.4) in the same example and set disableLogic prop to true on tds-popover-menu.
3. Try installing [this beta](https://www.npmjs.com/package/@scania/tegel-angular-17/v/1.18.0-popover-menu-beta.4) version in the same example and remove selector and referenceEl prop from tds-popover-menu.
4. If no2 and no3 work without posting errors in console.log, we are good to go.

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

[CDEP-3532]: https://tegel.atlassian.net/browse/CDEP-3532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/core/src/components/popover-canvas/popover-canvas.tsx
```diff
@@ -38,6 +38,9 @@ export class TdsPopoverCanvas {
   /** Array of modifier objects to pass to popper.js. See https://popper.js.org/docs/v2/modifiers/ */
   @Prop() modifiers: Object[] = [];
 
+  /** Decides if the popover should disable its internal logic. Leaving selector or referenceEl empty will disable logic too. */
+  @Prop() disableLogic: boolean = false;
+
```
mJarsater, line 43: Just a suggestion, I would instead look into creating logic for controlled vs uncontrolled states. 

Separate the props to defaultOpen and open. 

If none of the props are used we do all the opening/closing in the component (uncontrolled.)

If a user adds a false/true (not undefined) value for `open` prop we never do any internal logic and just opens/closes the Popover when the open prop changes value. (controlled)

If a users set defaultOpen to true the Popover is open by default and we do all the open/closing in the component (uncontrolled)

This would mirror Reacts uncontrolled vs controlled pattern. See: https://www.freecodecamp.org/news/what-are-controlled-and-uncontrolled-components-in-react/

If I don't want to control this state I would do:

```jsx
<tds-popover selector={'#selectorId'}>
   // My popover content
</tds-popover>
``` 

If I don't want to control this state, but I want it open by default I would do:

```jsx
<tds-popover selector={'#selectorId'} default-open>
   // My popover content
</tds-popover>
``` 

If I as a user wants to control the open/close state I would do: 

```jsx
<tds-popover open={myTrueOrFalseVaraibleThatWillUpdateInRuntime}>
   // My popover content
</tds-popover>
``` 

For the controlled one I would recommend exposing a tdsOpenChange event that fires when a user presses outside the Popover (so when it would close), and the user can listen to this and decide if they want to close the Popover.

mJarsater: Sorry for the long comment. But i think this would be a good approach for all interactive components. To try and think of how a users could either control it fully, not control it at all or maybe just set one initial state.

packages/core/src/components/popover-core/popover-core.tsx
```diff
@@ -225,15 +232,21 @@ export class TdsPopoverCore {
     this.popperInstance?.destroy();
   }
 
-  componentDidLoad() {
+  connectedCallback() {
+    if ((this.selector === undefined && this.referenceEl === undefined) || this.disableLogic) {
+      this.disableLogicState = true;
+      console.warn('TDS-POPOVER-CORE: Popover internal logic disabled.');
+      return;
+    }
```
theJohnnyMe, line 240: @mJarsater  - That is a good approach and something we could do for other components. Over here I am on trail of what are you saying - if a user has provided neither selector nor reference, then component logic goes down. A recommendation was to use an additional prop for those users that want to be 100% sure, aka setting it explicitly. 

The thing is that Popover core is used by three other components so I have to check how much can be done with other props before making braking change, but overall I like idea of controlled and uncontrolled behaviours of component.

packages/core/src/components/popover-core/popover-core.tsx
```diff
@@ -225,15 +232,21 @@ export class TdsPopoverCore {
...
+
     this.initialize({
       referenceEl: this.referenceEl,
       trigger: this.trigger,
     });
   }
 
   componentDidRender() {
-    if (this.isShown && !this.renderedShowValue) {
+    if (this.isShown && !this.renderedShowValue && !this.disableLogicState) {
```
theJohnnyMe, line 249: @mJarsater  - State is made is other parts of components logic can be disabled.

mJarsater: Yeah but I think that rather than adding a prop that disables everything I would take a step back and see if there is a pattern you could follow for this and the same type of components. 

Ideally you'd only have to implement controlled vs uncontrolled in `popover-core` and then just forward the props in other component.

ckrook (approved):
I tested the changes in Stackblitz, look good. Nice job!
nodding-yes

theJohnnyMe:
Feeback from user: 
Screenshot 2024-11-12 at 09 56 39

packages/core/src/components/popover-core/popover-core.tsx
mJarsater: I think what I am aiming at is that a prop for disabling features in not very intuitive (to me atleast). I think it's a better approach to try and follow coding conventions and implement controlled and uncontrolled states. As far as I can tell this would be non-breaking. 

```
connectedCallback() {
    if ((this.selector === undefined && this.referenceEl === undefined) || this.show !== undefined) {
      return;
    }
```

theJohnnyMe: I checked the code and did some changes:

1. Currently, we can not put the `show` prop as a condition that enables a controlled state - aka makes component escape initialization. The reason for that is that other components depend on it. Header's Dropdown and App Launcher break as they need to do initialization to place themselves at the right spot. 

2. I have added the `defaultOpen` prop, which was a good idea. This way, users can put components to be initially visible, without losing closing on click outside etc. There is a rule that `open` prop will take precedence if set.

3. I have added a bit of text in the readme to explain the controlled and uncontrolled state of these components, however, I am not sure if users will have many benefits from the controlled state as in that case, I am certain that position won't work, you need to provide selector/referenceEl, otherwise, the component will be lost in woods. Aka - selector and reference are needed to make this component work as expected. Therefor I think that controlling ´open´ prop is the most we can do for now so I did update description of prop of Popover Canvas and Popover Menu to clarify it better. Tooltip has its ways of handling it. 

Good discussion with you Mr. @mJarsater  ;)

packages/core/src/components/popover-canvas/readme.md
```diff
@@ -27,20 +27,44 @@ use the `referenceEl` prop rather than the `selector` the referenced element can
 </TdsPopoverCanvas>
 
 ```
+
+## Controlled vs Uncontrolled
+
+- **Controlled**: 
+  How does component enter controlled mode?
+  - Leaving `selector` or `referenceEl` prop empty
```
mJarsater, line 35: I think the users should still be able to add a `referenceEl` to get the correct positioning.

This is how I imagine it:

```jsx
<button id="my-id" onClick={() => showPopover = !showPopover}>
 Open/close popover
</button>


<tds-popover-canvas selector="#my-id" show={showPopover}>
 // Content
</tds-popover-canvas>
```

So I think controlled should **only** case about if `show` is not null. If it is set to `true` or `false` the internal logic should be completely ignore appart from setting the correct reference/selector. 

Maybe this isn't possible with the current setup and if so I would say ignore my silly comments for now. But I do this the long term goal for all "show/hide-components" to have either a controlled or uncontrolled state.

theJohnnyMe: I get you, I think I need to rephrase it. Think is if users do not pass the selector or reference, a component can not initialize properly, therefore we have to block logic from running, otherwise, it will spit out errors in the console, first for missing reference and then for missing instance to run update script from popper. 

So, users can use selector/referenceEl and open prop today too, I think what I need to clarify is that for the right positioning of component, selector/reference are mandatory.

mJarsater: Ah okey! Yeah that sounds good. If no referenceEl or selector is passed we shouldn't really render anything, right? Instead just log an error.

theJohnnyMe: In that case, user will get a warning that component's internal logic is disabled. Component wont show up until user toggles open to true.

packages/core/src/components/popover-canvas/readme.md
```diff
@@ -27,20 +27,51 @@ use the `referenceEl` prop rather than the `selector` the referenced element can
...
+
+Example:
+
+```html
+<button id="trigger" onClick="{() => showPopover = !showPopover}">Open Popover</button>
+<tds-popover-canvas id="popover" selector="#trigger" open="showPopover">
+  <h2 class="tds-headline-02 tds-u-mt0">A Popover Canvas!</h2>
+</tds-popover-canvas>
+
+```
```
mJarsater, line 56: I think these should be swapped. When I as a user controlls the open state it is controlled. When I rely on the default logic it is uncontrolled.

packages/core/src/components/popover-core/popover-core.tsx
```diff
@@ -152,7 +157,7 @@ export class TdsPopoverCore {
     this.setIsShown(true);
   }.bind(this);
 
-  private handleHide = function handleShow(event) {
+  private handleHide = function handleHide(event) {
```
mJarsater, line 160:
```suggestion
  private handleHide = (event) =>  {
```

packages/core/src/components/popover-core/readme.md
```diff
@@ -1,23 +1,44 @@
 # tds-core-popover
 
+## Controlled vs Uncontrolled
 
+- **Controlled**: 
+  How does component enter controlled mode?
+  - Leaving `selector` or `referenceEl` prop empty
+
```
mJarsater, line 8:
```suggestion
- **Controlled**: 
  How does component enter controlled mode?
  - Setting the `show` prop to either `true` or `false`
```

packages/core/src/components/popover-menu/readme.md
```diff
@@ -32,20 +32,53 @@ use the `referenceEl` prop rather than the `selector` the referenced element can
...
+```html
+<button id="trigger" onClick="{() => showPopover = !showPopover}">Open Popover</button>
+<tds-popover-menu id="popover" selector="#trigger" open="showPopover">
+  <tds-popover-menu-item>
+    <a href="#">Action</a>
+  </tds-popover-menu-item>
+</tds-popover-menu>
+
+```
+
```
mJarsater, line 66: I think its the same here: https://github.com/scania-digital-design-system/tegel/pull/839/files#r1840354351

packages/core/src/components/tooltip/readme.md
```diff
@@ -28,20 +28,44 @@ use the `referenceEl` prop rather than the `selector` the referenced element can
 </TdsTooltip>
 
 ```
+## Controlled vs Uncontrolled
+
+- **Controlled**: 
+  How does component enter controlled mode?
+  - Leaving `selector` or `referenceEl` prop empty
```
mJarsater, line 35: https://github.com/scania-digital-design-system/tegel/pull/839/files#r1840361006

packages/core/src/components/popover-core/popover-core.tsx
theJohnnyMe: I will leave it as it is for now.

packages/core/src/components/popover-canvas/readme.md
theJohnnyMe: My bad :(

mJarsater (approved):
Really great stuff! Maybe double check with the user who tested the other alpha that this solves their issue?

Some things for future improvements:
On line [96](https://github.com/scania-digital-design-system/tegel/blob/5a3943370128dac8cfa54a3b9f9e6201cecd9eb0/packages/core/src/components/popover-core/popover-core.tsx#L96) in `popover-core.tsx` we are detecting clicks outside of the popover. When this runs I think we should omit an event, something line `tdsShowChange` that the users can listen to. This would allow for them to implement a controlled component that still closes when they click outside, example:

```jsx
<button 
   id="button-id" 
   onClick={() => showVariable = !showVariable}
>
   open/close popover
</button>

<tds-popover show={showVariable} onTdsShowChange={() => showVariable = !showVariable}>
   // Content
</tds-popover>
```


This would allow the users to control the open state based on other stuff that pressing a trigger, maybe it's when navigating to some other page, or when bluring focus from a textfield, what ever really. But they would also get the added option of not having to do the click outside logic on their own. 


PS. the event should also be preventable so users can stop the popover from closing when a press outside occurs.


PPS. I would consider changing `defaultOpen` to `defaultShow` just to align with the show prop.

theJohnnyMe:
> Really great stuff! Maybe double check with the user who tested the other alpha that this solves their issue?
> 
> Some things for future improvements: On line [96](https://github.com/scania-digital-design-system/tegel/blob/5a3943370128dac8cfa54a3b9f9e6201cecd9eb0/packages/core/src/components/popover-core/popover-core.tsx#L96) in `popover-core.tsx` we are detecting clicks outside of the popover. When this runs I think we should omit an event, something line `tdsShowChange` that the users can listen to. This would allow for them to implement a controlled component that still closes when they click outside, example:
> 
> ```js
> <button 
>    id="button-id" 
>    onClick={() => showVariable = !showVariable}
> >
>    open/close popover
> </button>
> 
> <tds-popover show={showVariable} onTdsShowChange={() => showVariable = !showVariable}>
>    // Content
> </tds-popover>
> ```
> 
> This would allow the users to control the open state based on other stuff that pressing a trigger, maybe it's when navigating to some other page, or when bluring focus from a textfield, what ever really. But they would also get the added option of not having to do the click outside logic on their own.
> 
> PS. the event should also be preventable so users can stop the popover from closing when a press outside occurs.
> 
> PPS. I would consider changing `defaultOpen` to `defaultShow` just to align with the show prop.

Good idea for defaultShow, my memory refreshed just now, all the time I was thinking it is `open` :( 
Event is a good idea, but I will probably add it in other PR, lets see if this one make it for release today.

timrombergjakobsson (approved):

Issue #833, opened 2024-11-07
[Bug report]: Tds-button disabled state only disables internal button and not tds-button wrapper


scania-mharlin:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [X] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [X] I have followed the installation guide.

### Package versions

"@scania/tegel-angular": "1.18.0",

### Browser

Chrome

### Framework

Angular

### Version

Angular 18

### Reproduction steps

1. Add component tds-button
2. Add click event listener to button and set disabled to be true
3. Add padding to the button
4. Click on the padding of the button
5. The click event is triggered even though the button is disabled 

### html
```
  <tds-button
    style="padding: 50px; background-color: fuchsia"
    [disabled]="true"
    type="button"
    variant="primary"
    (click)="clickButton()"
    text="test">
  </tds-button>
```

### script
```
  clickButton(): void {
    console.log("button clicked");
  }
  ```

### Screenshots

Add/Copy screenshot here

### Expected behaviour

This is mainly a problem for our test automation where our automation finds and clicks on the tds-button even though the button inside is still in a disabled state. This also means it's possible to trigger the click event of the tds-button programmatically when it's in a disabled state. 

## expected
It should not be able to trigger the click event for the Tds-button when it is in a disabled state

### Console errors

Add warning/error message

### Contact information

magnus.harlin@scania.com

theJohnnyMe:
The code that fixes this issue is released in version 1.20.0.

PR #855, opened 2024-11-19
fix(button): host in disabled state

theJohnnyMe:
## **Describe pull-request**  
Added `disabled` attribute on host level and `pointer-events: none` once component is in disabled state so no event fires when the component is in that state. 
Additionally, I cleaned up the focus state so it uses mixin and only in one place, other places seemed to be repetitive without any value. 

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-`: [CDEP-3600](https://tegel.atlassian.net/browse/CDEP-3600)    
- **GitHub:** https://github.com/scania-digital-design-system/tegel/issues/833

## **How to test**  
1. Read the GitHub issue and how to recreate it.
2. Go to preview link.
3. Try to recreate the issue.
4. Try the same in production Storybook.
5. Preview link should behave as expected.

- [x] All existing tests pass

## **Suggested test steps**
- [x] Keyboard operability
- [x] Storybook controls
- [x] Events

## **Additional context**  
As `pointed-events: none* rule are used, not-allowed cursor style is not possible to be used anymore. So test is updated to reflect it. 

[CDEP-3600]: https://tegel.atlassian.net/browse/CDEP-3600?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/button/button.scss
+++ b/packages/core/src/components/button/button.scss
@@ -181,8 +181,12 @@ button {
   }
 }
 
-:host([disabled]) {
+:host([disabled]:active) {
   pointer-events: none;
+}
+
+:host([disabled]) button {
+  cursor: not-allowed;
 }
 
 :host(tds-button[fullbleed]) {
--- a/packages/core/src/components/button/button.scss
+++ b/packages/core/src/components/button/button.scss
@@ -37,19 +37,12 @@ button {
   padding: var(--tds-spacing-element-20);
   border: 1px solid;
 
-  &:disabled,
-  &.disabled {
-    cursor: unset;
-  }
-
   &:focus {
     outline: none;
   }
 
   &:focus-visible {
-    // TODO: decide on what tod o with these @include tds-focus-state;
-    outline: 2px solid var(--tds-blue-400);
-    outline-offset: -2px;
+    @include tds-focus-state;
   }
 
   &.xs {
@@ -57,12 +50,6 @@ button {
     height: $btn-xs-height;
     font-size: 12px;
     border-radius: 2px;
-
-    &:focus-visible {
-      // TODO: decide on what tod o with these @include tds-focus-state;
-      outline: 2px solid var(--tds-blue-400);
-      outline-offset: -2px;
-    }
   }
 
   &.sm {
@@ -75,12 +62,6 @@ button {
 
     &.only-icon {
       padding: $btn-sm-only-icon-padding;
-    }
-
-    &:focus-visible {
-      // TODO: decide on what tod o with these @include tds-focus-state;
-      outline: 2px solid var(--tds-blue-400);
-      outline-offset: -2px;
     }
   }
 
@@ -95,10 +76,6 @@ button {
     &.only-icon {
       padding: $btn-md-only-icon-padding;
     }
-
-    &:focus-visible {
-      @include tds-focus-state;
-    }
   }
 
   &.lg {
@@ -111,10 +88,6 @@ button {
 
     &.only-icon {
       padding: $btn-lg-only-icon-padding;
-    }
-
-    &:focus-visible {
-      @include tds-focus-state;
     }
   }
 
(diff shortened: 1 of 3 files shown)


ckrook:
@theJohnnyMe I tested this in production and could not find the issue in Storybook.
Although your fix seems to work, it would be good for me to be able to recreate the issue in live environment

nathalielindqvist:
> @theJohnnyMe I tested this in production and could not find the issue in Storybook. Although your fix seems to work, it would be good for me to be able to recreate the issue in live environment

The TdsClick event fires when clicking on padding attached to the tds-button element, even if the Button is disabled. Made a little screen recording of the issue, hope it helps! :)


https://github.com/user-attachments/assets/6baaa4e7-a5e0-4859-a5f2-867af19fd189

packages/core/src/components/button/button.scss
```diff
@@ -209,12 +181,8 @@ button {
   }
 }
 
-:host(tds-button[disabled='true']) {
+:host([disabled]) {
   pointer-events: none;
```
mJarsater, line 185: Just a thought here, isn't this also gonna disabled any hover interactions on the `tds-button`. If you have a disabled button that on hover should show a tooltip or something to highlight why it is disabled? Could be worth investigating.

nathalielindqvist (approved):
Nice work! 👏

theJohnnyMe:
@ckrook & @nathalielindqvist please check if you can drop cache or use inspector in private/incognito mode.

Here is what I see:  
https://github.com/user-attachments/assets/949c9546-1a1c-4889-b5c8-2203e50f7ebe

packages/core/src/components/button/button.scss
theJohnnyMe: Hmm.. that is a good point and I will check it. Wonder what happens now if we attach Tooltip to it?

nathalielindqvist:
> @ckrook & @nathalielindqvist please check if you can drop cache or use inspector in private/incognito mode.
> 
> Here is what I see: https://github.com/user-attachments/assets/949c9546-1a1c-4889-b5c8-2203e50f7ebe

Sorry, I was a bit unclear in my comment - the screen recording I posted was of production, the preview link in this PR gives me the same result as your screen recording 👍

packages/core/src/components/button/button.scss
theJohnnyMe: I am afraid you are right @mJarsater, here is a small example with this code in: https://stackblitz.com/edit/vitejs-vite-bzyiej?file=package.json

I have no idea how I would disable clicking on Host level, I tried with onClick function that preventsDeafult and stopPropagation, but that did not work. Do you have some idea?

theJohnnyMe: I got idea, keep there...

mJarsater: Maybe either expose a `tdsClick` event that you can control,or add an eventListener to the host  in some lifecycle hook?

theJohnnyMe: Code updated!

mJarsater: Nice, maybe add a test to validate that this doesn't trigger a click?

theJohnnyMe: @mJarsater  - I looked into this, but could not find a proper way. 

Playwright documentation point to use `toBeDisabled`, already used in the button.e2e.ts file. Another one is to check cursor CSS or buttons CSS, which we do too.

I was thinking of implementing a small increment in the span element that exists next to the button, so if you click on the button, the number would go from 1 to 2, however, if disabling works, we should expect the number to remain the same - aka initial value.
When doing so, Playwright throws a timeout error as it can not click on the button, and the test breaks - which is expected from a logical view, but not if we consider it a mistake that breaks unit tests. 

That said, I will leave the test case as is.

mJarsater: I think `toBeDisabled` might only work for native elements. I'd say what we really wanna know is if a disabled button emits a click events. Might need to "force" a click: https://playwright.dev/docs/input#forcing-the-click

PR #856, opened 2024-11-19
ci: smoke test improvements

mistermalm:
## **Describe pull-request**  
This PR includes a different approach to make sure the right version of core i installed in wrappers. By making use of "npm pack" and upload/download artifacts.

## **How to test**  
You can run it locally using this act script:
`act -W .github/workflows/smoketest-webhook.yml -P ubuntu-latest=catthehacker/ubuntu:act-latest -s GITHUB_TOKEN={YOUR_GIT_TOKEN}`

- [ ] All existing tests pass

## **Additional context**  
When I run this I got the impression it's a bit slower, but more reliable, since we now make sure we get the right version. Thoughts?


theJohnnyMe (commented):
This is what I got after running it on my local, is this expected?

Screenshot 2024-11-20 at 16 25 31

theJohnnyMe:
I think the command in "How to test" might be wrong as file does not exist so I used this one:

`act --container-architecture linux/amd64 -W .github/workflows/smoke-test-failure-webhook.yml -P ubuntu-latest=catthehacker/ubuntu:act-latest -s GITHUB_TOKEN=*******`

Is that correct one?

theJohnnyMe (approved):
Lets test this one in action

Issue #778, opened 2024-09-30
[Feature Request]: add focus property


Luc-Siat:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [X] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

add focus property to focusable element to allow to programmatical focus.

### Background

We have a few component that we focus programmatically (eg. a input that is disabled until the edit pen is clicked and then we want to focus this input to avoid an additional click)

### Idea

add a focus property similar to the reset() on the tds-dropdown that we can call to focus the component.

### Goals and objectives

improve UX

PR #867, opened 2024-11-26
feat: add a method to programatically focus and open the dropdown element

NOMADE55:
## **Describe pull-request**  
Adds a method to focus and open dropdown element programmatically.

## **Issue Linking:**  
- **GitHub:** https://github.com/scania-digital-design-system/tegel/issues/778

## **How to test**  
Whether you're using angular, react or web components
1. Get dropdown's element reference.
2. Invoke the `focusElement()` method.
3. Dropdown should be focused and opened.

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Additional context**  
An important note, method is named `focusElement` as `focus` is a reserved name. 
I'm not really keen on this name, falling victim of the [second hardest problem in programming](https://martinfowler.com/bliki/TwoHardThings.html), so feel free to suggest/change it to something more Tegel-ly.


packages/core/src/components/dropdown/dropdown.tsx
mJarsater, line 1: I think a more scalable approach to these exposed methods is to create controlled vs uncontrolled open/closed state in the dropdown.

mJarsater: To have these methods forces users to grab a reference to the element. If we instead had the open/closed state be either controlled or uncontrolled they could simply toggle the prop and we'd update.

NOMADE55: @mJarsater I understand, I was under the impression the requirement was to focus on the element, the dropdown opening figured was a required side effect for this component. 

So, `open` should become a prop to be watched to determine if opened or closed?

mJarsater: I just wanna that I am not a part of the Tegel team 😄 I was, but now I'm just an outside enthusiast. 

My suggestion would be to instead of adding all of these methods, which would require users to do some DOM checkups and grab references to the elements we instead make sure that the `open` prop can be uncontrolled/controlled, adds a `defaultOpen` open prop that would have it open by default and then be uncontrolled, and also add a `autofocus` prop that would focus the element on first render. 

To "emulate" native DOM elements we could also add a `focus` method, much like the one you have here.

mJarsater: I think the Tegel team also needs to determine what focus actually means in this case. For a native select element focus does **not** mean open. So either focus should just focus the dropdown "trigger" element, or it should open focus the "trigger" **and** open the dropdown. Same would go for the `open` prop. What would happen if I as a user set it to true? The dropdown is opened but focus is one whatever the focus was on before, the dropdown is opened and focus is on the trigger element, the dropdown is opened and the focus is on the first option? 

https://github.com/user-attachments/assets/c33ba995-c2b4-422f-85fb-210865272615

mJarsater: Another this to consider and try and workout is to trap focus when a user have opened a select/dropdown. Meaning that I as a users should not be able to tab to or press some other element while the dropdown is open. 


https://github.com/user-attachments/assets/70926820-bf99-4271-86bb-b03a5ec712b7


Tbh I think there are a lot of a11y improvements that can be done in terms of focus trapping for dropdown, modal, popover, etc.

mJarsater: The registered keypress might need to be on the dialog element now that i think about it. 

But to trap focus I would try:
1. When the dialog element is rendered, check the DOM for all focusable children and save them in a list
2. Add a listener for tab
3. Prevent the default event for the tab press
4. Set focus on the next focusable element within the dialog element.

mJarsater: Example from a small sideproject:

https://github.com/user-attachments/assets/785c057f-8e06-4f04-900a-8ea56d4a312a

```jsx
import { Component, Element, Host, Listen, State, h } from '@stencil/core';

@Component({
  tag: 'mj-popover',
  styleUrl: 'popover.scss',
  shadow: true,
})
export class MjPopover {
  @Element() host!: HTMLElement;
  private firstFocusable: HTMLElement | null = null;
  private lastFocusable: HTMLElement | null = null;

  private setFocusTrap() {
    const focusableElements = this.host.children;

    if (!focusableElements?.length) return;

    this.firstFocusable = focusableElements[0] as HTMLElement;
    this.lastFocusable = focusableElements[
      focusableElements.length - 1
    ] as HTMLElement;
  }

  private handleKeyDown = (event: KeyboardEvent) => {
    if (event.key === 'Tab') {
      if (event.shiftKey && document.activeElement === this.firstFocusable) {
        // If Shift+Tab, move focus to last element
        event.preventDefault();
        this.lastFocusable?.focus();
      } else if (
        !event.shiftKey &&
        document.activeElement === this.lastFocusable
      ) {
        // If Tab, move focus to first element
        event.preventDefault();
        this.firstFocusable?.focus();
      }
    }
  };

  @Listen('focus', { capture: true })
  handleFocus(event: FocusEvent) {
    if (!this.host.contains(event.target as Node)) {
      this.firstFocusable?.focus();
    }
  }

  render() {
    this.setFocusTrap();
    this.firstFocusable?.focus();
    return (
      <Host part="host" onKeyDown={this.handleKeyDown}>
        <slot />
      </Host>
    );
  }
}

```

mJarsater: sorry for spamming comments, please just take all I've said as suggestions. You know both the components and your users best, do what you see fit with the input 🫶 

@theJohnnyMe

NOMADE55: Really good insight @mJarsater 

I'll wait on  @theJohnnyMe's response to move forward.

theJohnnyMe: Great discussion here and to be honest, it took some time for me to connect as ideas of possible improvements increased exponentially.

What Martin suggests is also on trail of what Stencil proclaims, rather use props and events than methods, also what he mentiones is something we would need to scope on epic level in Jira and prioritize it to improve a lot of components, but probably in smaller chunks in to lower chances for introduction of new bugs. Also, dropdown is delicate component as it combines regular dropdown features, filtering, multiselect and then combination of multiselect and filtering in one. To change something is like playing Jenga, one wrong move and entire thing might collapse. Idea to expose props like defaultOpen, open and autofocus is good one, but we (as Tegel team) would need to take step back and see how internal logic will be affected in all different features this component has.

Back to[ initial problem](https://github.com/scania-digital-design-system/tegel/issues/778), solution provided by @NOMADE55 corresponds to what user needs. Aslo, we have a lot of `@Method`
 in our system, specially on this component so my feeling is that users learned to use it. This specific user explicitly mentions "reset()" which I find interesting as to might be either system expected or experience from before (other system or native behavior). One more reason I like it is that it is simple and does not involve a lot of code changes. It is perfect contribution: fixes the problem, small amount of code changed, easy to test and review.

 I think when we created Tegel, we looked props from perspective of them defining the state of component that is not frequently changed, for anything more methods or relying on custom events would be a way to go. I think that is what users informally learned too.
 
 Doubts that I have for future and I would probably have to ask dev community in Scania:
 _Would you rather use `autofocus` prop to control focus state of component on that component as it does not need reference or you would prefer it to be more similar to native one where you would reference component and use `focus()` method?_  
I would like to hear what average user in Scania think.
 
 Summary:
 I think this PR looks good for previous mentioned reasons. 
Martin's comments about controlled/uncontrolled states and usage of methods is correct too, but we (as Tegel team) would need to take systematic approach to it and dedicate more time to it for two main reason: time to implement it and time to test it. If made through contribution I am afraid there won't be time to properly review it on complex components like this one and then quality might suffer. I will put an effort to present this to team, tech director and PO in January. Great ideas about trapping focus too!

theJohnnyMe (approved):

packages/core/src/components/dropdown/dropdown.tsx
NOMADE55: Adding one more thing to the ongoing conversation, I think Martin's approach of removing methods to rely only on props is sound and valid for more than one component in the library. 

Personally, I'm not a stencil pro nor a Scania dev, but methods in elements is something I'm all for as long as they are well documented 😅.

--

@theJohnnyMe added two more commits, so the approval has been dismissed.

mJarsater: @theJohnnyMe sounds like a great plan! Like mentioned, you know your users and components best. 

Adding one thing on the 'autofocus' prop vs method. I think you try to emulate the native elements here and expose both a prop for autofocus (focus on first render) and a method to let user choose that an element should be focused programmatically. For example focus the next input field when the user has filled out the first one.

Lunkan89 (approved):

Lunkan89:
@NOMADE55 you can rebase develop and then we can merge this one :)

NOMADE55:
I wasn't free until now. I see it already merged, glad to know!

PR #869, opened 2024-11-28
feat: adding motion variables

ckrook:
## **Describe pull-request**  
Adding the motion variables for Tegel


Lunkan89 (approved):
LGTM :)

motion/_motion.scss
mJarsater, line 1: Should these be demoed in storybook in any way?

timrombergjakobsson: Yes box! We will aim on merging the global variables not being tied to any usage in this PR.

mJarsater: Sure, but if these are meant to be distributed in the design system there is a value in demoing then so users could add them to their own components?

timrombergjakobsson: Ah, well, yes maybe something like that is a good idea.

motion/index.scss
```diff
@@ -0,0 +1,2 @@
+@import '_motion';
```
timrombergjakobsson, line 1: I think we can drop the underscore.

```suggestion
@import 'motion';
@import 'animation';
```

theJohnnyMe (approved):
LGTM

timrombergjakobsson (approved):
Looks great! Merge?

PR #876, opened 2024-12-08
feat: enable click on track in slider to update value

max-umain:
## **Describe pull-request**  
Enables ability to click on the track in tds-slider to update its value

## **Issue Linking:**  
- **Jira:** [CDEP-3599](https://tegel.atlassian.net/browse/CDEP-3599) 

## **How to test**  
1. Go to http://localhost:6006/?path=/story/components-slider--default
2. Test clicking anywhere on the track and the value of the slider should update 

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
Not applicable

## **Additional context**  
- Not sure if it's possible or necessary to add unit tests for this feature
- Not sure if documentation should be updated for this
- The height of the track is quite small, so it's not super easy to manage to click on it, unless you're really paying attention. For this reason I think it's appropriate to increase the "clickable height" of the track, but I would like someone else to comment on this first.

### DIFF
--- a/packages/core/src/components/slider/slider.tsx
+++ b/packages/core/src/components/slider/slider.tsx
@@ -186,7 +186,7 @@ export class TdsSlider {
     const numTicks = parseInt(this.ticks);
     const trackRect = this.trackElement.getBoundingClientRect();
     let localLeft = 0;
-    if (event.type === 'mousemove') {
+    if (event.type === 'mousemove' || event.type === 'click') {
       localLeft = event.clientX - trackRect.left;
     } else if (event.type === 'touchmove') {
       localLeft = event.touches[0].clientX - trackRect.left;
@@ -491,6 +491,7 @@ export class TdsSlider {
                 this.trackElement = el as HTMLElement;
               }}
               tabindex={this.disabled ? '-1' : '0'}
+              onClick={(event) => this.thumbCore(event)}
             >
               <div
                 class="tds-slider__track-fill"


max-umain:
In the first commit, I got these errors in sonarcloud:
- Visible, non-interactive elements with click handlers must have at least one keyboard listener.
- Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element.

To address this, I added [this commit](https://github.com/scania-digital-design-system/tegel/pull/876/commits/fa74ad8caa2820ecad78fcc96d1ae13f5bc95a71), 
where I changed this:
```
<div
  class="tds-slider__track"
  ref={(el) => {
    this.trackElement = el as HTMLElement;
  }}
  tabindex={this.disabled ? '-1' : '0'}
  onClick={(event) => this.thumbCore(event)}
>
```
to this:
```
<div
  class="tds-slider__track"
  ref={(el) => {
    this.trackElement = el as HTMLElement;
  }}
  tabindex={this.disabled ? '-1' : '0'}
  onClick={(event) => this.thumbCore(event)}
  onKeyDown={(event) => this.handleKeydown(event)}
  role="slider"
>
```

The thing is that the onKeyDown was already working for this slider, so with this change it now goes two steps instead of one with each key press (because handleKeydown is called twice), which is not what we want.

This change gave these errors in sonarcloud:
- Use <input type="range"> instead of the "slider" role to ensure accessibility across all devices.
- Elements with the ARIA role "slider" must have the following attributes defined: aria-valuenow

And regarding the first of these two errors, it doesn't work if we change it to an input element. One of this element's parent elements is already an input with type="range". Also, if I move the `onClick={(event) => this.thumbCore(event)}` to that input element, the "click on track to update value" functionality stops working.

What should I do here? And is there a way to get these warning messages on my computer without having to commit and push first?

max-umain:
> Use instead of the "slider" role to ensure accessibility across all devices.
I can't find a way to get rid of this error and still have a working solution

max-umain:
Replacing this PR with https://github.com/scania-digital-design-system/tegel/pull/885

PR #898, opened 2024-12-12
fix: tegel core version number

nathalielindqvist:
## **Describe pull-request**  
Due to failed release we need to update version number in core to align on a new release using holy grail script.


timrombergjakobsson (approved):
LGTM

Issue #896, opened 2024-12-12
[Feature Request]: Make it possible to programmatically focus the text field component


patrikhultgren:
### Requirements before reporting

- [X] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [X] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

Make it possible to programmatically focus the text field component.

### Background

In certain situations, we can enhance the user experience by programmatically focusing on text fields. This approach minimizes manual interactions and saves users time.

### Idea

Add a method or similar to the text field component.

### Pull request

https://github.com/scania-digital-design-system/tegel/pull/908

### Contact information

patrik.hultgren@scania.com

ckrook:
Since Patric resolved this on December 16th, I'm closing the issue.

PR #908, opened 2024-12-16
feat(text-field): expose focusElement() method

patrikhultgren:
## **Describe pull-request**  
Exposes focusTextField method for the text field component.

## **Issue Linking:**  
- **GitHub:** https://github.com/scania-digital-design-system/tegel/issues/896

## **How to test**  
Call the focusTextField() method programmatically for a tds-text-field element.

- [x] All existing tests pass


packages/core/src/components/text-field/text-field.tsx
```diff
@@ -130,6 +130,15 @@ export class TdsTextField {
     this.tdsBlur.emit(event);
   }
 
+  /** Method to handle focus */
+  @Method()
+  async focusTextField() {
```
mJarsater, line 135: Maybe we should try and replicate the focus method of an `<input/>` ?

```suggestion
  async focus() {
```

packages/core/src/components/text-field/text-field.tsx
```diff
@@ -130,6 +130,15 @@ export class TdsTextField {
...
+    if (this.textInput) {
+      const event = new FocusEvent('focus');
+      this.textInput.dispatchEvent(event);
```
mJarsater, line 138: To not mix tegel and browser events, maybe this should follow the convention of `tds{event}` -> `tdsFocus`?

packages/core/src/components/text-field/text-field.tsx
patrikhultgren: I’ve already tried that, but I’m getting this warning. What are your thoughts? @mJarsater 

> The @Method() name "focus" is a reserved public name. Please rename the "focus" method so it does not conflict with an existing standardized prototype member. Reusing method names that are already defined on the element's prototype may cause unexpected runtime errors or user-interface issues on various browsers, so it's best to avoid them entirely.

packages/core/src/components/text-field/text-field.tsx
patrikhultgren: Could you show me how to do that while ensuring the feature still works?

packages/core/src/components/text-field/text-field.tsx
mJarsater: Ah! Yeah I see, let's leave it at `focusTextField ` then for now

packages/core/src/components/text-field/text-field.tsx
mJarsater: I would suggest `tdsFocus`, so the users listen for an event called: `tdsFocus`

In react:

```jsx
<TdsTextField onTdsFocus={() => {}}/>
```

in plain js
```javascript
document.querySelector('tds-text-field').addEventListener(('tdsFocus') => {})
```

In angular (not quite sure about how this is done in angular):
```html
<tds-text-field (tdsFocus)={() => {}}/>
```

patrikhultgren: When focusTextField() is called, it triggers the handleFocus(event) method, which then emits the tdsFocus event. Isn't that the intended behavior you were looking for, @mJarsater?

mJarsater: Yeah that's how I would imagine it

patrikhultgren: Great! What adjustments are needed for this to be approved? I noticed I can’t rename the method focusTextField to tdsFocus because there’s already an attribute with that name. Any suggestions, @mJarsater?

patrikhultgren: I’ve tried this, @mJarsater, but it doesn’t work anymore after the change.
```
const event = new CustomEvent('tdsFocus');
this.textInput.dispatchEvent(event);
```

patrikhultgren: That worked! Done.

nathalielindqvist:
@patrikhultgren thank you for this contribution! 🎉  Just to keep you in the loop, the majority of IXIB will be off during the holidays so it might take longer than usual to get this approved. But rest assure that more people will look into this once they are back from vacation :)

packages/core/src/components/text-field/text-field.tsx
```diff
@@ -130,6 +130,14 @@ export class TdsTextField {
     this.tdsBlur.emit(event);
   }
 
+  /** Method to handle focus */
+  @Method()
+  async focusTextField() {
```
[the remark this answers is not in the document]
patrikhultgren: Okay, I've done that change now.

theJohnnyMe (approved):
Great work!

Lunkan89 (approved):

PR #973, opened 2025-01-20
fix(header-dropdown): click list item to close

max-umain:
## **Describe pull-request**  
~~Adds div to list of event sources that close the header dropdown.~~

~~Adds an event emitter to header-dropdown-list-item that is handled in header-dropdown, where it closes the dropdown.~~

Adds "pointer-events: none" for relevant inner elements to not interrupt event to close header dropdown. Thank you to @theJohnnyMe for help with the solution :)

## **Issue Linking:**  
- **Jira:** [CDEP-3774](https://tegel.atlassian.net/browse/CDEP-3774)    

## **How to test**  
1. Go to https://pr-210.d2vh2lmnzgzrkl.amplifyapp.com/
2. Click the avatar in the top right corner to open the header dropdown
3. Click on the settings icon or text, and notice that now it closes the dropdown

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
None

## **Additional context**  
~~I'm not sure if this is a very good solution, adding div as a source of event that closes the dropdown. But the way the dropdown is currently closed is handled like this (when the click is done inside the header-dropdown). I imagine a better solution would be to somehow specify the user clicks anywhere in a dropdown row, it closes it, without specifying that it should be a div.~~

~~The reason I added div specifically is because in the ticket for this PR, the problem was mentioned to be in React, and after investigating I found that the element that is clicked is a next Link (which is compiled to a div).~~

[CDEP-3774]: https://tegel.atlassian.net/browse/CDEP-3774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/components/header/header-dropdown/header-dropdown.tsx
+++ b/packages/core/src/components/header/header-dropdown/header-dropdown.tsx
@@ -45,7 +45,7 @@ export class TdsHeaderDropdown {
 
   handleSlottedItemClick = (event: MouseEvent | KeyboardEvent) => {
     const eventSource = (event.target as HTMLElement).tagName.toLowerCase();
-    if (['a', 'button'].includes(eventSource)) {
+    if (['a', 'button', 'div'].includes(eventSource)) {
       this.open = false;
     }
   };
--- a/packages/core/src/components/header/header-dropdown/header-dropdown.tsx
+++ b/packages/core/src/components/header/header-dropdown/header-dropdown.tsx
@@ -45,7 +45,7 @@ export class TdsHeaderDropdown {
 
   handleSlottedItemClick = (event: MouseEvent | KeyboardEvent) => {
     const eventSource = (event.target as HTMLElement).tagName.toLowerCase();
-    if (['a', 'button', 'div'].includes(eventSource)) {
+    if (['a', 'button'].includes(eventSource)) {
       this.open = false;
     }
   };
--- a/packages/core/src/components/header/header-dropdown/header-dropdown.tsx
+++ b/packages/core/src/components/header/header-dropdown/header-dropdown.tsx
@@ -43,10 +43,12 @@ export class TdsHeaderDropdown {
     this.open = !this.open;
   }
 
-  @Listen('closeDropdownFromListItem')
-  handleCloseDropdownFromListItem() {
-    this.open = false;
-  }
+  handleSlottedItemClick = (event: MouseEvent | KeyboardEvent) => {
+    const eventSource = (event.target as HTMLElement).tagName.toLowerCase();
+    if (['a', 'button', 'div'].includes(eventSource)) {
+      this.open = false;
+    }
+  };
 
   render() {
     return (
@@ -93,7 +95,14 @@ export class TdsHeaderDropdown {
                 },
               ]}
             >
-              {this.open ? <slot></slot> : null}
+              {this.open ? (
+                <span
+                  onClick={(e) => this.handleSlottedItemClick(e)}
+                  onKeyDown={(e) => e.key === 'Enter' && this.handleSlottedItemClick(e)}
+                >
+                  <slot></slot>
+                </span>
+              ) : null}
             </tds-popover-canvas>
           )}
         </div>
--- a/packages/core/src/components/header/header-dropdown/header-dropdown.tsx
+++ b/packages/core/src/components/header/header-dropdown/header-dropdown.tsx
@@ -43,12 +43,10 @@ export class TdsHeaderDropdown {
     this.open = !this.open;
   }
 
-  handleSlottedItemClick = (event: MouseEvent | KeyboardEvent) => {
-    const eventSource = (event.target as HTMLElement).tagName.toLowerCase();
-    if (['a', 'button', 'div'].includes(eventSource)) {
-      this.open = false;
-    }
-  };
+  @Listen('closeDropdownFromListItem')
+  handleCloseDropdownFromListItem() {
+    this.open = false;
+  }
 
   render() {
     return (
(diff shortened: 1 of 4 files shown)


Lunkan89 (commented):
Nice, it works as intended by my testing, just wondering if we need `a` & `button` if we add `div`? since `div` is the wrapping element? what element are is the source when clicking on the icon or text? should we be more specific, can we check if it is an `header-item`?

packages/core/src/components/header/header-dropdown/header-dropdown.tsx
```diff
@@ -45,7 +45,7 @@ export class TdsHeaderDropdown {
 
   handleSlottedItemClick = (event: MouseEvent | KeyboardEvent) => {
     const eventSource = (event.target as HTMLElement).tagName.toLowerCase();
-    if (['a', 'button'].includes(eventSource)) {
+    if (['a', 'button', 'div'].includes(eventSource)) {
```
theJohnnyMe, line 48: I looked very briefly on this one, but I wonder, can we provide a solution where clicking on parent element would provide a better outcome? Ofc, we would need to check which elements and update a logic a bit...

```suggestion
    if (['tds-header-dropdown-list-item'].includes(eventSource)) {
```

max-umain:
> Nice, it works as intended by my testing, just wondering if we need `a` & `button` if we add `div`? since `div` is the wrapping element? what element are is the source when clicking on the icon or text? should we be more specific, can we check if it is an `header-item`?

@Lunkan89 I tested it just now, and having only div causes it to not work when clicking outside of the text or icon

packages/core/src/components/header/header-dropdown/header-dropdown.tsx
max-umain: This change
`if (['tds-header-dropdown-list-item'].includes(eventSource)) {`
by itself doesn't work (I tested it), which I think is what you said. I will investigate if it's possible to fix a solution where clicking any element inside a parent element works

max-umain: The reason it doesn't work now is because the eventSource is one of 'a', 'button', 'div' when we click inside a dropdown row, even if they are child elements of 'tds-header-dropdown-list-item'

max-umain:
@Lunkan89 @theJohnnyMe Thanks for your feedback! I updated the code now, to instead emit an event from header-dropdown-list-item. So now, that a list item (row) is clicked, it closes the dropdown. 

Not sure if we always want clicks on list item to close the dropdown though... What do you think?

nathalielindqvist:
> @Lunkan89 @theJohnnyMe Thanks for your feedback! I updated the code now, to instead emit an event from header-dropdown-list-item. So now, if a list item (row) is clicked, it closes the dropdown.
> 
> Not sure if we always want clicks on list items to close the dropdown though... What do you think?

In the case that a user successfully slots in another dropdown as a list item it might be problematic if the main dropdown closes when a user clicks on the slotted dropdown in an attempt to open it. We might be shooting ourselves in the foot by harshly specifying that all list items closes the dropdown.

max-umain:
> > @Lunkan89 @theJohnnyMe Thanks for your feedback! I updated the code now, to instead emit an event from header-dropdown-list-item. So now, if a list item (row) is clicked, it closes the dropdown.
> > Not sure if we always want clicks on list items to close the dropdown though... What do you think?
> 
> In the case that a user successfully slots in another dropdown as a list item it might be problematic if the main dropdown closes when a user clicks on the slotted dropdown in an attempt to open it. We might be shooting ourselves in the foot by harshly specifying that all list items closes the dropdown.

That's a good point. For that reason perhaps it's better with the solution in the first commit... However, perhaps the element that is clicked to open the inner dropdown would also be a div, if it is then it will close the outer dropdown when trying to open the inner. I'll look into this.

Perhaps another solution would be to add a prop to the header-dropdown-list-item component such as closeOnClick which is true by default, but that could be set to false in case one wants to e.g. open an inner dropdown when clicking on the list item. Could this work?

nathalielindqvist (commented):
I do believe this would work, however I'm wondering about the implementation this would require. From what i can understand the user would have to listen to the newly created event 'closeDropdownFromListItem' on all list items in the header dropdown except the one/few that would potentially open something else, have I understood that correctly? If yes, it would mean a lot of repeating event listeners for a lot of list items. Maybe it would be more desirable to have something the user can opt out of instead, like a "don't close dropdown when this list item is clicked" or similar. Or have I completely misunderstood your implementation? 🙃

max-umain:
@nathalielindqvist 
I'm not sure about the events... I was thinking a prop that, when not passed to the list item, causes the dropdown to close when the list item is clicked. However, if the prop is passed to a list item, it will keep it open. So the default case, when nothing is specified, is to close on click. So my example was a prop called closeOnClick that is true by default, and then the user can set it to false in the few cases it's necessary. But, the opposite would work equally well, such as a prop called keepOpenOnClick that is set to false by default, and can be set to true if necessary :)

theJohnnyMe:
OK, let's go back tothe  initial issue. Problem is that something when user clicks directly on icon or text in the anchor tag, dropdown does not close. I believe reason is that we have code like this:

`<a href="/settings"><tds-icon name="settings" size="16px" class="hydrated"></tds-icon><div class="tds-u-pl1">Settings</div>
</a>`

Instead of 

`<a href="/settings">Settings</a>`

So when you click on an Icon or a text, you are clicking on it, not on the `a` or `button` that handleSlotteItemClick function expects.

In the browser I did a small test by setting `pointer-events: none` as CSS rule on tds-icon and div and then there are no issues. Please see if you can do this small change in CSS of the `header-dropdown-list-item.scss`, I think it should do the trick. 

Another, no so much desired way, would be to try to add `tds-icon` and `div` to the list of eventSources in `handleSlottedItemClick`.

As I see user has not requested a control to toggle off or on a logic for closing of dropdown so lets keep that discussion out.

max-umain:
Demo available here now: https://pr-210.d2vh2lmnzgzrkl.amplifyapp.com/

theJohnnyMe (approved):
Nice one! LGTM

ckrook (approved):

PR #991, opened 2025-01-31
fix: name change webhook file and print node version

mistermalm:
## **Describe pull-request**  
File name was changed. Probably that was the cause of the teams webhook not running.

Also added a job to echo out the node version.


theJohnnyMe (approved):
Nice catch!

ckrook (approved):
Lets try this fix!

PR #994, opened 2025-01-31
fix/smoke test node version

mistermalm:
## **Describe pull-request**  
The node version was only set when triggering manually because we use the inputs property.
Need to set a default node env to be used in scheduled runs.

In this PR i set the default as a fallback if no input was given by a manual run. Therefore in theory, the scheduled run should now use 22.11.0


theJohnnyMe (approved):
Give it a shot :)

PR #995, opened 2025-01-31
fix(tooltip): space claim and text wrapping

theJohnnyMe:
## **Describe pull-request**  
Tooltip tends to claim space in user's application as it has no position defined. We control it by visibility attribute but in some occasions, it tends to occupy space and break the layout. With position absolute issue is removed. 
Text wrapping: additional CSS rule is added to help text wrapping due to limited width.

## **Issue Linking:**  
- **No issue:** Came in support channel.

## **How to test**  
1. Compare the preview link with the prod Storybook.
2. There should be no changes.

- [x] No accessibility violations in storybook
- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

### DIFF
--- a/packages/core/src/components/tooltip/tooltip.scss
+++ b/packages/core/src/components/tooltip/tooltip.scss
@@ -1,6 +1,10 @@
 @import '../../mixins/z-index';
 @import '../../mixins/box-sizing';
 
+:host {
+  position: absolute;
+}
+
 .tds-tooltip {
   @include tds-box-sizing;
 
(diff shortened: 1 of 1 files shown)


packages/core/src/components/tooltip/tooltip.scss
```diff
@@ -1,6 +1,10 @@
 @import '../../mixins/z-index';
 @import '../../mixins/box-sizing';
 
+:host {
+  position: absolute;
```
timrombergjakobsson, line 5: Do we think this can cause any potential side effects?

theJohnnyMe: I do not think so. Underneath is tds-popover-core component that uses `fixed` display attribute and pretty much `@popperjs/core` is the one controlling the position of Tooltip. Use `absolute` here is just to prevent Tooltip from occupying extra space.

Lunkan89 (approved):

PR #1010, opened 2025-02-17
fix(block): nesting coloring issue

theJohnnyMe:
## **Describe pull-request**  
When nesting multiple blocks one within another, it seems that first element changes, while children remain the same. Issue was obvious when switching to secondary mode variant.

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-53`: [CDEP-](https://tegel.atlassian.net/browse/CDEP-53)    

## **How to test**  
1. Open: https://react-demo.tegel.scania.com/
2. Open preview link from draft PR: https://github.com/scania-digital-design-system/tegel-react-demo/pull/213
3. Switch between dark/light mode and its submodes primary/secondart
4. Link for preview contains code from this fix and background colors of block should update in all mode, all submodes and all nested Block elements.

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  


timrombergjakobsson:
Nice one @theJohnnyMe it seems to work now! Great that you set the switches to fixed also!

max-umain (changes requested):
Nice! It seems to work correctly in the demo pages, however, in the amplify preview for this PR (storybook), if you go to Block, it doesn't color correctly when setting mode-variant to secondary. 

I'm guessing this is because in storybook the mode-variant is set directly on the outermost block element rather than a parent element (which I believe is the case in the demo page), and your implementation perhaps doesn't handle this case

theJohnnyMe:
> Nice! It seems to work correctly in the demo pages, however, in the amplify preview for this PR (storybook), if you go to Block, it doesn't color correctly when setting mode-variant to secondary.
> 
> I'm guessing this is because in storybook the mode-variant is set directly on the outermost block element rather than a parent element (which I believe is the case in the demo page), and your implementation perhaps doesn't handle this case

That is a good point, I will make an additional story for nested elements to show how they work ;)

theJohnnyMe:
@max-umain : Two stories are created, one for single and one for multiple nesting levels so user can experience how component works.

Because of how component works, I had to change unit test - remove the part for auto setting mode-variant as it messes up logic for nesting.

max-umain (approved):
Nice!

timrombergjakobsson (approved):
Great!

PR #1020, opened 2025-02-21
chore: npm commands for packing core and wrappers

timrombergjakobsson:
## **Describe pull-request**  
Adding some local npm commands to ease up testing code and features. There is one for each package, and one for all. 

## **Issue Linking:**  
- **Jira:** [CDEP-305](https://jira.scania.com/browse/CDEP-305)    

## **How to test**  
Pull down branch and try out commands locally

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  


package.json
```diff
@@ -15,7 +15,12 @@
     "test:axe": "cd packages/core && npm run test:axe",
     "commit": "cz",
     "prepare": "husky install",
-    "reinstall": "./reinstall.sh"
+    "reinstall": "./reinstall.sh",
+    "pack-core": "cd packages/core && npm run build && npm pack",
```
nathalielindqvist, line 19: Should we try to align the format of the build and pack scripts? The build scripts use colon instead of dash, like "build:all", should these maybe follow the same convention?

package.json
```diff
@@ -15,7 +15,12 @@
...
+    "pack-react": "cd packages/react && npm install $(ls -t ../core/*.tgz | head -1) && npm pack",
+    "pack-angular": "cd packages/angular && npm install $(ls -t ../core/*.tgz | head -1) && npm pack",
+    "pack-angular-17": "cd packages/angular-17 && npm install $(ls -t ../core/*.tgz | head -1) && cd dist/components && npm pack",
```
nathalielindqvist, line 22: It was not obvious where the packed file would end up when packing the angular-17 wrapper, but I guess it has to go under the dist folder? Or is it possible to move where the file ends up so it get highlighted? Not a pressing matter at all, just a convenience thing :)

package.json
timrombergjakobsson: Very good catch!

package.json
timrombergjakobsson: Yes it ends up under dist/components. Same as "npm_pack": "cd dist/components && npm pack"-command that we have already in package.json of angular-17.

mistermalm (approved):

package.json
nathalielindqvist: Since the file is created in the dist folder, it is not highlighted when created like the other files are. So on a glance, it was not obvious where it was. If it's an easy fix to have it be created in the angular-17 folder rather than in its dist folder then it would be nice, so it's highlighted in creation. But again, it's only a convenience thing since these scripts are just for us :)

timrombergjakobsson: Yes that is correct. Well as it most likely wont be included in a commit and merely used for local testing I would say it is not necessary to have the diff. But I'm open to discussion :).

nathalielindqvist: Let's keep it as is I think, it was just one of those random thoughts that made it into a comment 🙃  You can go ahead and resolve this thread!

PR #1037, opened 2025-02-27
chore(textfield): removed linter warning

mistermalm:
removed linter warnings:
`@stencil/core` import should occur before import of `../../utils/hasSlot`
`Arrow function should not return assignment.`


nathalielindqvist (approved):
👍

ckrook (approved):

PR #1051, opened 2025-02-28
feat(badge): accessibility improvements

ckrook:
## **Describe pull-request**  
The pull request addresses the issues we have found in the a11y audit related to the badge component.

✅ Added analyze test file to component
✅ Made sure screen readers are notified when the badge value changes
✅ Added clear context when badge is empty
✅ Make decorative elements not interfere with screen readers
✅ Handling numbers over 99 for screen readers

## **Issue Linking:**  
- **Jira:**  [CDEP-299](https://jira.scania.com/browse/CDEP-299)    

## **How to test**  

**Scenario: Screen readers are notified when the badge value changes**

1. Go to aws link
2. Use a screen reader (VoiceOver for Mac) and load the badge component.
   -  I managed to get this to work with firefox & Voiceover. (Arc browser did not work so there seem to be working with only some browsers. At least we can confirm it does work.)

3. Change the badge value (e.g., from 3 to 4).
4. Verify the screen reader announces the updated value.
5. Check that aria-live="polite" or aria-live="assertive" is set on the badge.

**Scenario: Clear context when badge is empty**

1. Load the badge with no value or an empty value.
2. Confirm the screen reader announces a descriptive label (e.g., “No new notifications”).
3. Check that aria-label is updated to reflect the empty state

**Scenario: Decorative elements do not interfere with screen readers**

1. Add an arbitrary number value below 99. (e.g. 2)
2. Make sure the Screen reader do nor read out the decorative text.
3. The screenreader should _not_ read out "2".
4. Make sure aria-hidden="true" is applied to this decorative text in devtool

**Scenario: Consistent handling of numeric and non-numeric values**

1. Display any arbitrary number over 99 in the badge value control in Storybook
3. Listen with a screen reader to confirm it reads “99 or more notifications.”
4. Verify the badge uses aria-label="Notification badge with 99+ new notifications".

- [x] No accessibility violations in storybook
- [x] All existing tests pass

## **Suggested test steps**
- [x] Keyboard operability
- [x] Interactive elements have labels.
- [ ] Storybook controls
- [ ] Events

### DIFF
--- a/packages/core/src/components/badge/badge.tsx
+++ b/packages/core/src/components/badge/badge.tsx
@@ -16,7 +16,7 @@ export class TdsBadge {
   @Prop() size: 'lg' | 'sm' = 'lg';
 
   /** Defines aria-live attribute */
-  @Prop() ariaLiveAnnounce: 'off' | 'polite' | 'assertive' = 'polite';
+  @Prop() tdsAriaLive: 'off' | 'polite' | 'assertive' = 'polite';
 
   @State() ariaLabelValue: string = '';
 
@@ -54,7 +54,7 @@ export class TdsBadge {
           'tds-badge-hidden': this.hidden,
         }}
         aria-label={this.ariaLabelValue}
-        aria-live={this.ariaLiveAnnounce}
+        aria-live={this.tdsAriaLive}
       >
         <div class="tds-badge-text" aria-hidden="true">
           {this.text}
--- a/packages/core/src/components/badge/badge.tsx
+++ b/packages/core/src/components/badge/badge.tsx
@@ -16,16 +16,13 @@ export class TdsBadge {
   @Prop() size: 'lg' | 'sm' = 'lg';
 
   /** Defines aria-live attribute */
-  @Prop() liveAnnounce: 'off' | 'polite' | 'assertive' = 'polite';
+  @Prop() ariaLevelValue: 'off' | 'polite' | 'assertive' = 'polite';
 
-  /** Defines the template for aria-label, allowing localization */
-  @Prop() ariaLabelTemplate: string = 'Notification badge with {value} new notifications';
+  @State() ariaLabelValue: string = '';
 
   @State() shape: string = '';
 
   @State() text: string = '';
-
-  @State() ariaLabel: string = '';
 
   @Watch('value')
   @Watch('size')
@@ -43,12 +40,6 @@ export class TdsBadge {
       this.shape = this.value.length >= 2 ? 'pill' : '';
       this.size = 'lg';
       this.text = valueAsNumber.toString().length >= 3 ? '99+' : valueAsNumber.toString();
-      this.ariaLabel = this.ariaLabelTemplate.replace('{value}', this.text);
-    } else {
-      this.ariaLabel =
-        this.value.trim() === ''
-          ? this.ariaLabelTemplate.replace('{value}', 'no')
-          : 'The provided value is either empty or a string, please provide a number.';
     }
   }
 
@@ -62,8 +53,8 @@ export class TdsBadge {
           'tds-badge-pill': this.shape === 'pill',
           'tds-badge-hidden': this.hidden,
         }}
-        aria-label={this.ariaLabel}
-        aria-live={this.liveAnnounce}
+        aria-label={this.ariaLabelValue}
+        aria-live={this.ariaLevelValue}
       >
         <div class="tds-badge-text" aria-hidden="true">
           {this.text}
--- a/packages/core/src/components/badge/badge.tsx
+++ b/packages/core/src/components/badge/badge.tsx
@@ -16,7 +16,7 @@ export class TdsBadge {
   @Prop() size: 'lg' | 'sm' = 'lg';
 
   /** Defines aria-live attribute */
-  @Prop() ariaLevelValue: 'off' | 'polite' | 'assertive' = 'polite';
+  @Prop() ariaLiveAnnounce: 'off' | 'polite' | 'assertive' = 'polite';
 
   @State() ariaLabelValue: string = '';
 
@@ -54,7 +54,7 @@ export class TdsBadge {
           'tds-badge-hidden': this.hidden,
         }}
         aria-label={this.ariaLabelValue}
-        aria-live={this.ariaLevelValue}
+        aria-live={this.ariaLiveAnnounce}
       >
         <div class="tds-badge-text" aria-hidden="true">
           {this.text}
--- a/packages/core/src/components/badge/badge.tsx
+++ b/packages/core/src/components/badge/badge.tsx
@@ -55,6 +55,7 @@ export class TdsBadge {
   render() {
     return (
       <host
+        role="status"
         class={{
           'tds-badge': true,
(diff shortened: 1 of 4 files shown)


max-umain (approved):
Great work!

max-umain:
Some questions: 

1. Regarding "Confirm the screen reader announces a descriptive label (e.g., “No new notifications”).", my screen reader app reads this on page reload but gets interrupted in the middle of the sentence and reads something else. Do you experience this too?
2. How can I make my screen reader ignore an element? In "Scenario: Decorative elements do not interfere with screen readers"

ckrook:
> Some questions:
> 
> 1. Regarding "Confirm the screen reader announces a descriptive label (e.g., “No new notifications”).", my screen reader app reads this on page reload but gets interrupted in the middle of the sentence and reads something else. Do you experience this too?
> 2. How can I make my screen reader ignore an element? In "Scenario: Decorative elements do not interfere with screen readers"

1. Which browser are you using? The behavior can vary depending on the browser. In Firefox, for example, setting the value to 0 in the Storybook controls works as expected.

2. If the screen reader is not reading the badge’s text at all, that suggests the browser is ignoring its value. You can verify this by inspecting the badge and checking if aria-hidden is set, which would explain why the text isn’t being announced.

packages/core/src/components/badge/badge.tsx
```diff
@@ -15,47 +15,42 @@ export class TdsBadge {
...
+    let text = '';
+    let ariaLabel = '';
+
     if (!Number.isNaN(valueAsNumber) && this.size !== 'sm') {
-      this.shape = this.value.toString().length >= 2 ? 'pill' : '';
-      this.size = 'lg';
-      this.text = valueAsNumber.toString().length >= 3 ? '99+' : valueAsNumber.toString();
+      shape = this.value.length >= 2 ? 'pill' : '';
+      text = valueAsNumber.toString().length >= 3 ? '99+' : valueAsNumber.toString();
+      ariaLabel = `Notification badge with ${text} new notifications`;
```
mJarsater, line 27: Should this not be set by the users? Might want to have a text in some other language?

ckrook: Good point, @MartinPiq. Since the repository currently only supports English, I wanted to maintain consistency by following the same approach here. Given that the component already has access to the value, I thought it made sense to handle the logic for the aria-label within the component itself.

ckrook: Since you also reacted to Martins comment, please give your thoughts as input @timrombergjakobsson

packages/core/src/components/badge/badge.tsx
```diff
@@ -15,47 +15,42 @@ export class TdsBadge {
...
-          'tds-badge': true,
-          [`tds-badge-${this.size}`]: true,
-          'tds-badge-pill': this.shape === 'pill',
-          'tds-badge-hidden': this.hidden,
-        }}
+        role="alert"
+        class={`tds-badge tds-badge-${this.size} ${shape === 'pill' ? 'tds-badge-pill' : ''} ${
+          this.hidden ? 'tds-badge-hidden' : ''
+        }`}
+        aria-live="assertive"
```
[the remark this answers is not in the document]
mJarsater: Yeah I'd say we should only enable them to set values, but they should control which values are being set to what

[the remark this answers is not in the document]
timrombergjakobsson: A bit late on here, but that aligns with keeping the component flexible. But then if we provide a prop (e.g., ariaLive) that developers can set themselves if needed, we should document the recommended values (polite as a safe default, assertive only for critical updates?

timrombergjakobsson (approved):
Nice work!

packages/core/src/components/badge/badge.tsx
```diff
@@ -15,10 +15,18 @@ export class TdsBadge {
   /** Sets component size. */
   @Prop() size: 'lg' | 'sm' = 'lg';
 
+  /** Defines aria-live attribute */
+  @Prop() liveAnnounce: 'off' | 'polite' | 'assertive' = 'polite';
```
theJohnnyMe, line 19: Since we had a case in Accordion Item where it was not possible to use ariaLevel, I wonder can we follow convention here and have it named ariaLiveValue instead?

theJohnnyMe (changes requested):
Lets see how to move forward with naming a11y props naming convention and internal logic for enabling better UX & DX for translation.

packages/core/src/components/badge/badge.tsx
```diff
@@ -15,10 +15,18 @@ export class TdsBadge {
...
+
+  /** Defines the template for aria-label, allowing localization */
+  @Prop() ariaLabelTemplate: string = 'Notification badge with {value} new notifications';
```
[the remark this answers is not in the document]
mJarsater: My recommendation would probably be to do as little "magic" as possible. The `ariaLabelValue` should directly apply to aria-label, and if the user want to include the amount of notifications in the aria-label they can just include that value in the `ariaLabelValue`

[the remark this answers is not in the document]
ckrook: I agree with Martin, and as discussed with Tim, Johnny, and me, we are moving forward with the decision to place the responsibility on the developer to add the ariaLabelValue.

theJohnnyMe:
I have issues with trying to focus on badge and screen reader. It reads the frame and once I tab again it goes to controls of Storybook, am I missing something?

theJohnnyMe:
Should we be worried about this? 

Screenshot 2025-03-10 at 17 20 07

mJarsater:
> Should we be worried about this?
> 
> Screenshot 2025-03-10 at 17 20 07

Ah yes, we can't prefix aria values with `aria`. I guess you guys would need to agree on a prefix that works.

packages/core/src/components/badge/badge.tsx
```diff
@@ -15,6 +15,11 @@ export class TdsBadge {
   /** Sets component size. */
   @Prop() size: 'lg' | 'sm' = 'lg';
 
+  /** Defines aria-live attribute */
+  @Prop() ariaLiveAnnounce: 'off' | 'polite' | 'assertive' = 'polite';
+
+  @State() ariaLabelValue: string = '';
+
```
theJohnnyMe, line 22:
```suggestion
  /** Defines aria-live attribute */
  @Prop() tdsAriaLive: 'off' | 'polite' | 'assertive' = 'polite';

  @State() ariaLabelValue: string = '';

```

mJarsater: `ariaLabelValue` should probably also be changed. `tdsAriaLabel`

mJarsater: I this prefixing all of these with `tds` works fine. Could even add a page in storybook to document this

theJohnnyMe: ariaLabelValue as it is state could remain as it wont create issue, but for the sake of convention we should probably do that one too.

theJohnnyMe (approved):
Nice work @ckrook !

PR #1055, opened 2025-03-05
feat(breadcrumbs): improve accessibility

max-umain:
## **Describe pull-request**  
Improves accessibility in the Breadcrumbs component.

## **Issue Linking:**  
- **Jira:** [CDEP-302](https://jira.scania.com/browse/CDEP-302)

## **How to test**  
Scenario 1
1. Go to preview link and breadcrumbs page
2. Inspect the tds-breadcrumbs and make sure that the nav tag has aria-label="Breadcrumb"
3. Using a screen reader, tab to the breadcrumbs element and verify that it is correctly announced by the screen reader.

Scenario 2
1. Go to preview link and breadcrumbs page
2. Inspect the tds-breadcrumbs and make sure that it consists of `ol` tags that contain `li` tags.
3. Screen reader should recognize it as a structured list

Scenario 3
1. Go to preview link and breadcrumbs page
2. For the current page "Page 3", inspect and verify that it has the aria-current="page" attribute.
3. The screen reader should announce the current page

- [] I have added unit tests for my changes (if applicable)
- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
None

## **Additional context**  
None


packages/core/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -14,10 +14,10 @@ export class TdsBreadcrumbs {
   render() {
     this.host.children[this.host.children.length - 1]?.classList.add('last');
     return (
-      <nav>
-        <div role="list" class={'tds-breadcrumb'}>
+      <nav aria-label="Breadcrumb">
+        <ol>
           <slot></slot>
-        </div>
+        </ol>
```
mJarsater, line 20: Have you tested this in terms of valid DOM nesting? 

I think you will up with this in the DOM:
```html
<tds-breadcrumbs>
   // shadow DOM
   <ol>
     <tds-breadcumb>
      // shadow DOM
         <li>
            // slotted content
         </li>
       </tds-breadcumb>
   </ol>
</tds-breadcrumbs>
```

Which will mean the the li are not direct descendants of the ol.
Screenshot 2025-03-06 at 14 21 55

packages/core/src/components/breadcrumbs/breadcrumb/breadcrumb.tsx
```diff
@@ -14,9 +14,14 @@ export class TdsBreadcrumb {
 
   render() {
     return (
-      <div role="listitem" class={`${this.current ? 'current' : ''}`}>
-        <slot />
-      </div>
+      <span>
```
ckrook, line 17: Why do we want the <span> here? I think this is causing the DOM nesting issue.

ckrook (changes requested):
I have put the PR in draft mode until we have resolved the DOM nesting issue :)
Besides that, I think this looks good!

packages/core/src/components/breadcrumbs/breadcrumb/breadcrumb.tsx
max-umain: @theJohnnyMe and I looked at it a bit and tried removing it but it didn't help the error

mJarsater: @max-umain if you keep the old structure. Do you still get an error?

mJarsater: I think your'e gonna want to move the `role` attribute to the host. And then add a class for the underlying div to be able to target it with CSS.

max-umain:
@mJarsater @ckrook 
Thank you for your feedback :)

I now made a new commit that overwrote the previous code. In this solution, we stick with using divs with role="list" and role="listitem". The reason for this is @theJohnnyMe and I looked at it a bit and we concluded that there probably isn't a way to get the `<li>` to be a direct child to `<ol>`. 

That means we will still get the warning
> `❌ Incorrect use of role="list" for breadcrumbs.`
> `🧩 To solve this, use semantic <li> element for better accessibility.`

(from the accessibility spreadsheet, not from test results when running npm run test:axe)

One way to get rid of this warning I assume is to remove the role attributes, but then it would possible be even worse in terms of accessibility... I suggest we leave it like this for now and add it to the backlog

mJarsater:
> @mJarsater @ckrook Thank you for your feedback :)
> 
> I now made a new commit that overwrote the previous code. In this solution, we stick with using divs with role="list" and role="listitem". The reason for this is @theJohnnyMe and I looked at it a bit and we concluded that there probably isn't a way to get the `<li>` to be a direct child to `<ol>`.
> 
> That means we will still get the warning
> 
> > `❌ Incorrect use of role="list" for breadcrumbs.`
> > `🧩 To solve this, use semantic <li> element for better accessibility.`
> 
> (from the accessibility spreadsheet, not from test results when running npm run test:axe)
> 
> One way to get rid of this warning I assume is to remove the role attributes, but then it would possible be even worse in terms of accessibility... I suggest we leave it like this for now and add it to the backlog

So just to double check. role="list" and "listitem" would be ok when running text:axe? 

To me it just seems like you need to reshuffle things a bit and move the role="list" to be on the actual `tds-breadcrumbs` and the role="listitem" to be on the `tds-breadcrumb`. Then you would have a DOM nesting which is OK, a list and the lists direct children are listitems.

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -14,8 +14,8 @@ export class TdsBreadcrumbs {
   render() {
     this.host.children[this.host.children.length - 1]?.classList.add('last');
     return (
-      <nav>
-        <div role="list" class={'tds-breadcrumb'}>
+      <nav aria-label="Breadcrumb">
```
mJarsater, line 17: I think this should be set by the users. They could want it in a different language or to add more info.

max-umain: @mJarsater I agree, and I added it to my new commit. @ckrook what do you think?

max-umain:
> So just to double check. role="list" and "listitem" would be ok when running text:axe?

Exactly, axe doesn't complain about them. I think the warning `the breadcrumb list should use <ul> with <li> elements instead of role="list"` came from some other tool @ckrook and @Lunkan89 used.

> To me it just seems like you need to reshuffle things a bit and move the role="list" to be on the actual tds-breadcrumbs and the role="listitem" to be on the tds-breadcrumb. Then you would have a DOM nesting which is OK, a list and the lists direct children are listitems.

That's a good suggestion, but as it won't resolve any if the warnings as we will still keep role="list", I'll leave it as it is

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
ckrook: Sounds good! Lets always be as agnostic as possible.

max-umain:
> I think the warning the breadcrumb list should use <ul> with <li> elements instead of role="list" came from some other tool @ckrook and @Lunkan89 used.

Talked to @ckrook and iirc he got it from guidelines rather than a test result for this specific component

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -11,11 +11,19 @@ import { Component, h, Element } from '@stencil/core';
 export class TdsBreadcrumbs {
   @Element() host: HTMLElement;
```
ckrook, line 13: Add description for documentation here :)

ckrook (changes requested):

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -11,18 +11,20 @@ import { Component, h, Element, Prop } from '@stencil/core';
...
-  @Prop() ariaLabelValue: string;
+  /** The value to be used for the aria-label attribute */
+  @Prop() tdsAriaLabel: string;
 
   render() {
     this.host.children[this.host.children.length - 1]?.classList.add('last');
 
-    if (!this.ariaLabelValue) {
+    if (!this.tdsAriaLabel) {
       console.log(
```
mJarsater, line 21:
```suggestion
      console.warn(
```

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
max-umain: I assume you're talking about the prop below that one. Fixed it now :)

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
max-umain: Thanks. Fixed it now :)

ckrook (approved):

packages/core/src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -11,11 +11,21 @@ import { Component, h, Element } from '@stencil/core';
...
+  @Prop() tdsAriaLabel: string;
+
   render() {
     this.host.children[this.host.children.length - 1]?.classList.add('last');
+
+    if (!this.tdsAriaLabel) {
+      console.warn(
+        'Tegel Breadcrumbs component: please include tdsAriaLabel prop for improved accessibility',
+      );
+    }
```
ckrook, line 24: Move to connectedCallback so we avoid re-render logs :)

theJohnnyMe (approved):
LGTM, nice work @max-umain !

packages/core/src/components/breadcrumbs/breadcrumb/breadcrumb.tsx
max-umain: > I think your'e gonna want to move the `role` attribute to the host. And then add a class for the underlying div to be able to target it with CSS.

@mJarsater is this still relevant?

mJarsater: I think the role="listitem" should be applied to the host element, and not the div inside the shadowDOM

mJarsater:
Styling is broken after the move of listitem to the host element. So the CSS selectors needs updates @max-umain 

Screenshot 2025-03-17 at 13 39 10

max-umain:
@mJarsater right, totally forgot that the css used the role="listitem", thank you :)

PR #1061, opened 2025-03-11
feat(text-field): updated readonly state, added feat to hide icon, improved disabled

mistermalm:
## **Describe pull-request**  

- Updated look and feel of the readonly state styling
- Added feature to toggle readonly icon on/off
- Disabled state is now overriding readonly state to mimic the behavior of native HTML attributes.

## **Issue Linking:**  
- **Jira:** `Text field readonly state`: [CDEP-287](https://jira.scania.com/browse/CDEP-287)

## **How to test**  
1. Go to Text field component in storybook provided by the amplify link below
2. Toggle Read Only on/off, make sure it looks as expected in Figma design [EXEMPEL](https://www.figma.com/design/irjPzqBZm0rDpO4TDl9S2T/Read-only-states-exploration?node-id=0-1&p=f&m=dev)
3. Toggle "Hide Read Only Icon". Make sure it is working as expected
- 3.1. Check that padding is only there when icon is showing.
4. Turn on Disabled when Read Only is active
- 4.1. Check that readonly attribute is no longer on the input field 
- 4.2 Check that user can not select the content
- 4.3 Check that the correct cursor for disabled is active during hover

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Storybook controls


packages/core/src/components/text-field/text-field.tsx
```diff
@@ -94,8 +97,8 @@ export class TdsTextField {
   tdsInput: EventEmitter<InputEvent>;
 
   // Data input event in value prop
-  handleInput(event): void {
-    this.value = event.target.value;
+  handleInput(event: InputEvent): void {
+    this.value = (event.target as HTMLInputElement).value;
```
ckrook, line 101:
```suggestion
    handleInput(event: InputEvent): void {
    this.value = (event.target as HTMLInputElement).value;
        this.tdsInput.emit(event);
  }
```

Perhaps we want to avoid "as" and instead do:

```
  handleInput(event: InputEvent): void {
    const { target } = event;
    if (target instanceof HTMLInputElement) {
      this.value = target.value;
      this.tdsInput.emit(event);
    }
  }
```

ckrook (approved):
I tested the feature, and it works well in Storybook.

It's great that you fixed this! We also noticed the issue with the text field in read-only mode last week, and I see that it has been addressed here. Aside from my small comment, I’m happy to approve 🤘

theJohnnyMe:
@mistermalm check why we lost primary/secondary mode switch:

https://github.com/user-attachments/assets/5c5e13c6-cff1-44cb-bb97-1fe1e81fc6ad

theJohnnyMe (changes requested):
Please check what is off with primary/secondary mode.

mistermalm:
> Please check what is off with primary/secondary mode.
I don't think it is off, I just think Read only mode looks the same as Secondary. When Primary is active in Read Only mode it looks the same as Secondary mode, therefor it looks like its not changing.

What do you think from a design perspective?
@Simon-Philipson

theJohnnyMe:
@Simon-Philipson & @maxadolfsson-scania when text-field component is in read-only state, should component ignore mode-variant setting ?

Please respond here to understand how to proceed with review. :)

Simon-Philipson:
Text-field component aligned with Figma design. Reviewed design in storybook with @mistermalm , approved from design side 💃

theJohnnyMe (approved):
LGTM

PR #1062, opened 2025-03-11
feat(chip): improve accessibility

max-umain:
## **Describe pull-request**  
Improves accessibility in the Chip component.

## **Issue Linking:**  
- **Jira:** [CDEP-308](https://jira.scania.com/browse/CDEP-308)

### Scenario 1:
Not implemented. See additional notes.
 
### Scenario 2:
Not implemented. See additional notes.
 
### Scenario 3:
1. Go to https://pr-1062.d3fazya28914g3.amplifyapp.com/?path=/story/components-chip--default&args=inputType:Radio to test radio type chips
2. Open dev tools
3. Inspect the chips that are checked and verify that their input elements have the attribute aria-checked="true"
4. Inspect the chips that are not checked and verify that their input elements have the attribute aria-checked="false"
5. In the storybook controls, change the chip type to checkbox
6. Repeat steps 3-4

### Scenario 4:
1. Go to https://pr-1062.d3fazya28914g3.amplifyapp.com/?path=/story/components-chip--default to test button type chips
2. Open dev tools
3. Inspect and verify that all visible chips have role="checkbox", role="radio" or role="button" depending on the type you're testing
4. Open screen reader and verify that it announces the chips correctly
5. In the storybook controls, change type to Radio
7. Repeat steps 3-4
8. In the storybook controls, change type to Checkbox
9. Repeat steps 3-4

### Scenario 5:
1. Go to https://pr-1062.d3fazya28914g3.amplifyapp.com/?path=/story/components-chip--default
2. Open dev tools
3. Inspect the chip and verify that its input element has aria-label="Label" (or whatever text you put in Label text in the storybook controls)
4. In the storybook controls, in Label text, delete everything in the input field to make the button have no label
5. Inspect the chip and verify that its input element has aria-label="Chip" (the default when no label is provided)
**See the extra note about Scenario 5 in the additional notes.**

### Scenario 6:
1. Go to https://pr-1062.d3fazya28914g3.amplifyapp.com/?path=/story/components-chip--default&args=inputType:Radio to test radio type chips
2. Open dev tools
3. Inspect and verify that all chips are child elements to a div with role="group" that has an aria-labelledby="group-x" (where x is either 1 or 2). This references the label element with id="group-x".
4. In the storybook controls, change the chip type to Checkbox
5. Repeat step 3

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
None

### Regarding Scenario 1:
This has not been fixed.

From Jira ticket:
> Scenario: Ensure type radio chips are navigable via keyboard
> Given multiple radio chips are present in a group
> When the user navigates using the Tab key
> Then it should be possible to focus on unchecked chips
> But currently, only the checked chip is focusable

I'm not sure if this is something we should fix, simply it currently works exactly like it works in native html checkboxes and radio buttons. So, I have not fixed this. **Should we?**

### Regarding Scenario 2:
This has not been fixed.

From Jira ticket:
> Scenario: Ensure type checkbox chips have a visible focus ring
> Given a checkbox chip component
> When the chip is toggled
> Then the focus ring should be clearly visible
> But currently, the focus ring is not obvious
> (Sync with designer how the focus ring should look like)

I talked to Max the designer about this and we currently think that the best solution is to modify _focus-state.scss, which will affect all components. Therefore, I think it's best to fix it in a separate PR. **Maybe we should create a separate ticket for that?**

### Note about Scenario 5:
From Jira ticket:
> And screen readers should have a reliable text label for the chip

I believe this is what we accomplish with aria-label. The screen reader I use reads the text in the aria-label attribute. **Is this the desired behavior?**

### DIFF
--- a/packages/core/src/components/chip/chip.stories.tsx
+++ b/packages/core/src/components/chip/chip.stories.tsx
@@ -137,56 +137,62 @@ const Template = ({ inputType, size, label, icon, iconPosition, disabled }) => {
         display: flex;
         gap: 8px;
       }
+      label {
+        margin: 8px;
+      }
     </style>
 
+    <label id="group-1">Group 1</label>
     <div class="demo-wrapper">
-      <tds-chip type="checkbox" size="${
-        sizeLookUp[size]
-      }" checked ${disabledAttribute} value="checkbox-1-value">        
+      <div role="group" aria-labelledby="group-1">
+        <tds-chip type="checkbox" size="${
+          sizeLookUp[size]
+        }" checked ${disabledAttribute} value="checkbox-1-value">        
+            ${
+              icon && iconPosition === 'Prefix'
+                ? '<tds-icon slot="prefix" name="heart" size="16px"></tds-icon>'
+                : ''
+            }
+            <span slot="label">
+              ${label} 1
+            </span>
+            ${
+              icon && iconPosition === 'Suffix'
+                ? '<tds-icon slot="suffix" name="cross" size="16px"></tds-icon>'
+                : ''
+            }        
+        </tds-chip>
+        <tds-chip type="checkbox" size="${sizeLookUp[size]}" value="checkbox-2-value">
           ${
             icon && iconPosition === 'Prefix'
-              ? '<tds-icon slot="prefix" name="heart" size="16px"></tds-icon>'
+              ? '<tds-icon slot="prefix" name="email" size="16px"></tds-icon>'
               : ''
           }
-          <span slot="label">
-            ${label} 1
-          </span>
+            <span slot="label">
+              ${label} 2
+            </span>
+            ${
+              icon && iconPosition === 'Suffix'
+                ? '<tds-icon slot="suffix" name="cross" size="16px"></tds-icon>'
+                : ''
+            } 
+        </tds-chip>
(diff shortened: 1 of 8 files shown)


ckrook (approved):
Regarding Scenario 1. Good catch. You are correct. 

Regarding Scenario 2. Great that you talked with Max regarding that. Then we can wait with this and fix it in a separate PR later down the road.

Scenario 5. Yes. It works now as I would have excepted. Nice!

---
Overall great how to steps you provided to reproduce the tests. I really found it easy to test all the scenarios The only comment I have was the amount of code changes that appeared in the files. Perhaps your prettier is misaligned? Lets sync this with Jonathan. Becomes quite hard to review the code.

But other than that, great work! 🤘

theJohnnyMe (approved):
Scenarios 3,4,5,6 tested and work OK.

**Regarding Scenario 1**: 

> Keyboard navigation works so I do not think any additional action in needed here.


**Regarding Scenario 2:** 

> Focus ring is there but not so visible. Make a separate Jira ticket to either alter focus ring of Chip so it is more visible, or background color of Chip component should alter a bit when focused - check how primary button changes on focus, maybe that could be inspiration for it.

**Regarding Scenario 5:** 

> I think aria-lable does the job here. 

Nice work @max-umain !

PR #1074, opened 2025-03-17
feat(link): improve accessibility

Lunkan89:
## **Describe pull-request**  
Added accessibility test
preventing enter keydown when disabled
preventing focus when disabled

## **Issue Linking:**  
- **Jira:** [CDEP-417](https://jira.scania.com/browse/CDEP-417)

## **How to test**  
1. Go to amplify link
2. Navigate to Link component
3. Disable link and try to tab -> Enter on link

- [x] All existing tests pass

### DIFF
--- a/packages/core/src/components/link/link.tsx
+++ b/packages/core/src/components/link/link.tsx
@@ -22,10 +22,10 @@ export class TdsLink {
 
   connectedCallback() {
     const links = this.host.querySelectorAll('a');
-    if(links.length > 1) {
-      console.warn('tds-link is only intended to wrapp one <a> tag')
+    if (links.length > 1) {
+      console.warn('tds-link is only intended to wrap one <a> tag');
     }
-    const link = links[0]
+    const link = links[0];
     if (link) {
       if (this.disabled) {
         link.setAttribute('tabindex', '-1');
--- a/packages/core/src/components/link/link.tsx
+++ b/packages/core/src/components/link/link.tsx
@@ -21,7 +21,20 @@ export class TdsLink {
   @Prop() standalone: boolean = false;
 
   connectedCallback() {
-    this.host.children[0].classList.add('tds-link-component');
+    const links = this.host.querySelectorAll('a');
+    if(links.length > 1) {
+      console.warn('tds-link is only intended to wrapp one <a> tag')
+    }
+    const link = links[0]
+    if (link) {
+      if (this.disabled) {
+        link.setAttribute('tabindex', '-1');
+        link.setAttribute('aria-disabled', 'true');
+      } else {
+        link.removeAttribute('tabindex');
+        link.removeAttribute('aria-disabled');
+      }
+    }
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


packages/core/src/components/link/link.tsx
```diff
@@ -20,6 +20,24 @@ export class TdsLink {
   /** Displays the Link as a standalone component. Not part of a paragraph. */
   @Prop() standalone: boolean = false;
 
+  /** Listens to keydown to prevent key event when diabled */
+  @Listen('keydown', { target: 'window' })
+  handleKeydown(event) {
```
Lunkan89, line 25: One can argue that there will be no keydown when you cannot focus, maybe remove this one altogether.

Lunkan89:
> @Lunkan89 I'm curious about your approach to handling the disabled state in tds-link. Wouldn't it make more sense to modify the element directly, since it's the actual link?
> 
> For example, with your implementation, I noticed that aria-disabled="true" isn't applied, which is important for accessibility. I think I would just add this logic in the connectedCallback which would handle the link disabled scenario.
> 
> ```
> connectedCallback() {
>   const link = this.host.querySelector('a');
>   if (link) {
>     link.classList.add('tds-link-component');
>     if (this.disabled) {
>       link.setAttribute('tabindex', '-1');
>       link.setAttribute('aria-disabled', 'true');
>     } else {
>       link.removeAttribute('tabindex');
>       link.removeAttribute('aria-disabled');
>     }
>   }
> }
> ```
> 
> Want to hear your thoughts on this!

@ckrook I think your on to something of a better implementation, my idea was to remove focus for all links within component. Probably just me over thinking the implementation.
why add the class thought?
what happens if multiple elements are put in?

ckrook:
> > @Lunkan89 I'm curious about your approach to handling the disabled state in tds-link. Wouldn't it make more sense to modify the element directly, since it's the actual link?
> > For example, with your implementation, I noticed that aria-disabled="true" isn't applied, which is important for accessibility. I think I would just add this logic in the connectedCallback which would handle the link disabled scenario.
> > ```
> > connectedCallback() {
> >   const link = this.host.querySelector('a');
> >   if (link) {
> >     link.classList.add('tds-link-component');
> >     if (this.disabled) {
> >       link.setAttribute('tabindex', '-1');
> >       link.setAttribute('aria-disabled', 'true');
> >     } else {
> >       link.removeAttribute('tabindex');
> >       link.removeAttribute('aria-disabled');
> >     }
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Want to hear your thoughts on this!
> 
> @ckrook I think your on to something of a better implementation, my idea was to remove focus for all links within component. Probably just me over thinking the implementation. why add the class though? what happens if multiple elements are put in?

I added the class because I thought I had seen it before, but you can ignore it.

I think you're right—we need to handle the scenario where multiple elements are present. We should account for that by selecting all <a> elements and iterating through them

max-umain (approved):

ckrook (approved):

PR #1075, opened 2025-03-17
feat(icons): added ky icons

mistermalm:
## **Describe pull-request**  
Added KY icons and re-generated files with gulp
Icons were not treated in SVGOMG

## **Issue Linking:**  
- **Jira:** `Add KY Icons To Repo`: [CDEP-460](https://jira.scania.com/browse/CDEP-460)

## **How to test**  
1. Go to storybook
2. Check Icons gallery


maxadolfsson-scania:
Two things:

1. We are not sure how to work with the SVGOMG tool for almost 100 icons at the same time. We will most likely wait for Simon to handle and suggest a way forward for this, to ensure a more solid handover process.

2. In the gallery I can see that some of them still need some more attention, hopefully Simon sits on the correct exported files already but still wanted to note these details here:
  - `cookie_inactive` (most of the cookie seems to be missing, only the little dot is visible in Storybook)
  - `drop` (should not be filled)

Screenshot from Figma:

> image

Let me know if this makes sense or not!

ckrook (changes requested):
I think these icons need to be optimized in order for my approval. What are the reason they are not?

Maybe this is a great opportunity to create a Github action with [SVGOG](https://www.npmjs.com/package/svgo). I think this would save everyone time.

icons/src/svg/dollar–sign.svg
maxadolfsson-scania, line 1: I think this is a duplicate, due to different name in Figma. We already have `dollar.svg` and should keep it. This file can be removed.

icons/src/svg/skip_backward.svg
maxadolfsson-scania, line 1: I think this is a duplicate, due to different name in Figma. We already have `skip_backards.svg` (with `s` at the end) and should keep it. This file can be removed.

maxadolfsson-scania:
@ckrook - Makes sense! Like I said we are waiting for Simon provide the full set of updated, optimised icons. Depending on how you prefer to do it, I would change this to a draft PR meanwhile.

ckrook (approved):

PR #1079, opened 2025-03-19
feat(dropdown): accessibility improvements

Lunkan89:
## **Describe pull-request**  
* Added accessibility test for basic dropdown
* Added accessibility test for filter dropdown
* Moved current `aria`-attributes to interactive elements
* Added `aria-label`, `aria-disabled`, `aria-orientation`, `aria-multiselectable`
* Fixed failing test (Removed some `aria-selected` expects in e2e test since we have axe tests now)

**NOTE**
Not tested `multiselect` yet since that uses `tds-checkbox`( which is not ready yet) within `tds-dopdown-option`

## **Issue Linking:**  
- **Jira:** [CDEP-432](https://jira.scania.com/browse/CDEP-)

## **How to test**  
1. Go to Amplify.
2. See that `aria`-attributes behave as expected.
3. Try out a wider wrapper to confirm responsiveness.

- [x] All existing tests pass

### DIFF
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -698,7 +698,7 @@ export class TdsDropdown {
         <div
           role="listbox"
           aria-label={this.tdsAriaLabel}
-          aria-orientation="horizontal"
+          aria-orientation="vertical"
           aria-multiselectable={this.multiselect}
           ref={(element) => {
             this.dropdownList = element;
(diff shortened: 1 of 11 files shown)


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -533,13 +536,15 @@ export class TdsDropdown {
     if (form) {
       form.removeEventListener('reset', this.resetInput);
     }
+    if (!this.tdsAriaLabel) {
+      console.warn('Tegel Dropdown component: tdsAriaLabel prop is missing');
+    }
```
max-umain, line 541: Not sure if this works, in the other components we put this warning in connectedCallback? Also, can you add aria label to the storybook controls so it can be tested?

max-umain (changes requested):
See the comment I wrote about disconnectedCallback and storybook controls

packages/core/src/components/dropdown/dropdown.tsx
Lunkan89: Yes it should be in `connectedCallback`, as I thought is was, now I see `disconnected`.
Storybook controls coming up 👍

ckrook (approved):

max-umain (approved):

PR #1096, opened 2025-03-27
fix(text-field): accessibility improvements

max-umain:
## **Describe pull-request**  
Accessibility improvements for the text-field component.

## **Issue Linking:** 
- **Jira:** [CDEP-378](https://jira.scania.com/browse/CDEP-378)

#### Scenario 1
Scenario: TextField adapts to different screen sizes 

> Given the component is rendered in Storybook 
> When viewport changes width 
> Then the component should adjust its size accordingly without layout breaks 
> And it should maintain a clear and consistent UI

This already works. To better demonstrate it working in Storybook, I updated the Storybook page so that if you set the noMinWidth prop to true, then the TextField component will adjust to very narrow screen sizes.

#### Scenario 2

> Scenario: TextField focus indication 
> Given the TextField is rendered 
> When the user focuses on the TextField using Tab key 
> Then a clear visual indication of focus should be present (focusring)

This one is not fixed yet. I have a draft of the code for achieving the (most likely) design in fix/a11y-text-field-testing, the problem is just that the design hasn’t been fully decided on yet. Will address this in a separate ticket.

#### Scenario 3: 

> Scenario: Readonly field keyboard navigation issue
> Given the TextField is in a readonly state
> When the user navigates to it using the keyboard
> Then no placeholder text should appear

I talked with @ckrook about this and we decided this is not something we should implement.

#### Scenario 4

> Scenario: Label Association 
> Given a screen reader is active 
> When the text-field is rendered with a label 
> Then it should properly announce the label associated with the field

Fixed this by using the label as the value for aria-label. However, if the prop tdsAriaLabel is specified, that one will be used as the value for aria-label instead. 

How to test:
1. Go to https://pr-1096.d3fazya28914g3.amplifyapp.com/?path=/story/components-text-field--default&args=label:Now+the+label+is+visible;labelPosition:Outside;tdsAriaLabel
2. Inspect the text-field and verify that the aria-label is the value in Label, in the storybook controls
3. In the storybook controls, to the Aria Label control, add any text
4. Inspect the text-field again and verify that it now has the aria-label with the same value as in Aria Label in the storybook controls

#### Scenario 5

> Scenario: Error State Announcements 
> Given a screen reader is active 
> When the TextField is in an error state 
> Then it should announce the error message clearly with aria-live

Fixed with aria-live="assertive” on the parent element of the element containing the helper text.

How to test:
1. Go to https://pr-225.d2vh2lmnzgzrkl.amplifyapp.com/about
2. Activate screen reader
3. Input more than 5 characters in to the text field
4. Notice it interrupts and reads the error message

#### Scenario 6

> Scenario: Ensure helper text is programmatically connected 
> Given the TextField has helper text 
> When the screen reader reads the input field 
> Then the helper text should be programmatically connected via aria-describedby

I fixed so that helper text is pointed to by aria-describedby in the input element. *However, when there is text in the aria-describedby element (which in this case is the helper text), the placeholder of the Text Field is not ignored by screen reader (at least voiceover on mac)* So, my question is: *do we want the placeholder to be read too? If yes I can add it to the aria-describedby for the input element*

How to test:
1. Go to https://pr-1096.d3fazya28914g3.amplifyapp.com/?path=/story/components-text-field--default&args=readonly:true;tdsAriaLabel:This+is+the+aria+label
2. Activate screen reader
3. Reload the page
4. It should read the text "This is the aria label"

#### Scenario 7

> Scenario: Readonly State Description 
> Given a TextField is in readonly state 
> When a screen reader navigates to it 
> Then it should announce that the field is non-editable

I added the tdsAriaLabel prop, so that it can be used for informing about the fact that the text field is read only. So they need to keep aria-label (through tdsAriaLabel) updated based on whether this.readOnly is true.

How to test:
1. Go to https://pr-1096.d3fazya28914g3.amplifyapp.com/?path=/story/components-text-field--default&args=readonly:true;tdsAriaLabel:This+is+the+aria+label+of+a+text-field+that+is+not+editable+as+it+is+read-only
2. Activate screen reader
3. Focus on the text field and it should read the message "This is the aria label of a text-field that is not editable as it is read-only"

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
None

## **Additional context**  
None


packages/core/src/components/text-field/text-field.stories.tsx
```diff
@@ -246,7 +255,7 @@ const Template = ({
     <style>
     /* demo-wrapper is for demonstration purposes only*/
   .demo-wrapper {
-    width: 200px;
+    ${noMinWidth ? '' : 'width: 200px;'}
```
ckrook, line 258: Why do we not just set max-width: 200px here?

ckrook (approved):

Lunkan89 (approved):

PR #1113, opened 2025-04-01
feat(spinner): multibrand token usage

theJohnnyMe:
## **Describe pull-request**  
Added multibrand tokens to spinner component
This component is relaying on using direct values instead of tokens, you will see that by observing there are no variables connected, but rgba values for different brands. This is one of the rare components that has it like that.

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-`: [CDEP-540](https://jira.scania.com/browse/CDEP-540)

## **How to test**  
1. Go to the preview link
2. Open production Storybook
3. Compare the colors of spinner - there should be no difference.
4. Test switching to traton brand by adding class "traton" in body tag when you inspect Storybook code. 

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Additional context**  
Brand switch control will be added as a global control in Storybook in one of the upcoming PRs, for now, inject traton class manually to test it.


packages/core/src/components/spinner/spinner.scss
```diff
@@ -1,70 +1,110 @@
...
+
+.traton,
+.traton .tds-mode-light {
+  --spinner-background: rgba(0, 72, 82, 1);
+}
+
+.traton .tds-mode-dark {
+  --spinner-background: rgba(194, 191, 182, 1);
+}
+
```
mJarsater, line 26: I was just thinking, instead of you shipping both styles to the browser. Have you looked at [style modes](https://stenciljs.com/docs/styling#style-modes)?

timrombergjakobsson: That is maybe a good idea! Maybe it would work nice together with the token structure and Style Dictionary?
So then we can do this then?

```
@Component({
  tag: 'tds-spinner',
  styleUrls: {
    'scania-light': 'spinner.scania-light.css',
    'scania-dark': 'spinner.scania-dark.css',
  },
  shadow: true,
})
```
and then set it like this?

`<tds-spinner mode="scania-light"></tds-spinner>`

mJarsater: Yeah that one part of it 👍🏻 But also the different brand. So when the import tegel they could do:
```jsx
import { setMode } from '@stencil/core';
setMode(() => 'traton'); 
```

timrombergjakobsson: Ah yeah, then it is more on a global level in comparison to a more "local" mode(per component).
But maybe both scenarios fills a purpose.

theJohnnyMe: Thank you for sharing idea, however this would not be a use case for Tegel. We are now working on establishing theming system that would really on new token structure. Until now, we had a system where we had a component level tokens, now we are moving to multibrand approach and usage tokens. Usage tokens are like API for generation of new brand, we would ask a brand to fill up values for eg. 300 tokens and there will be another brand showing up.

Right now, we have a lot of tokens defined with :root, however with new implementation only scania will remain with :root while additionally it will have a class .scania. Trtaon will have its tokens limited .traton class. Those CSS variables can shift instantly in browser and has proved to work really good so far in our dark/light mode. 

So what does that mean for a component level CSS: It will have definitions where color, typography and sizing are represented with design system usage variables which (those 300 cca)every brand will define and that is pretty much the core of brand. Component, no matter web component or pure CSS one can easily consume them. 

https://stenciljs.com/docs/styling#style-modes  - Style modes might be of use in case we do not have proper token structure in place or where two CSS files that style component are radically different between brands. It would be of use where we have system only built with web-components, however, there is a proposal of creation of tegel-light, a CSS only version of Tegel, so new tokens can be used there with no problems.

One might say, but why then spinner has these brand classes in side and what it is not relying on globally defined values? Designers, even they worked hard to standardize entire system to rely on new tokens found a 3-4 component that due to design needs could not follow token structure so they will have local overrides. Even thou it is not desired, it is still much better than every component having its own variables - that would be nightmare to align between brands.

Those are the reasons why Style Modes would not work for us. We are going towards idea of creation of "white label" design system where those 300 variables in one file is themeing file for both webcomponents and upcoming CSS only components.

mJarsater: Cool beans! 🚀  I was just a bit concerned when I saw duplicate styles be shipped to the browser. 

Nice stuff 🙌🏻

mistermalm (approved):

max-umain (approved):

PR #1115, opened 2025-04-02
feat(modal): accessibility improvements

Lunkan89:
## **Describe pull-request**  
Improved accessibility for modal component.

## **Issue Linking:**  
- **Jira:** [CDEP-419](https://jira.scania.com/browse/CDEP-419)

## **How to test**  
**Feature: Ensure component responsiveness**
Given the component is rendered in Storybook 
When viewport changes width 
Then the component should adjust its size accordingly without layout breaks 
And it should maintain a clear and consistent UI

_Seems to be working already, please resize window and look at container and text behavior._

### Feature: Keyboard interactability

**Scenario: Moving Focus to the First Tabbable Element Inside the Dialog**
Given a dialog is open and the focus is on the last tabbable element inside the dialog
 When the user presses Tab to navigate further
 Then the focus should move to the first tabbable element inside the dialog
 And this ensures a circular focus loop within the dialog
 But if the dialog has no tabbable elements, focus should not shift unexpectedly

_This is achieved by trapping the tab-keydown-event and taking control over it, tab-focus a interactable element inside the modal you should be in a circular tab behavior, using `tab` and `shift-tab`._

**Scenario: Closing the Modal with the Escape Key**
Given a modal or dialog is open
 When the user presses the Escape key
 Then the modal should close
 And focus should return to the element that triggered the modal
 But if the modal is not closable, the Escape key should have no effect

_Added listener for escape on keydown_

**Scenario: Ensuring Dialog Is Not Closable When closable Is False**
Given the closable property of the dialog is set to false
 When the user tries to close the dialog
 Then it should not be possible to close the dialog
 And the dialog should remain open regardless of user interaction
 But if closable is true, the dialog should be closable with the Escape key or a close button

_Listener for escape should only trigger `handleClose()` when `prevent` prop is false._

**Scenario: Returning Focus to the Invoking Element After Closing the Dialog**
Given a dialog is closed
 When the modal is closed
 Then the focus should return to the element that invoked the dialog
 And this ensures continuity of the user's navigation
 But if the dialog was triggered by a non-focusable element, focus may return to the document body or another appropriate location

_Finding referenced element or passed selector for reference element, if that element is a custom element we dig deeper for interactive element, test by closing modal with close button or escape and focus should return to tds button and open on direct enter click_

### Feature: Component compliant with screen reader behavior

**Scenario: Adding role="dialog" to tds-modal**
Given a modal component tds-modal is displayed
 When the modal is rendered
 Then the tds-modal element should have the role="dialog" attribute
 And this helps screen readers identify it as a dialog
 But if the modal serves as an alert or other specific function, a more appropriate role, like role="alertdialog", should be used instead

_Added prop for this, inspect to see that role is applied_

**Scenario: Adding aria-modal="true" to tds-modal**
Given the modal component tds-modal is displayed
 When the modal is opened
 Then the tds-modal element should have the aria-modal="true" attribute
 And this will inform assistive technologies that the modal is modal and blocks interaction with other content
 But if the dialog is not truly modal (e.g., it allows interaction with background content), this attribute should not be added

_Added attribute, inspect to confirm that attribute is there_

**Scenario: Adding aria-labelledby for the Header in tds-modal**
Given a tds-modal with a header section
 When the modal is rendered
 Then the modal should have the aria-labelledby attribute
 And the attribute should reference the ID of the header element to provide a clear label for the dialog
 But if the modal lacks a header or title, the aria-labelledby attribute should be omitted

_Added dynamic id for header, inspect that header id is the correct id used in `aria-labelledby`_

**Scenario: Adding aria-describedby for the Body in tds-modal**
Given a tds-modal with a body section containing content
 When the modal is rendered
 Then the modal should have the aria-describedby attribute
 And the attribute should reference the ID of the body content to describe the purpose or details of the modal
 But if the body contains no essential information or is empty, the aria-describedby attribute may be omitted

_Added dynamic id for body, inspect that body-slot-wrapper id is the correct id used in `aria-describedby`_

- [x] All existing tests pass

### DIFF
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -276,7 +276,7 @@ export class TdsModal {
     return (
       <Host
         role={this.tdsAlertDialog}
-        aria-modal={this.isShown}
+        aria-modal="true"
         aria-describedby={bodyId}
         aria-labelledby={headerId}
         class={{
@@ -287,7 +287,7 @@ export class TdsModal {
       >
         <div class="tds-modal-backdrop" />
         <div class={`tds-modal tds-modal__actions-${this.actionsPosition} tds-modal-${this.size}`}>
-          <div class="header">
+          <div id={headerId} class="header">
             {this.header && <div class="header-text">{this.header}</div>}
             {usesHeaderSlot && <slot name="header" />}
 
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -57,10 +57,22 @@ export class TdsModal {
   // State that keeps track of show/closed state for the Modal.
   @State() isShown: boolean = false;
 
+  // Focus state index in focusable Elements
+  @State() activeElementIndex = 0;
+
   /** Shows the Modal.  */
   @Method()
   async showModal() {
     this.isShown = true;
+
+    // Set focus on first element when opened
+    requestAnimationFrame(() => {
+      const focusableElements = this.getFocusableElements();
+      if (focusableElements.length > 0) {
+        focusableElements[0].focus();
+        this.activeElementIndex = 0;
+      }
+    });
   }
 
   /** Closes the Modal. */
@@ -81,7 +93,7 @@ export class TdsModal {
   @Listen('keydown', { target: 'window' })
   handleKeyDown(event: KeyboardEvent) {
     if (event.key === 'Escape' && this.isShown && !this.prevent) {
-      this.handleClose(event)
+      this.handleClose(event);
     }
   }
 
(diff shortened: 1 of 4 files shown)


Lunkan89:
needs improvements!

packages/core/src/components/modal/modal.tsx
```diff
@@ -179,8 +270,15 @@ export class TdsModal {
     const usesHeaderSlot = hasSlot('header', this.host);
     const usesActionsSlot = hasSlot('actions', this.host);
 
+    const headerId = this.header ? `tds-modal-header-${generateUniqueId()}` : undefined;
+    const bodyId = `tds-modal-body-${generateUniqueId()}`;
+
     return (
       <Host
+        role={this.tdsAlertDialog}
+        aria-modal={this.isShown}
```
ckrook, line 279: hardcode this to true since its always is true

ckrook (approved):

max-umain (approved):

PR #1132, opened 2025-04-09
feat(stepper): improved accessibility

Lunkan89:
## **Describe pull-request**  
Improved accessibility for stepper component

## **Issue Linking:**  
- **Jira:** [CDEP-426](https://jira.scania.com/browse/CDEP-426)

### Feature: Component compliant with screen reader behavior

**Scenario: Indicate Current Step Dynamically**
 Given the component is rendered in Storybook
 When the user navigates through the steps
 Then the aria-current="step" attribute should be dynamically set on the active step
 But currently, this attribute is not updated correctly, affecting accessibility

_Added prop for user to add current step dynamically on implementation/usage_

**Scenario: Define Proper Roles for Stepper and Steps**
 Given the component is rendered in Storybook
 When the stepper is displayed
 Then each step should have role="listitem"
 And the stepper container should have role="list"
 But currently, these roles may be missing, impacting screen reader interpretation

_This was already in place, please confirm by inspecting stepper component_

**Scenario: Maintain Accessibility on Orientation Change**
 Given the component is rendered in Storybook
 When the orientation of the stepper changes
 Then all visual and structural changes should be applied dynamically
 And the step relationships should be revalidated with screen readers
 But currently, orientation changes may not properly update accessibility attributes

_**Up for discussion**: My initial thought is that this is not needed_

- [x] All existing tests pass


ckrook:
Added some fixes so the stepper works with screenreader.

ckrook (approved):

PR #1137, opened 2025-04-11
feat(datetime): accessibility improvements

max-umain:
## **Describe pull-request**  
Improves accessibility for the datetime component.

## **Issue Linking:**  
- **Jira:** [CDEP-411](https://jira.scania.com/browse/CDEP-411)

#### Scenario 1:

> Given the component is rendered in Storybook
> When viewport changes width
> Then the component should adjust its size accordingly without layout breaks
> And it should maintain a clear and consistent UI 

How to test:
1. Go to https://pr-1137.d3fazya28914g3.amplifyapp.com/?path=/story/components-datetime--default
2. Change the width of the window and verify that the datetime component adjusts its width.

#### Scenario 2: 

> Scenario: Opening the Datepicker with Enter Key
> Given a user is focused on the datepicker input field
> When the user presses the Enter key
> Then the datepicker should open
> And the user should be able to navigate and select a date using the keyboard
> But if the datepicker is already open, pressing Enter should not close it unexpectedly. 

How to test:
1. Go to https://pr-1137.d3fazya28914g3.amplifyapp.com/?path=/story/components-datetime--default
2. Tab e.g. the year "placeholder" inside the datetime component
3. Press the enter key and verify that it opens the datepicker
4. Try navigating around inside the datepicker with the arrow keys and tab
5. Press enter to select a date

Note that this part:

> But if the datepicker is already open, pressing Enter should not close it unexpectedly. 

has not been fixed, see Additional context.

#### Scenario 3:

This has not been fixed, see Additional context.

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
None

### Regarding scenario 2:

As mentioned, this part:

> But if the datepicker is already open, pressing Enter should not close it unexpectedly. 

**has not been fixed**. This is because this is how it works with the native date input datepicker in html, when the user presses enter, a date is selected and the datepicker is closed. **Therefore I suggest we keep it like this.**

### Regarding scenario 3:

> Scenario: Proper Screen Reader Announcement for Unset Datepicker Values
> Given a datetime picker with no date or time selected
> When a screen reader reads the year, month, or day fields
> Then it should not announce a percentage value
> And it should instead announce "unset" or a similar appropriate message
> But if a date or time is selected, it should read out the correct value

**This has not been fixed.**

This seems to be a bug with chromium-based browsers, see https://issues.chromium.org/issues/361250561 

This bug is also present in native date input elements in html.

It works correctly in non-chromium browsers. Also I can't find a way to change what the screen reader announces when editing the different numbers of the date inside the datetime component.


ckrook (approved):

mistermalm:
open datepicker when pressing enter does not work on firefox

max-umain:
> open datepicker when pressing enter does not work on firefox

Good catch. I pushed some changes now where we only open the datepicker on enter key press in Chromium-based browsers, and also hide the duplicate icon in Firefox browsers.

In Safari and Firefox, I don't think there's a way to programmatically open the datepicker. In Firefox it now works when focusing on the datepicker icon (works by default), but in Safari it only works when clicking the input field, however the user can of course set a date by using the up and down keys while focusing on the different parts of the date/datetime-local (year, month, etc)

ckrook (approved):

PR #1138, opened 2025-04-11
feat(text-field): custom handling for min & max when input type is Number

patrik-nyfeldt:
## **Describe pull-request**  
Handling of input values when type is Number and Min or Max props is used.
Current behaviour allows typing of higher/lower values then set in max/min - the limits only applies when setting the number with input 'arrows'.

## **Issue Linking:**  
- **Jira:** Add ticket number after `CDEP-604`: [CDEP-604](https://jira.scania.com/browse/CDEP-604)
- **No issue:** 

## **How to test**  
1. Go to Storybook component Text Field
2. Set Type to Number and add a min/max value
3. Confirm that it is not possible to type higher/lower numbers 

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  

### DIFF
--- a/packages/core/src/components/text-field/text-field.tsx
+++ b/packages/core/src/components/text-field/text-field.tsx
@@ -101,10 +101,31 @@ export class TdsTextField {
 
   // Data input event in value prop
   handleInput(event: InputEvent): void {
-    this.value = (event.target as HTMLInputElement).value;
+    const inputEl = event.target as HTMLInputElement;
+    let { value } = inputEl;
+    
+    // Custom handling of number inputs when min is set
+    if (this.type === 'number' && this.min !== undefined) {
+      const numericValue = Number(value);
+      if (numericValue < Number(this.min)) {
+        value = String(this.min);
+        inputEl.value = value;
+      }
+    }
+  
+    // Custom handling of number inputs when max is set
+    if (this.type === 'number' && this.max !== undefined) {
+      const numericValue = Number(value);
+      if (numericValue > Number(this.max)) {
+        value = String(this.max);
+        inputEl.value = value;
+      }
+    }
+  
+    this.value = value;
     this.tdsInput.emit(event);
   }
-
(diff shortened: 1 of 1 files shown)


packages/core/src/components/text-field/text-field.tsx
```diff
@@ -101,10 +101,31 @@ export class TdsTextField {
 
   // Data input event in value prop
   handleInput(event: InputEvent): void {
-    this.value = (event.target as HTMLInputElement).value;
+    const inputEl = event.target as HTMLInputElement;
```
timrombergjakobsson, line 104: Nice, what do you think about doing like this:

```
handleInput(event: InputEvent): void {
  const inputEl = event.target as HTMLInputElement;
  let value = inputEl.value;

  if (this.type === 'number') {
    const numericValue = Number(value);

    if (this.min !== undefined && numericValue < Number(this.min)) {
      value = String(this.min);
    }

    if (this.max !== undefined && numericValue > Number(this.max)) {
      value = String(this.max);
    }

    inputEl.value = value;
  }
}
```
?
Then we don't have to repeat parsing of value and this.min/this.max. And we only set inputEl.value once after all conditions are checked.

max-umain (approved):
Nice!

packages/core/src/components/text-field/text-field.tsx
patrik-nyfeldt: Excellent idea, I'll test it! Thanks 🙏

ckrook (approved):

packages/core/src/components/text-field/text-field.tsx
patrik-nyfeldt: @timrombergjakobsson I pushed this change now if you want to have a look

PR #1144, opened 2025-04-16
fix(datetime): workaround firefox

mistermalm:
## **Describe pull-request**  
Firefox does not support, hiding/styling/overriding their native icon for input type date field.

This workaround uses the native firefox icon instead which looks OK i would say. To really fix this issue we might need to refactor and build our own custom datepicker.

Looks better then duplicate Icons anyways.

## **Issue Linking:**  
- **Jira:** `Duplicate Icon in Firefox`: [CDEP-362](https://jira.scania.com/browse/CDEP-362) 

## **How to test**  
1. Go to datetime component
2. Check that Icon is not duplicated
3. Repeat for Safari, Firefox and Chrome.

- [x] All existing tests pass

## **Suggested test steps**


timrombergjakobsson:
Tricky one this, there is no duplicate, but the padding gets a bit messed up:
Firefox:

Screenshot 2025-04-16 at 14 01 04

Chrome:

Screenshot 2025-04-16 at 14 01 54

Don't know if it can be adjusted, or if it is even worth it? Maybe design should have a say here as well?

packages/core/src/components/datetime/datetime.scss
```diff
@@ -399,3 +399,10 @@ input[type='time']::-webkit-inner-spin-button,
 input[type='time']::-webkit-calendar-picker-indicator {
   opacity: 0;
 }
+
+@supports (-moz-appearance: none) {
```
timrombergjakobsson, line 403: Yay first `@supports` in Tegel!

timrombergjakobsson (approved):

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -254,20 +254,20 @@ export class TdsDatetime {
             />
 
             <div class="datetime-icon icon-datetime-local">
-              <tds-icon size="20px" name="calendar" svgTitle="Calendar"></tds-icon>
+              <tds-icon size="20px" name="calendar" />
```
max-umain, line 257: I think we should keep the svgTitles for the icons, this is something that we get axe accessibility issues for

packages/core/src/components/datetime/datetime.scss
```diff
@@ -399,3 +399,10 @@ input[type='time']::-webkit-inner-spin-button,
...
+  .icon-datetime-local,
+  .icon-time {
```
max-umain, line 405: I don't think we should hide icon-time as firefox doesn't have a built-in icon that replaces this. Try going to the amplify preview and changing the type control, in firefox. 

Also now if the type is "month" no icon is shown in firefox either. If you want some inspiration on how this can be handled: https://github.com/scania-digital-design-system/tegel/pull/1137/commits/f632e38cb57a8460ac7cd5b7b850e1334a34385d

maxadolfsson:
@mistermalm @timrombergjakobsson - From the screenshot it looks like Firefox is adding more characters to the field, Chrome only shows `--:--` where Firefox add at least two more `--:-- --`. Do you see that as well?

I will take a look in Firefox now, it sounds like the suggestion is the way to go! 

> […] Looks better then duplicate Icons anyways.

packages/core/src/components/datetime/datetime.scss
maxadolfsson: In addition to `month` type, `time` and `week` also have no icons at all in Firefox.

mistermalm: so we should only hide datetime-local is that what you mean?

maxadolfsson: Additionally, the icon seems to be the trigger for the native calendar view. Without an icon it is not possible to trigger the view where you pick dates/time/weeks:
 
Screenshot 2025-04-24 at 11 16 38

max-umain:
> @mistermalm @timrombergjakobsson - From the screenshot it looks like Firefox is adding more characters to the field, Chrome only shows `--:--` where Firefox add at least two more `--:-- --`. Do you see that as well?
> 
> I will take a look in Firefox now, it sounds like the suggestion is the way to go!
> 
> > […] Looks better then duplicate Icons anyways.

Yeah I noticed too that Firefox and Chrome have different formats on the date, and one of them have 12 hour format instead of 24. I haven't looked that much into it but I don't think it can be fixed

timrombergjakobsson:
> > @mistermalm @timrombergjakobsson - From the screenshot it looks like Firefox is adding more characters to the field, Chrome only shows `--:--` where Firefox add at least two more `--:-- --`. Do you see that as well?
> > I will take a look in Firefox now, it sounds like the suggestion is the way to go!
> > > […] Looks better then duplicate Icons anyways.
> 
> Yeah I noticed too that Firefox and Chrome have different formats on the date, and one of them have 12 hour format instead of 24. I haven't looked that much into it but I don't think it can be fixed

Yes, I think still this is better than double icons, as long as design approves this compromise for now. And then in the long run we should create a custom component instead?

mistermalm:
@maxadolfsson @max-umain 
I pushed an update for this, now icon should be visible for all different types, and no duplicates, it also seems like it the Tegel Icon is shown in both cases which is good!

max-umain (approved):

PR #1145, opened 2025-04-16
fix(side-menu): align padding with figma

mistermalm:
## **Describe pull-request**  
Align padding with [Figma](https://www.figma.com/design/Oyi7D6ATKzqtohR9xTsd8X/Side-menu-measurements?node-id=0-1&p=f&m=dev) example.

## **Issue Linking:**  
- **Jira:** [CDEP-209](https://jira.scania.com/browse/CDEP-209) `Align padding in code with figma design`

## **How to test**  
1. Go to Side-menu component
2. Inspect elements and compare with figma design

### DIFF
--- a/packages/core/src/components/side-menu/side-menu-close-button/side-menu-close-button.scss
+++ b/packages/core/src/components/side-menu/side-menu-close-button/side-menu-close-button.scss
@@ -1,9 +1,11 @@
 :host {
+  margin-left: -2px;
+
   button {
     height: 64px;
     width: 100%;
     text-align: left;
-    padding: 0 24px;
+    padding: 0 22px;
     border: none;
     background-color: var(--tds-sidebar-side-menu-background-cover);
     font: var(--tds-headline-07);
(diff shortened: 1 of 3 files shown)


packages/core/src/components/side-menu/side-menu-close-button/side-menu-close-button.scss
```diff
@@ -1,9 +1,11 @@
 :host {
+  margin-left: -2px;
```
theJohnnyMe, line 2: Can we avoid this somehow?

max-umain (approved):

packages/core/src/components/side-menu/side-menu-close-button/side-menu-close-button.scss
timrombergjakobsson: Maybe can be fixed in the parent layout?

timrombergjakobsson: True, `transform: translateX(-2px);` is slightly less risky than negative margins. However feels a bit odd to use it as primarily meant for animations or micro-adjustments.

PR #1152, opened 2025-04-22
ci(release): add early git tagging after core publish and safeguard

timrombergjakobsson:
## **Describe pull-request**  
In our current release workflow, the Git tag (e.g., @scania/tegel@1.28.0) is only created at the very end of the pipeline, after merging the release branch into develop.

This causes an issue when the core package is published successfully to npm, but the Git tag hasn’t been created yet. If the pipeline fails afterward (e.g., a wrapper build fails), our next release might incorrectly calculate the version based on the previous tag, leading to misaligned or even lower version numbers (e.g., 1.27.1 after 1.28.0).

Solution
This PR adds:

✅ A new step in the release-core job to create and push the Git tag immediately after the core is published.

✅ A safeguard in the create-git-tag job that conditionally checks if the tag already exists before trying to push it again.

This ensures:

Git and npm always stay in sync, even if the workflow fails halfway.

No duplicate tag errors.

Future version bumps are always calculated correctly.

### DIFF
--- a/.github/workflows/holy-grail.yml
+++ b/.github/workflows/holy-grail.yml
@@ -132,6 +132,14 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+      # ✅ Tag immediately after successful publish
+      - name: Core - Tag version
+        working-directory: ${{ env.WORKING_DIRECTORY }}
+        run: |
+          VERSION=$(jq -r .version package.json)
+          git tag @scania/tegel@$VERSION
+          git push --no-verify origin @scania/tegel@$VERSION
+
       - name: Core - Commit and push changes
         run: |
           git add .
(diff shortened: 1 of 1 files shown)


.github/workflows/holy-grail.yml
```diff
@@ -132,6 +132,14 @@ jobs:
...
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+      # ✅ Tag immediately after successful publish
+      - name: Core - Tag version
+        working-directory: ${{ env.WORKING_DIRECTORY }}
+        run: |
+          VERSION=$(jq -r .version package.json)
+          git tag @scania/tegel@$VERSION
+          git push --no-verify origin @scania/tegel@$VERSION
+
```
theJohnnyMe, line 142: For this to work for us, git tag has to be "baked in" in develop branch, so we would need to make sure that is in place.

timrombergjakobsson: But we also have a fallback that checks and creates the tag again from develop after the merge.

If we want to guarantee the tag always points to a commit in develop, we could either:

- Rely on the fallback step (as-is), or
- Force-move the tag to the merged commit in develop after the squash merge?

Issue #1143, opened 2025-04-16
[Feature Request]: add property label-position to tds-datetime


QuentinBourasse1:
### Requirements before reporting

- [x] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [x] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

Adding the property **label-position** to the component **tds-datetime** with the values : inside / outside / 'no-label' = 'no-label'.

The idea is to have the possibility to choose the position of the label of **tds-datetime** like we are able to do with other components like **tds-dropdown** or **tds-textfield**

### Background

In order to standardize everything in forms and to not create some visuals mismatches like below : 

![Image](https://github.com/user-attachments/assets/29f67525-02a6-4a07-85cb-eca6a2d64e71)

### Idea

Adding the same property **label-position** and same style that **tds-dropdown** or **tds-textfield**

### Goals and objectives

Be able to align correctly items without overloading the css style

### Contact information

quentin.bourasse@scania.com

mjukiscissi:
Hi @QuentinBourasse1! Good idea. Let us have a look at this and get back to you!

max-umain:
Hi @QuentinBourasse1 

Do you want to be able to position the label inside the datetime component similarly to how it can be done for Text field? See https://tds-storybook.tegel.scania.com/?path=/story/components-text-field--default&args=labelPosition:Outside

QuentinBourasse1:
Hi @max-umain ! Yes it is exactly what I wish

PR #1162, opened 2025-04-24
feat(datetime): add labelPosition prop

max-umain:
## **Describe pull-request**  
Adds the labelPosition prop to datetime, to allow the label to be positioned inside of the component, in addition to above it.

## **Issue Linking:** 
- **Jira:** [CDEP-658](https://jira.scania.com/browse/CDEP-658)
- **GitHub:** https://github.com/scania-digital-design-system/tegel/issues/1143

## **How to test**  
1. Go to https://pr-1162.d3fazya28914g3.amplifyapp.com/?path=/story/components-datetime--default
2. Try changing the labelPosition in the storybook controls

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

## **Screenshots**  
None

## **Additional context**  
1: In the figma design (for the date picker component), when the date hasn't been set yet and the component isn't focused, only the label should be shown, and not both the label and the "date placeholder". There is no trivial way I could find to achieve this. So I suggest we leave it like this for now, i.e. that the "focused" state in figma is the one that is always shown, even when the datetime's value hasn't been modified.

~2: I set the default labelPosition to outside (rather than no-label which is the default for other components such as text-field) to not cause any "breaking changes" for users currently using tegel, so that their label doesn't just disappear (with the current solution before this PR, there is no labelPosition prop so the label is always shown as long as the label prop contains some text~


ckrook:
Looks good. However I'm wondering if this have been approved by design :)

maxadolfsson:
Nice catch @max-umain @ckrook! I think it should behave in the very same way as the Text Field, where the inside label sits nicely positioned in the field (idle state), then moves out of the way and make room for the placeholder text (focused state). When checking this now (in mobile) the inside label does not really sit where I would expect it to sit :)

I will have to double-check Figma to validate how it looks in there.  With that said, would it work for you?

max-umain:
@maxadolfsson does it look good on your computer?

I checked on my phone now as well and I get the same problem as you. This problem isn't visible when inspecting on computer and setting the page to a mobile size. I suspect this is because browsers have built-in styling for date type input elements, and the one on mobile chrome isn't the same as on desktop. I'll look at how to solve this

max-umain:
@maxadolfsson 

> I think it should behave in the very same way as the Text Field, where the inside label sits nicely positioned in the field (idle state), then moves out of the way and make room for the placeholder text (focused state).

This is quite hard to achieve due to the fact that the "placeholder" and date text styling is built-in in the browser. I might be able to fix it, but it would take some time, so I suggest we don't fix it and instead fix it in the date-picker component when we replace datetime with that one.

So my suggestion is to keep it in the focused state. What do you think about it?

theJohnnyMe:
As we spoke, see if you can add:

1. Clicking on label outside sets component in focus
2. Component in focus has blue underline (line Text-Field and Textarea)

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -50,6 +51,9 @@ export class TdsDatetime {
   /** Label text for the component */
   @Prop() label: string = '';
 
+  /** Position of the label */
+  @Prop() labelPosition: 'inside' | 'outside' | 'no-label' = 'outside';
+
```
theJohnnyMe, line 56: Set "no-label" as default as it is a new prop so we do not surprise a user with new feature. Also other components tend to have "no-label" as fallback/default option.

```suggestion
  /** Position of the label */
  @Prop() labelPosition: 'inside' | 'outside' | 'no-label' = 'no-label';

```

packages/core/src/components/datetime/datetime.stories.tsx
```diff
@@ -115,6 +115,17 @@ export default {
...
+    labelPosition: {
+      name: 'Label position',
+      description: 'Sets the label text position.',
+      control: {
+        type: 'radio',
+      },
+      options: ['No label', 'Inside', 'Outside'],
+      table: {
+        defaultValue: { summary: 'outside' },
+      },
```
theJohnnyMe, line 127: Do the same here.

```suggestion
      table: {
        defaultValue: { summary: 'no-label' },
      },
```

theJohnnyMe (dismissed):

theJohnnyMe:
Minot things to correct, otherwise, great work.

packages/core/src/components/datetime/datetime.tsx
max-umain: I addressed this here

> 2: I set the default labelPosition to outside (rather than no-label which is the default for other components such as text-field) to not cause any "breaking changes" for users currently using tegel, so that their label doesn't just disappear (with the current solution before this PR, there is no labelPosition prop so the label is always shown as long as the label prop contains some text

Considering that, do you still think no-label should be defaul?

theJohnnyMe: Yes, please do set it to "no-label". Thanks for double check ;)

PR #1186, opened 2025-05-12
fix(readme): update to correct version of node in reamde text

timrombergjakobsson:
## **Describe pull-request**  
Old version of Node mentioned

## **Issue Linking:**  
- **No issue:** Updating README.md with correct Node version.


theJohnnyMe (approved):
Nice catch!

PR #1187, opened 2025-05-12
feat: storybook 8.6

theJohnnyMe:
WORK IN PROGRESS - DO NOT MERGE


mjukiscissi:
Looks nice @theJohnnyMe @mistermalm! Great with clearer structure in the side menu. Some thoughts from me:

- I suggest we move the Icon-story from Foundations to Components so that we establish it more as the actual icon component. The Icon-gallery can stay as is within foundations. 
- Can we move Patterns down beneath Components in the side menu?
- Can motion be part of Foundations? then we have less headings in the side menu.
- We have now Color within both Utilities and Foundations. Can we make a clearer distinction between them?
- Within a story, we should be more consistent on the structure (see image for reference). In this case I think we should have Default and Nested directly after each other so that Notes and Docs comes in the end of the dropdown OR group the variants as we do with Table.
Skärmbild 2025-05-13 134536
image

theJohnnyMe:
@cissi1995 all done, please check on preview link.
Let me know if something is missing.

mjukiscissi (approved):
Looks good! Great job with this :)

PR #1198, opened 2025-05-22
feat(tokens): Semantic color tokens 

timrombergjakobsson:
## **Describe pull-request**  
Base semantic tokens layer to be used in components.


theJohnnyMe (approved):
LGTM

packages/core/src/global/global.scss
mistermalm, line 1: should this file be names scania.scss? 
since the other one in global is named traton.scss?

mistermalm (approved):

PR #1206, opened 2025-05-27
docs(toast): add accessibility notes

ckrook:
## **Describe pull-request**  
This PR adds accessibility documentation for the Toast component to address screen reader announcement issues. 

When toasts render on initial page load, screen readers often fail to announce their content because they only detect changes in aria-live regions after the page has loaded. The documentation provides a straightforward solution: either render toasts initially hidden and reveal them with showToast() after a brief delay, or create them dynamically when needed. A practical code example demonstrates the proper implementation, helping developers create more inclusive user experiences. 

## **How to test**
Read through the documentation and make sure its correct.

## Extra Disclaimer
I put the documentation above the  comment.
I think this is the way to get around my docs not being overwritten. Please let me know if this is incorrect.

### DIFF
--- a/packages/core/src/components/toast/readme.md
+++ b/packages/core/src/components/toast/readme.md
@@ -1,5 +1,31 @@
 # tds-toast
 
+## Accessibility Notes
+
+### ARIA Live Announcements
+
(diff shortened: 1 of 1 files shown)


theJohnnyMe (approved):
LGTM

PR #1223, opened 2025-06-03
refactor(block): implementation of new variable tokens

mistermalm:
## **Describe pull-request**  
Refactor block component to use new variable tokens

## **Issue Linking:**  
- **Jira:** `block component tokens`: [CDEP-744](https://jira.scania.com/browse/CDEP-744)

## **How to test**  
1. Go to Block component in storybook
2. Check that it looks ok.
3. Try primary/secondary mode
4. Add tds-block-even or tds-block-odd class
5. Switch brand


packages/core/src/components/block/block.scss
```diff
@@ -1,28 +1,37 @@
 @import '../../mixins/box-sizing';
+@import '../../../../../typography/utilities/typography-utility';
+
+:host {
+  --primary-block-background-color: var(--color-background-layer-01);
+  --secondary-block-background-color: var(--color-background-layer-02);
+  --text-color: var(--color-foreground-text-strong);
+}
```
theJohnnyMe, line 8: Host wont work here as component does not <Host syntax in `block.tsx` file. Try replacing `:host` with `.tds-block` as that class is all the time applied.
After it make sure that both Default and Nested story work the same way as compared to prod Storybook. Now it seems that Nested is broken and "Inherit from parent" does not work in any of them.

ckrook (commented):
@mistermalm
I've noticed that this block behaves differently from the one currently in production when Inherit from parent is selected.

In production, Inherit from parent seems to adopt the primary mode, whereas in this branch, it defaults to secondary mode.

PR #1240, opened 2025-06-09
feat(text-field): new tokens usage

theJohnnyMe:
## **Describe pull-request**  
Introducing new tokens to Text Field component to support Scania and Traton designs.

## **Issue Linking:**  
- **Jira:** [CDEP-764](https://jira.scania.com/browse/CDEP-764)

## **How to test**  
1. Open [Traton UI Kit: Text Field component](https://www.figma.com/design/6osYTOfd4MgDq7LO6BCoUO/TRATON-Component-Library?node-id=26756-59844&m=dev)
2. Open preview link of this PR
3. Compare styles from with with one present in Storybook

## **Additional context**  
Icon component is improved so now it should be able to switch even when it is nested in ShadowDOM components like Message. 
edit_inactive icon is added to Traton as it was missing from icons gallery that Traton has while we need it for read-only state. 
Separate file for overrides is made to handle difference for components that have a lot of changes - to be discussed if we can scope this down to component level or even have in component scss file instead.


packages/core/src/components/text-field/text-field.tsx
```diff
@@ -289,6 +287,7 @@ export class TdsTextField {
           )}
           <span class="text-field-icon__readonly-label">This field is non-editable</span>
         </div>
+        <div class="text-field-bar"></div>
```
mistermalm, line 290: unless you have anything in the div you can just close it directly
`<div class="text-field-bar"/>`

packages/core/src/components/text-field/text-field.stories.tsx
```diff
@@ -289,18 +289,21 @@ const Template = ({
...
+            ? prefixType === 'Text'
+              ? '<span slot="prefix">$</span>'
+              : '<tds-icon slot="prefix" name="truck" size="20px"></tds-icon>'
             : ''
         }
+        ${
+          suffix
+            ? suffixType === 'Text'
+              ? '<span slot="suffix">$</span>'
+              : '<tds-icon slot="suffix" name="truck" size="20px"></tds-icon>'
```
mistermalm, line 302: `<tds-icon slot="suffix" name="truck" size="20px"/>`

mistermalm (approved):
just added minor JSX improvements as comments if you want to fix it :)

mistermalm (commented):
If you go to [Dropdown](https://tds-storybook.tegel.scania.com/?path=/story/components-dropdown--default) in production, and compare with Text Field since they both are kind of a input field. For dropdown when you use label inside and remove the placeholder text, the label gets the same color as placeholder. This is not the case for Text field, is this intended or should one of them be changed for consistency?

packages/core/src/components/text-field/text-field.stories.tsx
```diff
@@ -289,18 +289,21 @@ const Template = ({
...
-              ? `<span slot="${prefix ? 'prefix' : 'suffix'}">$</span>`
-              : `<tds-icon slot="${
-                  prefix ? 'prefix' : 'suffix'
-                }" name="truck" size="20px"></tds-icon>`
-          }
-        `
+          prefix
+            ? prefixType === 'Text'
+              ? '<span slot="prefix">$</span>'
+              : '<tds-icon slot="prefix" name="truck" size="20px"></tds-icon>'
```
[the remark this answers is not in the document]
theJohnnyMe: Trick does not work in stories, it seems to need to be like it is as if I change it, I am not getting both prefix and suffix, only one of them renders.

packages/core/src/components/text-field/text-field.tsx
theJohnnyMe: Good idea!

theJohnnyMe:
> If you go to [Dropdown](https://tds-storybook.tegel.scania.com/?path=/story/components-dropdown--default) in production, and compare with Text Field since they both are kind of a input field. For dropdown when you use label inside and remove the placeholder text, the label gets the same color as placeholder. This is not the case for Text field, is this intended or should one of them be changed for consistency?

Nice catch, fixed!

Issue #1226, opened 2025-06-03
[Feature Request]: Support for name and value attributes on <tds-button>


klarajons:
### Requirements before reporting

- [x] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [x] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

Add support for **name** and **value** attribute on the `<tds-button>`.  NOTE: name and value need to be passed to the native button element as well.

`<TdsButton
            id="kc-submit"
            type="submit"
            name="accept"
            value="yes"
          />`

### Background

In our application, we're integrating with the Keycloak API, which expects a specific key-value pair based on which form button is clicked. Not being able to set name and value directly on the button element makes it challenging to support this use case without "not-so-pretty" workarounds.

For context we are using the React version of the library.

### Goals and objectives

- Being able to use a TdsButton like this `<TdsButton
            id="kc-submit"
            type="submit"
            name="accept"
            value="yes"
          />`
- Clicking this button should submit the form and include the corresponding key-value pair in the `formData`.

### Contact information

klara.jonsson@scania.com

klarajons:
Works like a charm with 1.31.1-feat-button-name-attribute-beta.2 👍

max-umain:
A fix for this has now been merged to the main branch, and will be available in the next official release.

PR #1244, opened 2025-06-10
feat(button): name and value props

max-umain:
## **Describe pull-request**  
This introduces the props name and value that can be specified for the button component, which is sometimes necessary when using the button in certain forms.

## **Issue Linking:**  
- **GitHub:** https://github.com/scania-digital-design-system/tegel/issues/1226

## **How to test**  
1. Go to https://pr-1244.d3fazya28914g3.amplifyapp.com/?path=/story/components-button--default&args=name:name1;value:value1;brand:scania
2. Open browser developer tools and inspect the button
3. Verify that both the tds-button and button elements have the attributes name="name1" and value="value1"

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
None

## **Additional context**  
None

### DIFF
--- a/packages/core/src/components/button/button.tsx
+++ b/packages/core/src/components/button/button.tsx
@@ -76,8 +76,8 @@ export class TdsButton {
       >
         <button
           type={this.type}
-          name={this.name}
-          value={this.value}
+          name={this.name ? this.name : undefined}
+          value={this.value ? this.value : undefined}
           disabled={this.disabled}
           class={{
             'primary': this.variant === 'primary',


timrombergjakobsson (approved):

packages/core/src/components/button/button.tsx
```diff
@@ -70,6 +76,8 @@ export class TdsButton {
       >
         <button
           type={this.type}
+          name={this.name}
+          value={this.value}
```
ckrook, line 80: Since the value attribute is used during form submission, it's undesirable to include it if its value might be undefined. A more clean approach would be to conditional render it

```
<button
  type={this.type}
  {...(this.value ? { value: this.value } : {})}
  ...
  ```

ckrook (changes requested):

PR #1304, opened 2025-07-07
fix(holy-grail): ensure npm version bump works in scheduled runs

augustinescania:
## **Describe pull-request**  
Scheduled (cron) triggers don’t provide workflow inputs like version, which previously caused npm version to silently fail during automated dry runs.

This PR introduces a global env: block to define fallback defaults:

image

Now, version bumping and related actions use:

`${{ github.event.inputs.version || env.VERSION }}`

This ensures versioning works reliably in both manual and scheduled executions.
## **Issue Linking:**  

## **How to test**  
1. Go to...
2. Check in...
3. Run ...

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  

### DIFF
--- a/.github/workflows/holy-grail.yml
+++ b/.github/workflows/holy-grail.yml
@@ -38,6 +38,11 @@ on:
   schedule:
     - cron: '0 3 * * 1-5' #3AM every weekday
 
+env:
+  VERSION: patch
+  NODE_VERSION: '22.11.0'
+  TAG: latest
+
 jobs:
   #CREATE RELEASE BRANCH
   create-release-branch:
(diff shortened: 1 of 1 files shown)


theJohnnyMe (approved):
I think this should work.

What could be an improvement is to define default values as environmental variables -  "env" so all it is controlled from one place and maybe multiple version would be needed like:

`env:

  VERSION: patch
  NODE_VERSION: '22.11.0'
`

 And then use it like this in rest of code:
 
 `run: npm version ${{ github.event.inputs.version || env.VERSION }} --no-git-tag-version`
 
 Worth exploring.

augustinescania:
> I think this should work.
> 
> What could be an improvement is to define default values as environmental variables - "env" so all it is controlled from one place and maybe multiple version would be needed like:
> 
> `env:
> 
> VERSION: patch NODE_VERSION: '22.11.0' `
> 
> And then use it like this in rest of code:
> 
> `run: npm version ${{ github.event.inputs.version || env.VERSION }} --no-git-tag-version`
> 
> Worth exploring.

Good idea @theJohnnyMe! I implemented your suggestion, let me know what you think! :)

ckrook (approved):

PR #1306, opened 2025-07-08
feat(dropdown): add documentation on updateDisplayValue

ckrook:
## **Describe pull-request**  
We received a complaint in our development support channel where a user did not have the possibility to update the dropdown display value when the text content of dropdown options was programmatically changed. This was because the dropdown component wasn't automatically detecting text content changes in its options.

## **Issue Linking:**  
- [Teams conversation](https://teams.microsoft.com/l/message/19:5e33f67fe502441f914fbcdc6e2548f5@thread.skype/1751374723404?tenantId=3bc062e4-ac9d-4c17-b4dd-3aad637ff1ac&groupId=79f9bfeb-73e2-424d-9477-b236191ece5e&parentMessageId=1751374723404&teamName=Tegel%20Design%20System&channelName=Development%20support%20-%20Tegel&createdTime=1751374723404)
- Jira: [CDEP-1081](https://jira.scania.com/browse/CDEP-1081)
## **How to test**  
1. Go to my [demo page](https://pr-243.d2vh2lmnzgzrkl.amplifyapp.com/about)
2. Follow the instructions to test that this works as I would like

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -138,6 +140,13 @@ export class TdsDropdown {
       .filter((v) => v !== '');
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  private hasValueChanged(newValue: string[], currentValue: string[]): boolean {
+    if (newValue.length !== currentValue.length) return true;
+    return newValue.some((val) => !currentValue.includes(val));
+  }
```
augustinescania, line 147: I noticed we now have two hasValueChanged functions (one top-level helper + one private class method). Is that intentional, or can we drop one? Please see **line 20-23** in this file.

`function hasValueChanged(newValue: string[], currentValue: string[]): boolean {`
`  if (newValue.length !== currentValue.length) return true;`
`  return newValue.some((val) => !currentValue.includes(val));`
`}`

ckrook: Interesting that this function came in my PR. This code was added by Tim 4 months ago...

ckrook: Probably something that should not be in the PR since @max-umain merged function hasValueChanged 8 hours ago. And this function you are referenced is not used :)

theJohnnyMe (changes requested):
While mutationObserver is nice way to solve this, it may affect performance. Looking at discussion with user, user is programmatically updating component text which is intervention that is not a common case when using dropdown. I would propose we rely only on having method for display update, so when develop start consciously to interacts with component programmatically, let it call the method that will update the display too. It becomes predictable too, developer knows exactly when it happen while mutationObserver is less predictable. Also, we tend to make dropdown logic less complex.
 
Please update the code and provide an example in dropdown readme.

PR #1391, opened 2025-08-13
feat(side-menu): add multibrand tokens

max-umain:
## **Describe pull-request**  

## **Issue Linking:**  

## **How to test**  
1. Go to...
2. Check in...
3. Run ...

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  


packages/core/src/components/icon/icon.tsx
```diff
@@ -26,7 +26,7 @@ export class Icon {
   @Prop({ reflect: true }) name: IconNames = 'truck';
 
   /** Pass a size of icon as a string, for example, 32px, 1rem, 4em... */
-  @Prop({ reflect: true }) size: string = '16px';
+  @Prop({ reflect: true }) size: string = undefined;
```
timrombergjakobsson, line 29: I know its still in progress, but I think you can just do :

```suggestion
  @Prop({ reflect: true }) size?: string;
```
or 

```suggestion
  @Prop({ reflect: true }) size: string;
```

packages/core/src/components/icon/icon.tsx
```diff
@@ -42,10 +42,27 @@ export class Icon {
   @State() arrayOfIcons = [];
 
   componentWillLoad() {
-    this.detectAndSetBrand();
+    const brand = this.detectBrand();
```
timrombergjakobsson, line 45: Maybe we should consider flattening and simplifying the componentWillLoad() logic as shown below? This makes the brand-based sizing clearer and avoids nested conditionals- Something like this?

```
componentWillLoad() {
  const brand = this.detectBrand();
  this.setBrand(brand);

  if (this.size) return;

  const insideSideMenu = this.host.closest('tds-side-menu');

  const brandSizeMap: Record<string, string> = {
    scania: '24px', <-- should be replaced with new dimension values when in place
    traton: '16px', <-- should be replaced with new dimension values when in place
  };

  if (insideSideMenu && brandSizeMap[brand]) {
    this.size = brandSizeMap[brand];
  }

  if (!this.size) {
    this.size = '16px';
  }
}
```

packages/core/src/components/side-menu/side-menu-item/side-menu-item.scss
```diff
@@ -11,36 +11,45 @@
...
       width: 100%;
       display: flex;
       align-items: center;
-      gap: 16px;
-      padding: 0 22px 0 16px;
+      gap: var(--side-menu-space-sub-menu-item-profile-gap);
+      padding: var(--side-menu-space-sub-menu-item-profile-top-bottom)
+        var(--side-menu-space-sub-menu-item-chevron-right-padding)
+        var(--side-menu-space-sub-menu-item-profile-top-bottom)
+        var(--side-menu-space-sub-menu-item-profile-left);
```
timrombergjakobsson, line 45: I think we said that if one variable only would be used one time that it is redundant to allocate it to a new one.

timrombergjakobsson: Like this one:
https://github.com/scania-digital-design-system/tegel/pull/1391/files#diff-c24b3143436686758e984d21f8e3c65a00e74175b1985c3b90c01e3af9aaa9f3R11

timrombergjakobsson:
Regarding this:

> In Figma, the element that in this example contains the text “Wheel types”, has a padding-left of 2px. However, I didn’t find a good way to fix this in the code because the text itself isn’t actually inside its own element. However, this is barely noticeable visually. There are a couple different approaches to solving this, but I want to discuss it with the team more to ensure compatibility for projects already using the side-menu.

Could we maybe use pseudo selectors? Like `:before`

packages/core/src/components/side-menu/side-menu-item/side-menu-item.scss
timrombergjakobsson: Yes exactly as it is only used once right?

max-umain:
> Regarding this:
> 
> > In Figma, the element that in this example contains the text “Wheel types”, has a padding-left of 2px. However, I didn’t find a good way to fix this in the code because the text itself isn’t actually inside its own element. However, this is barely noticeable visually. There are a couple different approaches to solving this, but I want to discuss it with the team more to ensure compatibility for projects already using the side-menu.
> 
> Could we maybe use pseudo selectors? Like `:before`

I'm not sure that would work. Do you mean on the parent element? In that case that would push the icon (if there is one) 2px to the right, and we only want the text to be pushed to the right

timrombergjakobsson:
> > Collapsed state for Traton looks a bit off?
> > Screenshot 2025-08-20 at 10 11 45 Screenshot 2025-08-20 at 10 11 26
> 
> I noticed in your screenshot that the "Rim wheel" button is one pixel too short, so I fixed this.
> 
> But I'm assuming you're referring to the dropdown not being aligned with the side-menu button (under certain circumstances). This is something I noticed too, and this is actually a problem with Scania in production too (under the same circumstances), so I'll create a ticket for it.
> 
> You can replicate it here https://tds-storybook.tegel.scania.com/?path=/story/patterns-navigation-many-navigation-items--many-navigation-items by having the browser window short enough to see the scrollbar in side-menu, and the more you scroll down in the side-menu, when you hover the dropdown button (which goes higher up while you scroll), you'll notice they become more misaligned. It seems the dropdown shows up right next to where the button for it was originally when you haven't scrolled at all.

Yes exactly, but great that you create a ticker for it @max-umain!

packages/core/src/components/side-menu/side-menu-dropdown/side-menu-dropdown.scss
```diff
@@ -1,6 +1,14 @@
+@import '../../../../../../typography/utilities/typography-utility';
+
 :host {
   display: block;
 
+  // Custom variables for dropdown colors when not collapsed:
+  --side-menu-dropdown-list-item-custom-background: var(--side-menu-background-cover);
```
timrombergjakobsson, line 7: Isn't this already a custom variable?

max-umain: Yes, but for example with the variable --side-menu-dropdown-list-item-custom-background, when the dropdown is collapsed, it needs to point to another variable that is brand dependent, so for this one, when not collapsed:
`--side-menu-dropdown-list-item-custom-background: var(--side-menu-background-cover);`
and when collapsed (further down in that file):
`--side-menu-dropdown-list-item-custom-background: var(--side-menu-collapsed-dropdown-background);`

ckrook (approved):
@max-umain From a design perspective, the side menu looks fine to me. It might be worth getting one or two more opinions on how the variables are being used.

timrombergjakobsson (approved):
Some small tweaks still left, but let's ship it and iterate on those. Great job @max-umain!

PR #1412, opened 2025-08-28
fix(dropdown): remove focus outline on click

ckrook:
## **Describe pull-request**  
This PR fixes an issue with the dropdown component where both the dropdown trigger and its list items display a focus outline when clicked.

The expected behavior is that focus outlines should only appear during keyboard navigation, not on mouse clicks.

The fix replaces `:focus` with `:focus-visible`, ensuring outlines are only shown when appropriate.

## **Issue Linking:**  
- **Jira:**  [CDEP-1124](https://jira.scania.com/browse/CDEP-1124)

## **How to test**  
1. Go to the [dropdown component](https://pr-1412.d3fazya28914g3.amplifyapp.com/?path=/story/components-dropdown--default) in the preview link
2. Try to both click and navigate the dropdown with keyboard.
3. Make sure that the outline focus only appear when you navigate the dropdown with the keyboard.
4. Also make sure this behaviour works the same way in other major browsers.

- [x] All existing tests pass

## **Suggested test steps**
- [x] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
https://github.com/user-attachments/assets/41befec5-e06f-44ee-9167-99b5310335d9

## **Additional context**  
I had to also update the snapshots for the table pagination since we are using the dropdown within the table here


maxadolfsson:
I'm not sure this is a desired change from a UX perspective — if it is not too late to challenge, I would like us to revisit this before merging.

The action of clicking or tapping on a form element such as a Dropdown is to give it focus, quite literally. Should be same behaviour as a text field or textarea, for consistency reasons.

maxadolfsson:
> Not sure how I should move forward with this. The decision is yours!

Let's talk :)

maxadolfsson:
I promised @ckrook to revisit this, and I had to remind myself about our reasoning a little bit. TL;DR — do not remove focus outline on click. 

---

After looking at [the form demo page](https://react-demo.tegel.scania.com/form) I think it becomes quite clear that the interactive elements absolutely need a focus outline whenever they are focused (sounds obvious when saying it like that) – regardless if you use keyboard to focus on the dropdown, or if you tap with your finger or click with a mouse. 

Sometimes, and even in our own demo page, it is hard to know which element the list of options actually belongs to (list can show up both above and below), and sometimes it even blends too much with the background (could be a primary/secondary mode thing, in combination with the background of the form wrapper). 

Adding screenshots here, let me know if this reasoning makes sense or not. All in all, I would not recommend to get rid of the focus outline in any case for interactive elements. We should rather strive to improve the consistency between several form components (Text Field, Textarea, Radio Button, Checkbox, and so on).

---

From Storybook in this PR, the difference between keyboard focus (with focus outline) and click to focus (without focus outline):

Screenshot 2025-09-16 at 16 00 02

Screenshot 2025-09-16 at 16 00 37

From demo page, outside of this PR, where it is blending with the background and hard to know where the dropdown list starts or ends – the focus outline makes the difference:

Screenshot 2025-09-16 at 16 41 53
Screenshot 2025-09-16 at 16 41 50
Screenshot 2025-09-16 at 16 41 55

---

Let me know what you think!

PR #1414, opened 2025-08-29
feat: synced back tokens folder with variables against tegel-light

timrombergjakobsson:
## **Describe pull-request**  
Manual update of token folder. Adding some updates from variables, adding `.tl`-selector to the scope.


max-umain (approved):
Looking at all these files together I notice that the structure of them are very inconsistent, and in some cases it seems like some things are missing, although it probably does work correctly for the most part as it's from approved code. However, maybe we should look into improving the structure and make sure that it's correct, but you can decide if you think this PR is ready to be merged anyway.

timrombergjakobsson:
> @timrombergjakobsson I Tried this locally. The styling works without issues in Tegel light. But i noticed that the icons don't show up anymore.
> 
> To replicate: Open up for e.g link component in tegel light. Notice that if you want to click on the dropdown controller to change icon - nothing happens when you click on it (Probably because it's not aware anymore of where the icons are located)
> 
> Also when you try to run `npm run build:tegel-light` locally, you get an error that the icon variables are undefined.

Hm, changing of the links works for me @augustinescania

augustinescania (approved):
LGTM now! 😄

tokens/scss/component/chip.scss
```diff
@@ -30,7 +30,7 @@ tds-chip {
   --component-chip-foreground-foreground-disabled-selected: var(
     --color-foreground-text-inverse-subtle
   );
-  --component-chip-border-selected-focus-border: var(--color-system-info-default);
+  --component-chip-border-selected-focus-border: var(--color-system-info-defauslt);
```
augustinescania, line 33: I Guess this is not intentional 🙂

PR #1415, opened 2025-08-29
feat(tag): introducing new tag component

ckrook:
## **Describe pull-request**  
Introducing the new tag component in Tegel library

## **Issue Linking:**  
- **Jira:** [CDEP-1317](https://jira.scania.com/browse/CDEP-1317)

## **How to test**  
1. Go to [the tag component](https://pr-1415.d3fazya28914g3.amplifyapp.com/?path=/story/components-tag--default) in Storybook
2. Compare the component with [the figma design](https://www.figma.com/design/d8bTgEx7h694MSesi2CTLF/Tegel-UI-Library?node-id=75607-1218&p=f&t=8gQAD6mpeh3kgXqv-0)
3. Test the component in various major browsers

4. Now look at the tag component in the three beta releases I have done for [React](https://pr-250.d2vh2lmnzgzrkl.amplifyapp.com/), [Angular](https://pr-147.d1ax919pz49q2z.amplifyapp.com/) & [Angular-17](https://pr-67.d24qjtb7na4hmr.amplifyapp.com/) And make sure it also looks fine.

5. Make sure the code looks OK
6. Make sure the documentation looks OK

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  


packages/core/src/components/tag/tag.scss
```diff
@@ -0,0 +1,82 @@
+@import './tag-vars';
+@import '../../mixins/box-sizing';
+
+:host {
+  @include tds-box-sizing;
+
+  display: inline-block;
+}
+
+:host {
```
timrombergjakobsson, line 10: Is this intended, the repetition?
Otherwise can be simplified:

```
:host {
  @include tds-box-sizing;

  display: inline-flex;
  align-items: center;
  border-radius: 4px;
  background-color: var(--tds-grey-150);
  color: var(--tds-grey-700);

  &.tds-tag-large {
    padding: 4px;
    gap: 4px;
  }
}

```

packages/core/src/components/tag/tag.scss
```diff
@@ -0,0 +1,82 @@
+@import './tag-vars';
```
timrombergjakobsson, line 1: This can be removed from here and then imported in core.scss

packages/core/src/components/tag/tag.tsx
```diff
@@ -0,0 +1,43 @@
...
+  @Prop() size: 'Large' | 'Small' = 'Large';
+
+  /** Sets the variant mode of the tag */
+  @Prop() variant: 'Success' | 'Warning' | 'New' | 'Neutral' | 'Information' | 'Error' = 'Neutral';
+
+  render() {
+    const hasIconSlot = hasSlot('icon', this.host);
+    return (
+      <Host
+        class={{
```
timrombergjakobsson, line 30: What do you think about optimizing?

```
// Current approach
class={{
  'tds-tag': true,
  [`tds-tag-${this.size.toLowerCase()}`]: true,
  [`tds-tag-${this.variant.toLowerCase()}`]: true,
}}

// Consider using a utility function like:
private getTagClasses() {
  return {
    'tds-tag': true,
    [`tds-tag-${this.size.toLowerCase()}`]: true,
    [`tds-tag-${this.variant.toLowerCase()}`]: true,
  };
}
```

ckrook:
> Quick question, should text prop be optional or required? As it is now it is optional so the component will only take space if content is provided.

Good catch! Other major design systems make it required. Therefore I think we should too

timrombergjakobsson:
> > Quick question, should text prop be optional or required? As it is now it is optional so the component will only take space if content is provided.
> 
> Good catch! Other major design systems make it required. Therefore I think we should too

I agree :)

ckrook:
> Looks nice! Great findings previously as well, nice to see it is addressed already!
> 
> Two things:
> 
> * Fix: small variant should have a different font size
> * Note: icon placement could probably be specified as a prefix slot, since that's how we defined it in design to future-proof the component (if/when it becomes interactive we will need a trailing slot as well)

Adressed in latest commit. Thanks for the finding 🙏

packages/core/src/components/tag/tag.scss
```diff
@@ -0,0 +1,61 @@
...
+    color: var(--tds-tag-information-color);
+  }
+
+  &.--error {
+    background-color: var(--tds-tag-error-background);
+    color: var(--tds-tag-error-color);
+  }
+}
+
+.tds-tag__content {
```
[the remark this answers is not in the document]
timrombergjakobsson: Maybe, but consistency is more valuable than following BEM if the rest of the codebase doesn't use it. Other components use the current pattern. Also the BEM syntax here is incorrect. It should be:
```
&--sm 
&--lg
```

not 

```
&.--sm
&.--lg
```
Right now the DOM generated looks like this. 

Image

Image

In comparison with .tl components that follows the BEM syntax:

Image

Image

Currently we do like this in components:

```
&.sm
&.lg 
```

I suggest we keep that until we have had the discussion that we want to go in another direction. What do you think?

augustinescania (approved):
For me, the code looks good and consistent. And the component is aligned with the figma design. Great work with this @ckrook, always fun with a new component for our users! 🚀 💯 

The **ONLY** minor thing I'd point out (not a blocker for approval) is the tag-vars.scss naming. Since we introduced design tokens and working with TRATON, we've agreed to avoid using -color in variable names in that project. It’s redundant in usage since it's already clear they're color values. For example:

`color: var(--tds-tag-success);`

– here it's obvious --tds-tag-success is a color."

But of course, this is something that we can change in the future when we introduce the tag component to the design tokens project 😄

Issue #1331, opened 2025-07-16
[Bug report]: Clear searched text under dropdown should not reset the whole component


simonaapostolscania:
### Requirements before reporting

- [x] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [x] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [x] I have followed the installation guide.

### Package versions

"@scania/tegel-react": "^1.32.0"

### Browser

Chrome

### Framework

React

### Version

"react": "^18.2.0"

### Code sandbox

<TdsDropdown
        id='someid'
        ref={viewsDropdownRef}
        placeholder='Välj'
        filter={true}
        defaultValue={defaultSelectedDashboard.dashboardReference}
        label='ÄndraVy'
        label-position='outside'
        modeVariant='primary'
        size='md'
        tdsAriaLabel='ÄndraVy'>
        {allDashboards.map((view) => (
          <TdsDropdownOption
            key={view.dashboardReference}
            value={view.dashboardReference}
          >
            <span>{view.title.text}</span>
          </TdsDropdownOption>
        ))}
      </TdsDropdown>

### Reproduction steps

1. Add dropdown component with filter property set to true and 5 options (option 1, option 2, ..., option 5)
2. Select one option in it ('option 3' for example)
3. click outside -> now you see 'option 3' in the dropdown's placeholder place
4. click on its placeholder (the search input becomes editable)
5. type one short text ('as' for example)
6. click 'X' button to clear the typed search text
7. click outside -> now you see the placeholder text instead of option 3 previously selected, but on refresh option 3 is still selected

### Screenshots / Screen recordings

recording:

https://www.awesomescreenshot.com/video/42093365?key=3666d6283cbe823fa1fe377c4a4a34f3

### Expected behaviour

Component tds dropdown with filter should not completely reset itself when we click on the 'X' (clear filter) button. Instead it should just clear the exact text the user typed in and show the previously selected option or the default selected value (not the placeholder). Also, the previously selected option before typing in search should remain selected (should have the check mark in the right side of the option as being selected)

[...]

PR #1434, opened 2025-09-11
fix(dropdown): only reset filter query on clear

Lunkan89:
## **Describe pull-request**  
Only reset filter query on clear, if filter is present selected value should not be effected

## **Issue Linking:**  
- **GitHub:** https://github.com/scania-digital-design-system/tegel/issues/1331

## **How to test**  
1. Go to amplify link and go to dropdown component
2. Select `multiselect` & `filter`
3. Make some selections and then type in filter, now clear that filter
4. Selected values should remain.
5. you can compare this flow and behavior with production to see that whenever you clear filter, the selected value/values also get cleared

## **Checklist before submission*
- [x] All existing tests pass

### DIFF
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -99,6 +99,8 @@ export class TdsDropdown {
 
   @State() private selectedOptions: string[] = [];
 
+  @State() filterQuery: string = '';
+
   private dropdownList: HTMLDivElement;
 
   private inputElement: HTMLInputElement;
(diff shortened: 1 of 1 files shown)


Lunkan89:
> The x button isn't visible when having options selected, but it should be (as in production), right?
> 
> Also, when pressing the x button, the selected options appear, and then disappear when stopping the pressing of the x button. This is a minor detail though that perhaps doesn't need to be fixed

I dont see that behavior in production?
This is specific fix for clearing only query when query is present, I believe the "flashing" of selected values is present in prod as well

PR #1453, opened 2025-09-19
feat(tegel-light): updated tl-header

augustinescania:
## **Describe pull-request**  
Introducing a tegel light header component.

## **Issue Linking:**  
 [CDEP-977](https://jira.scania.com/browse/CDEP-977)

## **How to test**  
1. Go to preview link and navigate to tegel light and the header component
2. Check all settings and dark/light mode
3. Check Traton/Scania
4. Check Docs to see how styles are implemented with classes
5. Compare to [Figma](https://www.figma.com/design/6osYTOfd4MgDq7LO6BCoUO/TRATON-Component-Library?node-id=26566-59506&p=f&m=dev)
6. Check the code and give your thoughts on the BEM Structure and naming convention (Important one!)
7. Test responsitivity of the component
8. Test the controls in storybook and give your thoughts on if they make sense or not!

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events


packages/core/src/tegel-light/components/tl-header/tl-header.stories.tsx
```diff
@@ -1,53 +1,151 @@
...
+      control: 'boolean',
+      name: 'Include title?',
+    },
+    includeHamburger: {
+      control: 'boolean',
+      name: 'Include hamburger?',
+    },
+    includeBentoGrid: {
+      control: 'boolean',
+      name: 'Include bento? (Grid)',
```
timrombergjakobsson, line 20: Just a reflection, but do we need to write "Include" and then "?" everywhere? Does no merely "Bento" suffice? Just thinking for other booleans we don't write that.

ckrook: I agree, I think it should be just Bento (Grid) etc. Not a question

augustinescania: Yes you are correct, i have now removed the question marks from all the options.

packages/core/src/tegel-light/components/tl-header/_dropdown.scss
```diff
@@ -0,0 +1,286 @@
...
+  }
+}
+
+.tl-header__dropdown {
+  @include headline-07;
+
+  height: 100%;
+  position: relative;
+
+  &--open {
```
timrombergjakobsson, line 46: I think here it would make more sense to not cascade them. As we are basically writing `.tl-header--open .tl-header__dropdown-wrapper { ... }`.  Should not wrapper have its own modifier? So kinda like:

```
.tl-header__dropdown-wrapper {
  /* base styles */
}

.tl-header__dropdown-wrapper--open {
  background-color: var(--header-basic-element-background-pressed);
  color: var(--header-nav-item-dropdown-opened);
}
```
?

augustinescania: Yes they should, must've missed it. Its fixed now :)

Lunkan89 (commented):
I noticed that `<ul>` has direct child `div` which is not valid html, so I checked all the attributes in tds amplify link and copied the code to chatGPT:

### ❌ Where it drifts
1. **Invalid list structure**
   - `<div>`s appear directly inside `<ul class="tl-header__component-list">` (e.g. `tl-header__title`, `tl-header__dropdown`).
   - Some menus contain `<li>` but aren’t wrapped in a `<ul>` or `<ol>`.

2. **Elements as wrappers vs items**
   - `tl-header__dropdown` behaves like a list item but isn’t an `<li>`.
   - Breaks semantic HTML, even though naming fits BEM.

3. **State modifiers without a11y**
   - Modifiers like `--open` and `--selected` are correct BEM usage, but you should also toggle `aria-expanded`, `aria-hidden`, or `hidden` attributes for accessibility.

4. **Deep element chains**
   - Names like `tl-header__dropdown-menu-user-image` are acceptable in BEM, but can get unwieldy. Consider splitting into smaller sub-blocks if it grows too much.

### 🔧 Suggested cleanup
- Ensure every child of `<ul>` is an `<li>`.
- Use `<ul>/<li>` consistently for menus.
- Mirror `--open`/`--selected` modifiers with `aria-*` attributes.
- Keep element naming cohesive; split into new blocks if responsibilities diverge.

augustinescania:
@Lunkan89 

Thanks for the detailed review 🙏

I’ve gone through your points and made sure to address the semantic issues.

**Regarding accessibility:**
You are correct that modifiers like --open / --selected should be paired with attributes like aria-expanded and aria-hidden. However, a11y is not yet implemented in Tegel Lite at all. Since TL is based on the TRATON Component library, and that UI Kit itself is not a11y compliant yet, we’re holding off on adding accessibility until that work is in place. Even in the HTML.
This will of course be revisited at a later stage when it's time to implement a11y.

**On the “deep element chains” point:**
I agree the naming can get long, but it still fits within BEM. If the user profile dropdown grows further, we can look into splitting it into a dedicated sub-block. You are of course free to come up with a name suggestion! 🙂

ckrook (changes requested):
I noticed that the focus outline is currently rendered outside the items, while in the design it’s placed inside the element. Have you tried adjusting it to render inside instead? I know this isn’t how we handle it in the production Tegel header, but I think we should address it here since it’s not aligned with the design. :) 

This seems to concern all header items 

tl-header

PR #1458, opened 2025-09-22
feat(card): added expandable prop

augustinescania:
## **Describe pull-request**  
The Card component has a design variant that supports an expandable behaviour. This allows users to hide the card content under a dropdown-like interaction, toggled via a chevron icon in the header.

Until now, this functionality existed in design (Figma) but was missing in code. This PR adds that functionality

(I also added two action buttons for demo purpose that alligns with the current figma design)

## **Issue Linking:**  
- **Jira:** `[CDEP-1522](https://jira.scania.com/browse/CDEP-1522)

## **How to test**  
1. Go to preview link and open up the card component
2. Set the expandable prop to true
3. Click on the chevron that appears on the component and test the expand functionality.
4. Try the expanded prop via storybook control
5. Try having different options on at the same time (eg. body image, stretched etc..)
6. Compare to figma design [here](https://www.figma.com/design/d8bTgEx7h694MSesi2CTLF/Tegel-UI-Library?node-id=26478-59506&p=f&m=dev)

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  

### DIFF
--- a/packages/core/src/components/button/readme.md
+++ b/packages/core/src/components/button/readme.md
@@ -30,6 +30,19 @@
 | `"label"` | Slot for the text injection. Serves as alternative to text prop. |
 
 
+## Dependencies
+
+### Used by
+
+ - [tds-card](../card)
+
+### Graph
+```mermaid
+graph TD;
+  tds-card --> tds-button
+  style tds-button fill:#f9f,stroke:#333,stroke-width:4px
+```
+
 ----------------------------------------------
 
 *Built with [StencilJS](https://stenciljs.com/)*
--- a/packages/core/src/components/card/card.scss
+++ b/packages/core/src/components/card/card.scss
@@ -147,6 +147,6 @@ button {
   }
 }
 
-:host([expandable]:not([expanded])) .card-body slot {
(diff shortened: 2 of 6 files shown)


augustinescania:
> When the expandable content is closed it affects the size of the entire card component. I think expanding the content shouldn't affect the width of the card, only show/hide content when looking at the figma design.

Very good comment @NordbergAnna. That is because of the nature of how the card component works.

The component doesn't really have a set width or height, it is a modular component. Here is what the documentation say from tegel.scania.com:

image

A practical example that you can try to verify this is following:

1. Open up the card component in preview link
2. Right click on it and inspect the element.
3. There is a wrapper class called "demo-wrapper", click on it and set a custom width to 368px(That is the width of the card component when there is body content inside). Look at the screenshot below
4.  Now try the expand functionality again and you will see that the width won't change.

Skärmavbild 2025-09-22 kl  16 35 20



(Plus; If you also activate the body image, then it will look like the figma design and you can test it that way instead🙂 )

PR #1485, opened 2025-10-03
Fix/tl card

patrik-nyfeldt:
## **Describe pull-request**  

## **Issue Linking:**  

## **How to test**  
1. Go to...
2. Check in...
3. Run ...

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  

## **Additional context**  

### DIFF
--- a/packages/core/src/tegel-light/components/tl-card/tl-card.scss
+++ b/packages/core/src/tegel-light/components/tl-card/tl-card.scss
@@ -154,6 +154,7 @@
   align-items: flex-end;
   flex: 1 0 0;
   align-self: stretch;
+  gap: 16px;
 }
 
 .tl-card__stretch {
--- a/packages/core/src/tegel-light/components/tl-card/tl-card.scss
+++ b/packages/core/src/tegel-light/components/tl-card/tl-card.scss
@@ -1,5 +1,5 @@
 @import '../../../mixins/box-sizing';
-@import './_tl-card-vars';
+@import './tl-card-vars';
 @import '../../../../../../typography/utilities/typography-utility';
 
 .tl-card {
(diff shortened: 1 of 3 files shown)


packages/core/src/tegel-light/components/tl-card/tl-card.scss
```diff
@@ -1,5 +1,5 @@
 @import '../../../mixins/box-sizing';
-@import './_tl-card-vars';
+@import './tl-card-vars';
```
nathalielindqvist, line 2: I believe this file follows the convention of having an underscore before the file name, like all the other vars files we have for tl. Is there a reason to rename it?

packages/core/src/tegel-light/components/tl-card/tl-card.scss
```diff
@@ -160,6 +154,7 @@
   align-items: flex-end;
   flex: 1 0 0;
   align-self: stretch;
+  gap: 16px;
```
nathalielindqvist, line 157: The 'actions' part of the component, at least in the webcomponent version in Tegel, is meant as a slot where the user can slot in anything they want. We have slotted in buttons as an example, but it can be links, icons or other element. I think that might be why the gap was hard-coded - if we have it set in our CSS rules it might create unwanted space if the user decides to slot in something else than two buttons.

packages/core/src/tegel-light/components/tl-card/tl-card.scss
nathalielindqvist: @NordbergAnna @timrombergjakobsson do you have any input on this? I was unfortunately not present when the file structure was set, so I'm not sure about the reasoning around the underscore

timrombergjakobsson: Like Patrik said from a technical standpoint it is not needed. I would say best practice would be, use underscores in partial filenames, but omit them in imports. But then it needs to be consistent. So maybe then better to do it as one separate PR, where all are changed.

NordbergAnna: I agree with @timrombergjakobsson and @patrik-nyfeldt, let's go through files and remove from imports, I can do that in the fix/clean-up branch 👍

PR #1507, opened 2025-10-17
fix(build): ts strict mode

Lunkan89:
## **Describe pull-request**  
Started on using strict mode

## **Issue Linking:**  
- **Jira:** [CDEP-1394](https://jira.scania.com/browse/CDEP-1394)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,24 +42,24 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.26.2",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.26.2.tgz",
-      "integrity": "sha512-RJlIHRueQgwWitWgF8OdFYGZX328Ax5BCemNGlqHfplnRT9ESi8JkFlvaVYbS+UubVY6dpv87Fs2u5M29iNFVQ==",
+      "version": "7.27.1",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.27.1.tgz",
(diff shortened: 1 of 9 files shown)


packages/core/src/types/Attributes.ts
```diff
@@ -1 +1,2 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 export type Attributes = { [key: string]: any };
```
ckrook: @Lunkan89 Maybe we want to avoid `any` here?

packages/core/src/components/checkbox/checkbox.tsx
```diff
@@ -143,7 +143,9 @@ export class TdsCheckbox {
       <div class="tds-checkbox">
         <input
           // eslint-disable-next-line no-return-assign
```
ckrook: We can remove this now
```suggestion
          
```

tsconfig.json
```diff
@@ -15,4 +43,9 @@
     "jsx": "react",
     "jsxFactory": "h",
   },
+  "exclude": ["packages/core/src/components/**/test/*", "**/*.d.ts", "**/*.stories.*", "**/*.axe.ts", "**/*.e2e.ts", "packages/core/src/stories/**", "packages/core/src/components/_beta/**"],
+  "include": ["packages/core/**/*"],
+  "files": [
+    "packages/core/src/global.d.ts"
+  ]
```
lkritsimas: I don't think stories and test files should be excluded from using strict mode. Ideally it should apply to all TS files in the whole repository.

PR #1527, opened 2025-11-05
fix(plausible): remove data-api attribute to fix Plausible CORS error

timrombergjakobsson:
## **Describe pull-request**  
Removed the data-api attribute. The script will use Plausible's default endpoint handling, which avoids CORS issues.
What this fixes:

- Removes the CORS error by removing the custom API endpoint override
- Keeps tracking working — the manual script still sends events through Plausible's infrastructure
- Your custom pageview tracking code continues to work

Why this works:
The data-api attribute was forcing direct fetch calls to https://plausible.io/api/event, which triggers CORS. Without it, the Plausible script handles API calls through its own infrastructure, avoiding CORS.
The tracking will continue to work — pageviews and custom events will still be sent to Plausible, just without the CORS error.

## **How to test**  
Have a look at the preview link and compare with current Storybook to verify that it is fixed.

### DIFF
--- a/packages/core/.storybook/manager-head.html
+++ b/packages/core/.storybook/manager-head.html
@@ -1,75 +1,91 @@
 <!-- .storybook/manager-head.html -->
 
-<script
-  defer
-  data-domain="tds-storybook.tegel.scania.com"
-  src="https://plausible.io/js/script.manual.js"
-></script>
-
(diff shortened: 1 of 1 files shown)


augustinescania (approved):
I am still not getting any errors. @timrombergjakobsson

PR #1559, opened 2025-11-24
fix(tegel-lite): fixed build error and removed behaviour from tl-popover-canvas

augustinescania:
## **Describe pull-request**  
This PR fixes a missing import in the tl-footer component that caused the `npm run build:tegel-light` command to fail when building Tegel Lite.

It also removes the escape-to-close and click-outside-to-close behaviours from the tl-popover-canvas Storybook demo.
The reason for this change is to ensure consistent behaviour with tl-popover-menu, where we have chosen to remove these behaviours as well.


nathalielindqvist (approved):
:shipit:

PR #1624, opened 2026-01-07
test(tds-footer): update screenshot

augustinescania:
## **Describe pull-request**  
Recently, the tds footer component keeps failing on our tests. The issue is the dynamic year in the copyright text in the test file of the component. Since it's a new year, it now prints 2026 when testing and it doesnt match with the old 2025, hence why screenshot was failing when compared to eachother


timrombergjakobsson (approved):
🫅

PR #1632, opened 2026-01-13
refactor(tl-block): automatic alternating backgrounds and typography

EmelieLitwin:
## Describe pull-request
Refactors tl-block to use automatic alternating backgrounds and typography based on nesting level, removing the need for manual --even, --odd, --nested, and --nested-inner modifiers.

Current tl-block requires users to manually apply modifiers which is error-prone and inconsistent with tds-block web component. This PR implements automatic CSS-based alternation using descendant selectors, supporting up to 4 nesting levels.

## Issue Linking:
**Jira:** [CDEP-1875](https://jira.scania.com/browse/CDEP-1875)

## How to test
1. Go to Storybook → Tegel Lite (CSS) → Block → Nested
2. Check that nested blocks automatically alternate backgrounds without --even/--odd classes
3. Verify typography automatically adjusts (h2 → h3 → h4) without --nested/--nested-inner modifiers
4. Test with both --primary and --secondary parent modes
5. Verify up to 4 nesting levels work correctly

### DIFF
--- a/packages/core/src/tegel-lite/components/tl-block/tl-block-nested.stories.tsx
+++ b/packages/core/src/tegel-lite/components/tl-block/tl-block-nested.stories.tsx
@@ -34,22 +34,16 @@ const Template = ({ outerModeVariant, componentTag }) =>
       "@scania/tegel-lite/tl-block.css"
     -->
     <div class="tl-block tl-block--${outerModeVariant.toLowerCase()}">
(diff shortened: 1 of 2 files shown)


NordbergAnna (approved):
A lot easier for the user to only add one class and code looks so much lighter, nice work! 🌟

PR #1633, opened 2026-01-13
refactor(tl-breadcrumbs): simplify API by removing BEM classes

EmelieLitwin:
## Describe pull-request
Simplifies tl-breadcrumbs API by removing verbose BEM classes. Users now only need to apply the tl-breadcrumbs class to the nav element, with standard HTML elements (ol, li, a) inside. Current page detection is handled automatically via the aria-current attribute.

## Issue Linking:
**Jira:** [CDEP-1876](https://jira.scania.com/browse/CDEP-1876)

## How to test
1. Go to Storybook → Tegel Lite (CSS) → Breadcrumbs
2. Check that breadcrumbs render correctly with simplified HTML structure
3. Verify separators (›) appear between items automatically
4. Verify last item (with aria-current="page") has no separator
5. Test hover and focus states on links
6. Verify current page link is disabled and styled differently

### DIFF
--- a/packages/core/src/tegel-lite/components/tl-breadcrumbs/tl-breadcrumbs.scss
+++ b/packages/core/src/tegel-lite/components/tl-breadcrumbs/tl-breadcrumbs.scss
@@ -5,62 +5,62 @@
 
 .tl-breadcrumbs {
   @include tds-box-sizing;
-}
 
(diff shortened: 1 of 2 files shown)


NordbergAnna (approved):
LGTM, nice work! 🥳

PR #1640, opened 2026-01-15
fix(tl-card): update conditions of expandable settings

NordbergAnna:
## **Describe pull-request**  
This PR fixes conditions of the storybook settings to adress an issue with image positioned below and expandable being true at the same time. According to figma the expandable prop should only work when image is positioned **above** . This also caused the need to update the "clickable" condition. Clickable should never be true at the same time as expandable is true.

## **Issue Linking:**  
[CDEP-1863](https://jira.scania.com/browse/CDEP-1863)

## **How to test** 
1. Go to the preview link and navigate to the Tegel Lite Card component. 
2. Set the Body Image to true and Body Image placement to "Below"
3. Check that the Expandable setting is not visible
4. Chang Body Image placement to "Above"
5. Check that the Expandable setting is visible
8. When Expandable is true, the setting for Clickable should not be visible. Change Expandable to False and check that you now have a setting for Clickable.

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

### DIFF
--- a/packages/core/src/tegel-lite/components/tl-card/tl-card.stories.tsx
+++ b/packages/core/src/tegel-lite/components/tl-card/tl-card.stories.tsx
@@ -131,14 +131,14 @@ export default {
       name: 'Expandable',
       description: 'Toggles if the Card can expand/collapse its content.',
       control: 'boolean',
-      if: { arg: 'clickable', eq: false },
+      if: { arg: 'imagePlacement', neq: 'Below' },
       table: { defaultValue: { summary: false } },
     },
     expanded: {
       name: 'Expanded',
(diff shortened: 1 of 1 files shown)


nathalielindqvist (commented):
Looks like the Card changes size when expandable is set to true, is this intentional?

https://github.com/user-attachments/assets/a43692b5-071d-4e0d-9961-546e2340feec

NordbergAnna:
> Looks like the Card changes size when expandable is set to true, is this intentional?
> 

Same behaviour as web component so I would say okay for now at least? 😊

PR #1681, opened 2026-01-28
fix: correct chevron positioning in tl-dropdown select variant

EmelieLitwin:
## **Describe pull-request**  
Fixes misaligned chevron icon in tl-dropdown select variant across different label and helper text configurations. The chevron now properly aligns vertically with the input field for all combinations of label positions and helper text.

## **How to test**  
1. Go to Preview link → Tegel Lite → Dropdown → Select variant
2. Check chevron alignment in all combinations:
   - Label positions: Outside, Inside, None
   - Helper text: true/false
   - Sizes: Large, Medium, Small
3. Verify chevron is vertically centered in the input field for each configuration

### DIFF
--- a/packages/core/src/tegel-lite/components/tl-dropdown/tl-dropdown.scss
+++ b/packages/core/src/tegel-lite/components/tl-dropdown/tl-dropdown.scss
@@ -543,6 +543,10 @@
   transition: transform 0.2s cubic-bezier(0.4, 0, 0.2, 1), visibility 0s 0.2s;
   @include tl-scrollbar;
 
+  .tds-mode-dark & {
+    box-shadow: none;
+  }
+
   .tl-dropdown--dropup & {
     top: auto;
     bottom: 100%;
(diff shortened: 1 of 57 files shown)


packages/core/src/tegel-lite/components/tl-dropdown/tl-dropdown.scss
```diff
@@ -158,17 +163,12 @@
...
+.tl-dropdown__input-wrapper:has(.tl-dropdown__input[aria-expanded='true']:not(:placeholder-shown))::before {
   opacity: 1;
 }
 
 .tl-dropdown__button::after,
-.tl-dropdown__input-wrapper::after,
-.tl-dropdown:has(.tl-dropdown__select):not(:has(.tl-dropdown__button)):not(
-  :has(.tl-dropdown__input)
-)::after {
+.tl-dropdown__input-wrapper::after {
```
nathalielindqvist: I looks like the chevron still has a specific rule that's being applied when the label is outside and the helper text is set to false, which makes it a tiny bit off-center.

Image

augustinescania (approved):
LGTM!

PR #1730, opened 2026-02-19
feat(tds-table): added colspan and rowspan prop

antonstrandqvist:
## **Describe pull-request**
Adds `col-span` and `row-span` props to `tds-body-cell` and `tds-header-cell`, allowing cells to span multiple columns and/or rows in the table layout.

## **Issue Linking:**
- **No issue:** The table component lacked native support for cell spanning, which is a common requirement for complex table layouts (e.g. merged header groups, summary rows).

## **How to test**
1. Open Storybook and navigate to **Components/Table/Col and Row Span**
2. Use the Storybook controls to set **Header cell col-span**, **Body cell col-span**, and **Body cell row-span**
3. Verify cells visually span the correct number of columns/rows
4. Confirm that cells without these props are unaffected

Alternatively, use the HTML attributes directly:
- `<tds-header-cell col-span="2">` — header cell spans 2 columns
- `<tds-body-cell col-span="2">` — body cell spans 2 columns
- `<tds-body-cell row-span="2">` — body cell spans 2 rows

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [x] Storybook controls
- [ ] Events

## **Screenshots**
_Include before/after screenshots showing a table with and without colspan/rowspan applied._

### Implementation approach
Both `tds-header-cell` and `tds-body-cell` now render their actual `<th>`/`<td>` elements inside shadow DOM, with `display: contents` on the host element. This means the shadow `<th>`/`<td>` participates directly in the parent `<tr>`'s table layout, so the browser's table algorithm correctly processes `colspan` and `rowspan`.

Previously, the host element itself was styled as `display: table-cell`, but the browser's table layout algorithm only processes `colspan`/`rowspan` on native `<th>`/`<td>` elements — not on custom elements, even when styled as `display: table-cell`.

### Changes
- **`tds-header-cell`**: renders `<th colSpan rowSpan>` in shadow DOM; host uses `display: contents`
- **`tds-body-cell`**: renders `<td colSpan rowSpan>` in shadow DOM; host uses `display: contents`
- **`tds-table-header`** and **`tds-table-body-row`**: unchanged — still use a simple `<slot>` for cell projection
- **Storybook**: new story at _Components/Table/Col and Row Span_ with interactive controls
- **E2E tests**: added `test/colspan-rowspan/` fixture and test suite; updated `test/default/` locators from `tds-body-cell` to `tds-body-cell td` to pierce shadow DOM for visibility assertions

### Breaking changes
None. The `display: contents` approach does not change the rendered visual output for existing tables. The props `col-span` and `row-span` are optional and default to `undefined`, so all existing usage is unaffected.


ckrook:
When I'm testing this in React demo page I'm not able to make this work.
Here is the test PR where I attempt this.
https://github.com/scania-digital-design-system/tegel-react-demo/pull/281

Screenshot 2026-02-23 at 13 25 59

timrombergjakobsson:
> When I'm testing this in React demo page I'm not able to make this work. Here is the test PR where I attempt this. [scania-digital-design-system/tegel-react-demo#281](https://github.com/scania-digital-design-system/tegel-react-demo/pull/281)
> 
> Screenshot 2026-02-23 at 13 25 59

@ckrook I saw that you closed the PR, was something wrong?

antonstrandqvist:
@timrombergjakobsson @ckrook  yep, needed to fix some things...

packages/core/src/components/table/table-body-cell/table-body-cell.tsx
```diff
@@ -127,11 +133,11 @@ export class TdsTableBodyCell {
           'tds-table--divider': this.verticalDividers,
           'tds-table--no-min-width': this.noMinWidth,
         }}
-        style={dynamicStyles}
-        role="cell"
```
ckrook, line 131: Do we really want to remove the role here from an a11y perspective?

packages/core/src/components/table/table-header-cell/table-header-cell.tsx
```diff
@@ -280,13 +286,19 @@ export class TdsTableHeaderCell {
           'tds-table--toolbar-available': this.enableToolbarDesign,
           'tds-table--no-padding': this.disablePadding,
         }}
-        style={{ minWidth: this.customWidth }}
-        onMouseOver={() => this.onHeadCellHover(this.cellKey)}
-        onMouseLeave={() => this.onHeadCellHover('')}
-        role="columnheader"
```
ckrook, line 286: sam here, do we rly want to remove it?

ckrook: Perhaps we can jump in a call and talk this through :)

antonstrandqvist: Its because we are rendering a actuall th, so there is no need to set a role ;)

timrombergjakobsson: Yeah a native `<th`> already has an implicit ARIA role

ckrook (approved):
I think this looks good. One thing I was thinking is that perhaps we want to validate that the value is a positive integer but I don’t see any scenario where allowing negative values would currently break the component. (tested in Storybook)

patrik-nyfeldt (approved):

PR #1737, opened 2026-02-24
fix(tds-datetime): min max validation

mfilipadneto:
## **Describe pull-request**  
This PR fixes the validation errors reported on the CDEP-1752 for the datetime component.

We have implemented the validation of the user's input with the [validity of the HTMLInputElement API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/validity).

We have created a new _OPTIONAL_ prop to the datetime component to hold the error message. The user may want to specify the helper text for nominal usage of the component and set a different error message for the error state.

We have created the `iphone` css class to be applied for the error and success validation.

We have merged the error css for `md` and `sm` cases as they were the same.

We have added an extra validation case for invalid dates (for example, 31-02-2026 or 31-11-2026), as the error message should be specific to this case and not the generic error message specified by the user.

We have not done any snapshot tests for these changes since we drove the tests according to the user behaviour. 
We kept one test that renders the component in the error state to ensure that the graphic elements are all present (error icon and message in red (with the dark and light nuances)).

One of the acceptance criteria for this PR was that the `Date picker should respect min/max constraints on all platforms`. Due to [limitations of the native input type="date" element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/date#browser_compatibility) on iOS it is not possible to limit the range of the values shown on the date picker for this platform. However, the error message is displayed informing the user of their bad input.

## **Issue Linking:**  
- **Jira:** [CDEP-1752](https://jira.scania.com/browse/CDEP-1752)

## **How to test**  
1. Open Storybook and navigate to Components/Datetime/Default
2. Set the type to `date`, for example.
3. Set a `min` and `max` values for the date
4. Manually add a date that is outside of the limits set in 3.
5. Remove focus from the input element
6. Watch the error helper appear. If you want, in step 2 configure the helper error text to confirm that it renders your intended message 
7. Cycle through the other input types available.

In alternative, you can also checkout the branch and run the tests in docker. 
⚠️  if you run the tests locally with `npx playwright test --ui` the tests for the input type="date" might fail due to different date formats in different browsers. In docker, which is the reference environment, the format is "MM/dd/yyyy" and in our local environment it was "dd/MM/yyyy".

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
- the screenshots for the form that uses the datetime input has also been updated

## **Additional context**  
n/a


packages/core/src/components/datetime/datetime.tsx
```diff
@@ -201,13 +223,18 @@ export class TdsDatetime {
   }
 
   render() {
```
timrombergjakobsson, line 263: Maybe we can add a comment about this:
```
// iOS native date picker doesn't support min/max in the UI; we rely on validation + error message.
// The .iphone class enables iOS-specific styling for error/success states (see CDEP-1752).
```
or something similar? As this check is tied to a real platform quirk, not just “we want different styling on phones.”

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -168,6 +171,22 @@ export class TdsDatetime {
     this.focusInput = false;
   }
 
+  static parseDate = (stringDate: string) => {
+    const date = new Date(Date.parse(stringDate));
+    return date;
+  };
+
+  private validateDate = () => {
+    this.state = 'success';
```
timrombergjakobsson: Will this work?  `state` is a `@Prop()` not state, so mutating it might not work. In Stencil props are read-only from inside the component.

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -168,6 +171,22 @@ export class TdsDatetime {
     this.focusInput = false;
   }
 
+  static parseDate = (stringDate: string) => {
+    const date = new Date(Date.parse(stringDate));
```
timrombergjakobsson: I believe `parseDate` is redundant here. 
`const date = new Date(stringDate);` suffice.

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -42,8 +42,8 @@ export class TdsDatetime {
   /** Name property. Uses a unique ID as fallback if not specified. */
   @Prop() name = `datetime-${generateUniqueId()}`;
 
-  /** Error state of input */
-  @Prop() state?: string;
+  /** Switches between success and error state. */
+  @Prop() state?: 'none' | 'success' | 'error';
```
timrombergjakobsson: I believe either make it mutable or convert to internal state

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -281,13 +306,24 @@ export class TdsDatetime {
...
-            </div>
+            {(!this.textInput || !this.textInput.validity.badInput) && (
+              <div class="tds-helper">
+                <tds-icon name="error" size="16px" svgTitle="error" /> {this.helperError}
+              </div>
+            )}
+            {this.textInput && this.textInput.validity.badInput && (
+              <div class="tds-helper">
+                <tds-icon name="error" size="16px" svgTitle="error" />{' '}
+                {this.type === 'time' ? 'Invalid Time' : 'Invalid Date'}
```
ckrook: We try to refrain from populate content in the DS.
The reason to this is that this DS could be used in non-english speaking countries.

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -168,6 +171,22 @@ export class TdsDatetime {
     this.focusInput = false;
   }
 
+  static parseDate = (stringDate: string) => {
```
ckrook: Is this used anywhere?

packages/core/src/components/datetime/datetime.tsx
ckrook, line 136: If a user choose to programmatically set the value with this method, `validateDate()` will not be called.
out-of-range values will only show error when blur. Which I think is an issue.

packages/core/src/components/datetime/datetime.tsx
ckrook: I agree with Tim. But I also see an issue with setting the state to success before validating the value

packages/core/src/components/datetime/datetime.tsx
mfilipadneto: Makes sense, yes! We will add it :)

packages/core/src/components/datetime/datetime.tsx
mfilipadneto: Oooh ok. The thing is, we need to control the `state` somehow to be able to validate and show the error when appropriate. We tested it and it seems to be working... 🤔 
We propose to change the `state` variable to `Prop({ mutable: true })` as per the [docs](https://stenciljs.com/docs/properties#prop-mutability-mutable) instead of creating a @/State variable for the same purpose as the state prop.

packages/core/src/components/datetime/datetime.tsx
mfilipadneto: True! Thanks for the heads up 👍

packages/core/src/components/datetime/datetime.tsx
mfilipadneto: Making it mutable sounds like a more viable option, as we said in the previous comment, hehe. Thanks Tim!

packages/core/src/components/datetime/datetime.tsx
mfilipadneto: You're totally right, that makes a lot of sense. We distinguished this "Invalid Date" situation from the "normal" error because the root cause is different (outside min/max VERSUS invalid date).
Would it make sense to create an optional Prop specifically dedicated to the error message when the error is an Invalid Date? 
Something like the `helperError` but with the default value being the english one (as it is the output of the Date API when it parses an invalid date). Then the user would have the possibility to specify the message in their language. Something in the lines of:
`@Prop() helperErrorInvalid?: string = 'Invalid input';`

packages/core/src/components/datetime/datetime.tsx
mfilipadneto: Oopsie, in fact it is not 😅 we forgot to delete it from ancient times where we implemented some manual validations lol. Thanks Charles!

packages/core/src/components/datetime/datetime.tsx
mfilipadneto: It's basically a "reset". If it's already at state = "error", and a new input is received, it should "reset" so that the validation can go through. Maybe it should be set to 'none' instead of success though 😅

mfilipadneto:
@timrombergjakobsson  and @ckrook , we have implemented your comments and suggestions. 
Please take a look 😄

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -168,6 +193,25 @@ export class TdsDatetime {
...
+  validateDate(): void {
+    this.state = 'none';
+    if (
+      (this.min && this.textInput.validity.rangeUnderflow) ||
+      (this.max && this.textInput.validity.rangeOverflow) ||
+      this.textInput.validity.badInput
+    ) {
+      this.state = 'error';
+    } else {
+      this.state = 'success';
```
ckrook, line 224: Instead of forcing a success state I think we should be more careful here
For example imagine the user has two datetime components like so:

<tds-datetime id="start" type="date" />
<tds-datetime id="end" type="date" />

The user might want to pass down a error if the end date overlaps the start date.

Currently we forces success state even though the user might not agree. 

I think we need to rethink how we set the error on the state

```
private validationErrorActive: boolean = false;

validateDate(): void {
  const hasError = 
    (this.min && this.textInput.validity.rangeUnderflow) ||
    (this.max && this.textInput.validity.rangeOverflow) ||
    this.textInput.validity.badInput;

  if (hasError) {
    this.validationErrorActive = true;
    this.state = 'error';
  } else if (this.validationErrorActive) {
    // only clear if WE set the error — don't touch externally-set state
    this.validationErrorActive = false;
    this.state = 'none';
  }
}
```

packages/core/src/components/datetime/test/error/datetime.e2e.ts
```diff
@@ -16,76 +16,128 @@ testConfigurations.withModeVariants.forEach((config) => {
       await setupPage(page, config, componentTestPath, componentName);
     });
 
-    test('renders disabled datetime component correctly', async ({ page }) => {
-      /* Check diff on screenshot */
-      await expect(page).toHaveScreenshot({ maxDiffPixels: 0 });
+    test('renders error helper text in error state with red color and error icon', async ({
+      page,
+    }) => {
+      const helperText = "This is for status='error'";
```
ckrook:
```suggestion
      const helperText = "This is for state='error'";
```

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -200,14 +245,21 @@ export class TdsDatetime {
     this.value = value;
   }
 
+  // iOS native date picker doesn't support min/max in the UI; we rely on validation + error message.
+  // The .iphone class enables iOS-specific styling for error/success states.
   render() {
+    const iphone = navigator.userAgent.toLowerCase().includes('iphone');
```
ckrook, line 264: We do not need to run this on every render, I think its sufficient to put this inside the `componentWillLoad()` lifecycle

packages/core/src/components/datetime/test/error/datetime.e2e.ts
```diff
@@ -16,76 +16,128 @@ testConfigurations.withModeVariants.forEach((config) => {
...
+
+      await expect(page.getByText(errorText)).toBeVisible();
+      await expect(page.getByText(helperText)).not.toBeVisible();
+    });
+
+    test('renders the error message when manually selecting a time outside the min/max boundaries', async ({
+      page,
+    }) => {
+      const errorText = 'Value must be between 08:30 and 17:30';
+      const helperText = "Please enter a week with format 'HH:mm'";
```
ckrook: Perhaps a misstype. Shouldnt the helpertext be:  `Please enter a week with format 'HH:mm'`

Instead perhps:
`Please enter a time with format 'HH:mm'`

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -42,8 +55,8 @@ export class TdsDatetime {
   /** Name property. Uses a unique ID as fallback if not specified. */
   @Prop() name = `datetime-${generateUniqueId()}`;
 
-  /** Error state of input */
-  @Prop() state?: string;
+  /** Switches between success and error state. */
+  @Prop({ mutable: true }) state?: 'none' | 'success' | 'error';
```
timrombergjakobsson, line 59: I'm thinking a more clean design is to make state fully controlled by the consumer. 

`@Prop() state: 'none' | 'success' | 'error' = 'none';`

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -168,6 +193,25 @@ export class TdsDatetime {
     this.focusInput = false;
   }
 
+  // Decorator to activate validation when input value changes
+  @Watch('value')
+  onValueChanged() {
+    this.shouldValidate = true;
+  }
+
+  validateDate(): void {
```
timrombergjakobsson, line 215: Maybe we can hve the component compute validity and emit an event (tdsValidityChange) with valid + reason, but not set state itself?

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -168,6 +193,25 @@ export class TdsDatetime {
...
+    this.state = 'none';
```
timrombergjakobsson, line 216: So like mentioned, we remove any `this.state = ...` assignments from the component.

packages/core/src/components/datetime/datetime.tsx
timrombergjakobsson: I would also add an internal flag based on native validity and customValidator. 
`@State() internalError: boolean = false;`

packages/core/src/components/datetime/datetime.tsx
timrombergjakobsson: I would rather change this one to:

```
private validateNative(): void {
    if (!this.textInput) return;

    const validity = this.textInput.validity;
    const nativeError =
      (this.min && validity.rangeUnderflow) ||
      (this.max && validity.rangeOverflow) ||
      validity.badInput;
    const customError =
      this.customValidator && !this.customValidator(this.textInput.value);

    this.internalError = !!(nativeError || customError);
  }
```

And remove `componentDidRender()`

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -168,6 +204,24 @@ export class TdsDatetime {
...
```
ckrook, line 216: [Please have a look at MUI design system](https://mui.com/x/react-date-pickers/validation/)
If the user enters a valid date and then clear the input field I dont think it should be treated as an error unless the input field is required. (which we dont support currently)


So I would suggest doing:

```suggestion
    if (!this.textInput.value) {
      this.state = 'none';
      return;
    }
```

ritamcastro: Seems like this breaks the behavior when inputting an invalid date (2026-02-31)... Because the value of the textInput is not populated when the input is a bad one. 

image

What do you think of also calling `this.validateDate()` on the `handleFocusOut` function, while adding an else if in the validateDate()? 

```
 validateDate(): void {
    this.state = 'none';
    
    if (
      !this.nativeValidation() ||
      (this.customValidator && !this.customValidator(this.textInput.value))
    ) {
      this.state = 'error';
    } else if(this.textInput.value) {
      this.state = 'success';
    }
  }
```

This way, if a user clears the input using the keyboard, the `validateDate` will be run when clicking outside.

ckrook: Thats a good suggestion!

mfilipadneto:
We had to create a [new PR](https://github.com/scania-digital-design-system/tegel/pull/1795) because of some conflict resolution issues in this old branch.

Issue #1720, opened 2026-02-13
[Bug report]: InlineTabs | Active tab underline disappears after switching between tabs


patrikhultgren:
### Requirements before reporting

- [x] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [x] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [x] I have followed the installation guide.

### Package versions

@scania/tegel 1.47.2

### Browser

Chrome

### Framework

React

### Version

React 19

### Code sandbox

https://stackblitz.com/edit/vitejs-vite-6w2hcnng?file=src%2FApp.tsx

<img width="484" height="494" alt="Image" src="https://github.com/user-attachments/assets/4c45187b-c0cb-4d07-a6e5-07808bbde1f7" />

(Couldn't paste code with proper formatting...)

### Reproduction steps

1. Click Tab 1
2. Click Tab 2
3. Click Tab 1

The underline is missing on Tab 1.

### Screenshots / Screen recordings

Step 1

<img width="183" height="139" alt="Image" src="https://github.com/user-attachments/assets/2a1f8080-2e86-4e55-ab0a-24407b3fd249" />

Step 2

<img width="162" height="123" alt="Image" src="https://github.com/user-attachments/assets/095c5364-08b4-4418-a288-f6c52c1813d7" />

Step 3 
(shows the missing underline of Tab 1)

<img width="169" height="125" alt="Image" src="https://github.com/user-attachments/assets/9f6215e2-aaa1-460f-b50c-8233206edd41" />

### Expected behaviour

The underline should be displayed beneath Tab 1.

### Contact information

patrik.hultgren@scania.com

ckrook:
@patrikhultgren Thanks for reporting this. I have copied the issue to Jira!

ckrook:
@patrikhultgren we have released a fixed for this in last release

PR #1741, opened 2026-02-25
fix(inline-tabs): restore active indicator when switching back to first tab

ckrook:
## **Describe pull-request**
The active tab underline disappeared when navigating back to the first tab because handleSelectedIndexUpdate used a truthy check (if (this.selectedIndex)) that evaluates to false when selectedIndex is 0. The condition is now if (this.selectedIndex !== undefined) so tab[0] is correctly re-selected. A regression test covering the Tab1 → Tab2 → Tab1 sequence is included.

## **Issue Linking**
Jira: [CDEP-1994](https://jira.scania.com/browse/CDEP-1994)
Github: [Issue-1720](https://github.com/scania-digital-design-system/tegel/issues/1720)

## **How to test**
Created a betarelease with my fix which can be tested in this code sandbox
https://stackblitz.com/edit/vitejs-vite-nrmnprhw?file=package.json

### DIFF
--- a/packages/core/src/components/tabs/inline-tabs/inline-tabs.tsx
+++ b/packages/core/src/components/tabs/inline-tabs/inline-tabs.tsx
@@ -103,7 +103,7 @@ export class TdsInlineTabs {
       return tabElement;
     });
 
-    if (this.selectedIndex) {
+    if (this.selectedIndex !== undefined) {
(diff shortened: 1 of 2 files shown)


antonstrandqvist (approved):
Looks great!

patrik-nyfeldt (approved):

PR #1744, opened 2026-02-25
fix(tds-dropdown): persist labels in multiselect + filter

adelezili:
## **Describe pull-request**  
This PR improves the `tds-dropdown` component by addressing label persistence issues and aligning buttons implementation across all modes.

* In `multiselect + filter` mode, selected labels now persist when toggling the dropdown menu.
* In `multiselect` mode, added clear button (consistent with filter mode behavior).
* Replaced clickable tds-icon (role="button") with semantic <button> elements for improved accessibility and consistency with other components.
* Changed `:focus` to `:focus-visible` for keyboard-only focus styling.
* Updated e2e tests to align with semantic button structure and regenerated baseline snapshots.

## **Issue Linking:**  
- **Jira:** [CDEP-1764](https://jira.scania.com/browse/CDEP-1764)

## **How to test**  
1. Go to Storybook and open `Dropdown` component.
2. Enable `multiselect`.
3. Select one or more options.
4. Verify that selected values are displayed, clear button appears and clicking clear removes all selections.
5. Enable `multiselect` and `filter`.
6. Select one or more options, toggle menu to close and open.
7. Verify that: 
- selected values persist in the input when reopening
-  clicking inside the input allows filtering
- typing clears the previous display value and filters correctly
- clear button removes filter text first, then selected values on second click

## **Checklist before submission**
- [ ] Designer approves new design (if applicable)
- [ ] No accessibility violations in Storybook
- [ ] I have added unit tests for my changes (if applicable)
- [x] All existing tests pass
- [ ] I have updated the documentation (if applicable)
- [ ] Not breaking production behavior
- [ ] Behavior available in Storybook with documented descriptions (if applicable)
- [x] `npm run build:all` without errors

## **Suggested test steps**
- [x] Browser testing (Chrome, Safari, Firefox) 
- [x] Keyboard operability
- [ ] Interactive elements have labels.
- [ ] Storybook controls
- [ ] Design/controls/props is aligned with other components 
- [ ] Dark/light mode and variants 
- [ ] Input fields – values should be displayed properly 
- [ ] Events

PR #1751, opened 2026-02-27
fix(tds-dropdown): persist labels in multiselect + filter mode

adelezili:
## **Describe pull-request**  
This PR improves the `tds-dropdown` component by addressing label persistence issues and aligning buttons implementation across all modes.

* In `multiselect + filter` mode, selected labels now persist when toggling the dropdown menu.
* In `multiselect` mode, added clear button (consistent with filter mode behavior).
* Updated e2e tests to align with changed elements and regenerated baseline snapshots.

## **Issue Linking:**  
- **Jira:** [CDEP-1764](https://jira.scania.com/browse/CDEP-1764)

## **How to test**  
1. Go to Storybook and open `Dropdown` component.
2. Enable `multiselect`.
3. Select one or more options.
4. Verify that selected values are displayed, clear button appears and clicking clear removes all selections.
5. Enable `multiselect` and `filter`.
6. Select one or more options, toggle menu to close and open.
7. Verify that: 
- selected values persist in the input when reopening
-  clicking inside the input allows filtering
- typing clears the previous display value and filters correctly
- clear button removes filter text first, then selected values on second click

- [x] All existing tests pass

## **Suggested test steps**
- [x] Keyboard operability
- [ ] Storybook controls
- [ ] Events

Note:
This replaces PR #1744. All previous review feedback has been addressed.


nathalielindqvist (commented):
Nicely done! 🚀 
In my review, when selecting both multiselect and filter options, both the input text and the selected choices are cleared on the first click of the clear button. Has the intended behavior changed?

adelezili:
@nathalielindqvist thanks for checking! 
I think this part we had already: selected options in input field get cleared with single click, but if there's an unfinished query, that one gets cleared on first click and existing options in input field clear on the second click.
So nothing should be changed there 🤔

nathalielindqvist:
> @nathalielindqvist thanks for checking! I think this part we had already: selected options in input field get cleared with single click, but if there's an unfinished query, that one gets cleared on first click and existing options in input field clear on the second click. So nothing should be changed there 🤔

Gotcha, thanks for clarifying! I think this looks great, well done! 🥳

nathalielindqvist (approved):

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -575,9 +578,34 @@ export class TdsDropdown {
   };
```
timrombergjakobsson, line 594: Nice! Only a suggestion but maybe we could introduce a small helper that extracts the derive typed from displayValue and rawValue?

```
function getTypedQuery(rawValue: string, displayValue: string): string {
  if (rawValue.length <= displayValue.length + 1) {
    // Too short / ambiguous – just fall back to raw.
    return rawValue;
  }

  if (rawValue.startsWith(displayValue)) {
    // User typed after the labels: "<labels><typed>"
    return rawValue.slice(displayValue.length);
  }

  if (rawValue.endsWith(displayValue)) {
    // User typed before the labels: "<typed><labels>"
    return rawValue.slice(0, rawValue.length - displayValue.length);
  }

  // Fallback – user edited in the middle, pasted, etc.
  return rawValue;
}
```

And then we can use it in `handleFilter`

```
private handleFilter = (event: any) => {
  const rawValue: string = event.target.value;

  if (
    this.multiselect &&
    this.filterQuery.length === 0 &&
    this.selectedOptions.length > 0 &&
    this.inputElement
  ) {
    const displayValue = this.getValue();
    const typed = getTypedQuery(rawValue, displayValue);
    this.inputElement.value = typed;
  }

  this.tdsInput.emit(event);
  const query = this.inputElement
    ? this.inputElement.value.toLowerCase()
    : rawValue.toLowerCase();

  this.filterQuery = query;
  // ...rest of existing filter logic...
};
```

timrombergjakobsson: But really just a suggestion :)

ckrook (approved):
Really great work! Cannot find any issues on this one 🥇

packages/core/src/components/dropdown/test/default/disabled/dropdown.e2e.ts
```diff
@@ -17,7 +17,7 @@ testConfigurations.withModeVariants.forEach((config) => {
     });
 
     test('clicking the dropdown icon does not open the dropdown-list', async ({ page }) => {
-      const dropdownButtonIcon = page.locator('tds-icon');
+      const dropdownButtonIcon = page.locator('tds-icon[name="chevron_down"]');
```
ritamcastro: Here you can use the a11y locators from playwright:

`const dropdownButtonIcon = page.getByRole("button", { name: "Open/Close dropdown" });`

packages/core/src/components/dropdown/test/multiselect/disabled/dropdown.e2e.ts
```diff
@@ -17,7 +17,7 @@ testConfigurations.withModeVariants.forEach((config) => {
     });
 
     test('clicking the dropdown icon does not open the dropdown-list', async ({ page }) => {
-      const dropdownButtonIcon = page.locator('tds-icon');
+      const dropdownButtonIcon = page.locator('tds-icon[name="chevron_down"]');
```
ritamcastro: Same here, 

`const dropdownButtonIcon = page.getByRole("button", { name: "Open/Close dropdown" });`

packages/core/src/components/dropdown/test/multiselect/dropdown.e2e.ts
```diff
@@ -29,7 +29,7 @@ testConfigurations.withModeVariants.forEach((config) => {
     }) => {
       /* click the dropdown button */
       const dropdown = page.getByTestId('tds-dropdown-testid');
-      const dropdownButton = dropdown.getByRole('button');
+      const dropdownButton = dropdown.locator('button:not(.clear-icon)');
```
ritamcastro: We should aim to test for how the user is interacting with the screen, and not with how the elements are styled (css classes).

In this case, we can also do the 

```
await page.getByRole("button", { name: "Open/Close dropdown" }).click();
```

and then change the flow of the test to be more user driven, something like:

```
// Needs to add aria-label to the div containing the placeholder and the tds-aria-label in the index html 
const selectedOptionsText = page.getByLabel("Selected options for Dropdown component with multiselect") 

// Checks the default rendering state
await expect(selectedOptionsText).toHaveText("Placeholder")

 await page.getByRole("button", { name: "Open/Close dropdown" }).click();

/* Click the Option 1 button */
// We can simply click the Option 1 text, which is what the user is seeing
await page.getByText("Option 1").click()
await expect(page.getByRole("checkbox", { name: "Option 1" })).toBeChecked()

// Check that the dropdown button now has the text "Option 1"
// Note that it is not the button that has the value, it is the placeholder div
      await expect(selectedOptionsText).toHaveText("Option 1")
```

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -575,9 +578,34 @@ export class TdsDropdown {
...
 
   private handleFilter = (event) => {
+    if (
```
ritamcastro, line 598: Missing type for event in method signature

packages/core/src/components/dropdown/dropdown.tsx
ritamcastro, line 915: here we could use a `role='combobox'`

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -850,14 +940,37 @@ export class TdsDropdown {
                 <div class={`placeholder ${this.size}`}>
```
ritamcastro: and here for the placeholder div we can add a aria-label so that we can access it:

`                <div class={`placeholder ${this.size}`} aria-label={`Selected options for ${this.tdsAriaLabel}`}>`

packages/core/src/components/dropdown/test/default/disabled/dropdown.e2e.ts
adelezili: I think it's not great to use selector based on `role="button"` here (default non-filter mode), since chevron actually does not behave as a button. Its visual state is driven by the component's state (actual button element). Chevron currently has `role="button"` in filter mode only.
What do you think?

ritamcastro: You are right!
it shouldn't be the icon itself but the button/component itself ;)

- add the aria label to the index.html
- do the getbyRole("button", {name: "aria label from the index"});
- change variable name to remove icon ;)

packages/core/src/components/dropdown/test/multiselect/disabled/dropdown.e2e.ts
ritamcastro: ☝️ same as above

packages/core/src/components/dropdown/dropdown.tsx
adelezili: Seems like we wouldn’t get benefits from `combobox` role in this case. A button element here opens a list and then focus moves into options. “That pattern is better described as a button that opens a listbox, not a combobox.”
Do you agree?

packages/core/src/components/dropdown/test/multiselect/dropdown.e2e.ts
```diff
@@ -24,72 +24,63 @@ testConfigurations.withModeVariants.forEach((config) => {
...
-      // Check that the dropdown button now has the text "Option 1"
-      await expect(dropdownButton).toHaveText(/Option 1/);
+      /* Click Option 1 */
+      await page.locator('tds-dropdown-option[value="option-1"]').click();
+
+      /* Check that the selected text now shows "Option 1" */
+      await expect(page.getByRole('checkbox', { name: 'Option 1' })).toBeChecked();
+
+      /* Trigger now displays the selected label */
+      const valueText = dropdown.locator('div.placeholder');
```
ritamcastro, line 46: maybe instead of this locator we could go with ```dropdown.getByLabel("Selected options for {tds aria label of the parent}")```

packages/core/src/components/dropdown/test/multiselect/dropdown.e2e.ts
```diff
@@ -24,72 +24,63 @@ testConfigurations.withModeVariants.forEach((config) => {
...
-      await expect(dropdownButton).toHaveText(/Option 1, Option 3/);
-
-      await dropdownListElementFourButton.click();
-      // Now the button should show "Option 1, Option 3, Option 4"
-      await expect(dropdownButton).toHaveText(/Option 1, Option 3, Option 4/);
+      const triggerButton = dropdown.getByRole('button', { name: 'Label text' });
+      await triggerButton.click();
+
+      /* Open the dropdown */
+      // await page.getByRole('button', { name: 'Label text' }).click();
```
ritamcastro: 😅

packages/core/src/components/dropdown/test/multiselect/dropdown.e2e.ts
```diff
@@ -24,72 +24,63 @@ testConfigurations.withModeVariants.forEach((config) => {
...
+
+      /* Click Option 1 */
+      const option1 = page.locator('tds-dropdown-option[value="option-1"]');
+      await option1.click();
+      await expect(page.getByRole('checkbox', { name: 'Option 1' })).toBeChecked();
+      const valueText = dropdown.locator('div.placeholder');
```
ritamcastro, line 64: same as before (getByLabel("Selected options...")

packages/core/src/components/dropdown/test/multiselect/dropdown.e2e.ts
```diff
@@ -24,72 +24,63 @@ testConfigurations.withModeVariants.forEach((config) => {
...
-      await dropdownButton.click();
+      const triggerButton = dropdown.getByRole('button', { name: 'Label text' });
 
-      /* Click the Option 1 button */
-      const dropdownListElementOneButton = page
-        .locator('tds-dropdown-option')
-        .filter({ hasText: /Option 1/ });
-      await dropdownListElementOneButton.click();
+      /* Initial state: user sees placeholder text in the trigger */
+      await expect(triggerButton).toContainText('Placeholder');
```
ritamcastro, line 34: 👍 toContainText!!! Nice!

PR #1795, opened 2026-03-12
fix(tds-datetime): implement native min/max validation and add customValidator for users

mfilipadneto:
> [!warning]
> We had to create this new PR because of some conflict resolution issues with our old branch. In this branch we have the latest changes from develop. The [original PR](https://github.com/scania-digital-design-system/tegel/pull/1737) will be closed

## **Describe pull-request**  
This PR fixes the validation errors reported on the CDEP-1752 for the datetime component.

We have implemented the native validation of the user's input with the [validity of the HTMLInputElement API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/validity).
Additionally, we have created an _OPTIONAL_ `customValidator` function that the user may provide to perform their own validations on the input date/time.

We have created a new _OPTIONAL_ prop to the datetime component to hold the error message. The user may want to specify the helper text for nominal usage of the component and set a different error message.

We have created the `iphone` css class to be applied for the error and success validation.

We have merged the error css for `md` and `sm` cases as they were the same.

We have added an extra validation case for invalid dates (for example, 31-02-2026 or 31-11-2026), as the error message should be specific to this case and not the generic error message specified by the user.

We have created a new _OPTIONAL_ prop to the datetime component to hold the error message for invalid inputs. The user may want to specify the helper text for nominal usage of the component and set a different error message for the invalid input cases.

We have not done any snapshot tests for these changes since we drove the tests according to the user behaviour. 
We kept one test that renders the component in the error state to ensure that the graphic elements are all present (error icon and message in red (with the dark and light nuances)).

One of the acceptance criteria for this PR was that the `Date picker should respect min/max constraints on all platforms`. Due to [limitations of the native input type="date" element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/date#browser_compatibility) on iOS it is not possible to limit the range of the values shown on the date picker for this platform. However, the error message is displayed informing the user of their bad input.

## **Issue Linking:**  
- **Jira:** [CDEP-1752](https://jira.scania.com/browse/CDEP-1752)

## **How to test**  
1. Open Storybook and navigate to Components/Datetime/Default
2. Set the type to `date`, for example.
3. Set a `min` and `max` values for the date
4. Manually add a date that is outside of the limits set in 3.
5. Remove focus from the input element
6. Watch the error helper appear. If you want, in step 2 configure the helper error text to confirm that it renders your intended message 
7. Cycle through the other input types available.

In alternative, you can also checkout the branch and run the tests in docker. 
⚠️  if you run the tests locally with `npx playwright test --ui` the tests for the input type="date" might fail due to different date formats in different browsers. In docker, which is the reference environment, the format is "MM/dd/yyyy" and in our local environment it was "dd/MM/yyyy".

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**  
- the screenshots for the form that uses the datetime input has also been updated

## **Additional context**  
n/a


argo861 (approved):

ckrook (approved):

packages/core/src/components/datetime/datetime.tsx
```diff
@@ -267,7 +330,7 @@ export class TdsDatetime {
             />
 
             {this.labelPosition === 'inside' && this.size !== 'sm' && this.label && (
-              <label class={`tds-datetime-label-inside ${iphone && 'iphone'}`} htmlFor={this.name}>
+              <label class="tds-datetime-label-inside" htmlFor={this.name}>
```
timrombergjakobsson, line 333: We can ignore this.  False positive caused by rule/profile misclassification (React rules applied to Stencil TSX). The right fix is to adjust Sonar’s rule scope/profile for Stencil files or mark the issue as False Positive, not to change valid Stencil code.

timrombergjakobsson (approved):

PR #1812, opened 2026-03-19
fix(tds-dropdown): warning handling

adelezili:
## **Describe pull-request**  
This PR fixes premature dropdown validation when using async options. Validation previously ran too early, causing false "Option does not exist" warnings. 

Replaced the timing-based isInitialLoad flag with:
* Strict validation (warn + filter): Only user interactions - option clicks, clear, reset.
* Delayed warnings: unmatched values are tracked as "pending" and only warn if still unmatched on the next slot change. This eliminates spam with async options while still catching genuinely invalid values.

Tested: Angular demo page with a dropdown containing one static option, async enum options via *ngFor, and a pre-set value matching async options - no premature warnings, proper warnings when expected.

## **Issue Linking:**  
- **Jira:**  [CDEP-1962](https://jira.scania.com/browse/CDEP-1962)

## **How to test**  
REPRODUCE THE ISSUE
1. Go to: https://pr-91.d24qjtb7na4hmr.amplifyapp.com/stepper  
2. Click on `About` in the side menu  
3. Open DevTools → Console  
4. Observe warning:  
   "Option with value "NonExistent" does not exist"
5. Notice that the warning appears before async dropdown options finish loading

VERIFY THE FIX

6. Go to: https://pr-92.d24qjtb7na4hmr.amplifyapp.com/stepper
7. Click on `About` in the side menu  
8. Open DevTools → Console  
9. Verify that no warning appears

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -195,21 +197,19 @@ export class TdsDropdown {
...
       const isValid = children.some(
         (element) => convertToString(element.value) === convertToString(val),
       );
-      if (!isValid) {
-        console.warn(`Option with value "${val}" does not exist`);
+      if (!isValid && !this.isInitialLoad) {
+        console.warn(`TDS DROPDOWN: Option with value "${val}" does not exist`);
       }
-      return isValid;
+      return isValid || !this.isInitialLoad;
```
ckrook: `!this.isInitialLoad` is always `true` after load 

`isValid` could be either `true` or `false` depending on the validity.
Because we have an OR operator this means that this expression is always true after the initialLoad.

adelezili: Good catch, this was making warnings ineffective... Fixed now with another approach - changing timing based validation to be context based. 
Thanks for noticing this!

mfilipadneto:
Checking out this PR we got a little confused with the code on the branches in the Angular Demo. Just because it is said to verify the fix through opening the PR-91 branch (before fix) and PR-92 branch (after fix), where the 91 shows the warning in the console and 92 does not. However, the code for both branches is different, as in the `[value]` from the dropdown is `'nonExistent'` or `testEnum.Completed`. Which we find a little confusing, since nonExistent is not part of the testEnum so it's expected that the "Option does not exist" warning shows up. 
Maybe we are not following the issue completely. 
Can someone clear this up for us?

mfilipadneto:
The PR-91 (before the fix) from Angular 17 has been updated with the same `[value]="testEnum.Completed"` as PR-92 (after the fix) so that it can be compared appropriately. 

Indeed, it can be seen before the fix ([PR91](https://pr-91.d24qjtb7na4hmr.amplifyapp.com/stepper)) that the warning `Option with value "Completed" does not exist` shows, while, after the fix ([PR92](https://pr-92.d24qjtb7na4hmr.amplifyapp.com/stepper)), no warning related to that is shown. 

So, looks good to me 👍

mfilipadneto (approved):

antonstrandqvist (approved):

Issue #1845, opened 2026-04-03
CSS parsing error: Invalid custom property name containing % in tegel.css


raviscania:
When building a project that uses @scania/tegel, the bundler fails with a CSS parsing error.
The problem originates from an invalid CSS custom property name inside tegel.css.

**Error message**
```
Parsing CSS source code failed
'var' is not recognized as a valid pseudo-class.
```
Excerpt from the error:
```
--component-button-opacity-disabled: var(--scania-opacity-25%);
--component-button-opacity-disabled: var(--traton-opacity-25%);
```

**Root cause**
The file:
`node_modules/@scania/tegel/dist/tegel/tegel.css`
contains the following invalid custom property:
```
--scania-opacity-25%
--traton-opacity-25%
```
CSS custom property names cannot contain %, causing PostCSS / Vite / css-loader to fail.

Problematic code (from tegel.css)
`--component-button-opacity-disabled: var(--scania-opacity-25%);`
Since variable names cannot include %, this causes a syntax error during build.

**Expected behavior**
The CSS should use valid variable naming, e.g.:
```
--scania-opacity-25: 25%;
--component-button-opacity-disabled: var(--scania-opacity-25);
--traton-opacity-25: 25%;
--component-button-opacity-disabled: var(--traton-opacity-25);
```

mfilipadneto:
Hey Ravi! 
Sorry for the delayed response here. 
We will have a release tomorrow to fix this issue, we will let you know when that is done.
Thanks!

PR #1847, opened 2026-04-03
Fix opacity variable format for disabled chip states

raviscania:
CSS should use valid variable naming

--component-button-opacity-disabled: var(--scania-opacity-25);
--component-button-opacity-disabled: var(--traton-opacity-25)

## **Describe pull-request**  
CSS should use valid variable naming like above but current code with %

--component-button-opacity-disabled: var(--scania-opacity-25%);
--component-button-opacity-disabled: var(--traton-opacity-25%);

## **Issue Linking:**  
- **GitHub:** [Include issue link ](https://github.com/scania-digital-design-system/tegel/issues/1845)


ckrook:
Hi @raviscania , This file is auto generated. We do not change the file itself but rather want to change the script which generates the tokens.

Me and my colleague are addressing this in our ongoing variable work. Therefore I will close this!

Issue #1860, opened 2026-04-14
[Bug report]: Modal not shown without trigger element since v1.48.0


patrikhultgren:
### Requirements before reporting

- [x] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported.
- [x] I have checked the latest version if the bug exist there. See all available packages at [npmJS.com](https://www.npmjs.com/package/@scania/tegel)
- [x] I have followed the installation guide.

### Package versions

@scania/tegel: 1.51.0

### Browser

Chrome

### Framework

React

### Version

React 19

### Code sandbox

Modal is not displayed unless a trigger element is provided. This behavior differs from previous versions and breaks existing implementations where no trigger was required.

Example when modal is not displayed:
https://stackblitz.com/edit/vitejs-vite-3bmjcmvd?file=src%2FApp.tsx

### Reproduction steps

1. Open the link https://stackblitz.com/edit/vitejs-vite-3bmjcmvd?file=src%2FApp.tsx
2. The modal is not displayed.

### Screenshots / Screen recordings

The modal is not displayed:

<img width="1532" height="1653" alt="Image" src="https://github.com/user-attachments/assets/7506ffe7-4489-4b9c-bcb2-347c467d218d" />

### Expected behaviour

The modal should be shown, just like in the example below:
https://stackblitz.com/edit/vitejs-vite-ye371wce?file=src%2FApp.tsx

<img width="1531" height="1656" alt="Image" src="https://github.com/user-attachments/assets/076a6120-7593-4a97-8824-b6fa870c983a" />

### Contact information

patrik.hultgren@scania.com

mfilipadneto:
Hello @patrikhultgren 🙂 

Thanks for your report! We were able to reproduce the bug on our end, and after some investigation we have some idea where the bug originated. 

We will create a ticket in Jira and address this as soon as possible, letting you know when it is done. 

Thanks 😄

patrikhultgren:
Great! I'm looking forward for the solution 👍

ritamcastro:
Hey @patrikhultgren !
We have just released v1.53.0 with this correction. 
We are closing this issue but if you have any questions do not hesitate to reach out to us 😄
Cheers,
Tegel Team

PR #1870, opened 2026-04-22
fix(tds-modal): Modal not showing with show as true

mfilipadneto:
## **Describe pull-request**  
This PR fixes a bug reported on [GitHub](https://github.com/scania-digital-design-system/tegel/issues/1860). 
This bug was introduced in version 1.48.0 and unintentionally introduced a breaking change.

The modal wasn't showing when the `show` prop was set as true, when first opening the page containing the modal.
More details can be seen in the issue itself, since it is very well reported. 

We understood that the issue stemmed from the fact that the verifications needed to change the state `isShown` were only being done on the `connectedCallback` function, and not on the `componentWillLoad` function. 
So we simply added the same verifications on both functions of the lifecycle of the component.

## **Issue Linking:**
- **Jira:** [CDEP-2064](https://jira.scania.com/browse/CDEP-2064)

## **How to test**
1. Go to the [React demo page](https://pr-307.d2vh2lmnzgzrkl.amplifyapp.com) with this beta release ([PR-307](https://github.com/scania-digital-design-system/tegel-react-demo/pull/307) from the tegel-react-demo)
2. Open the modal component page from the side menu
3. Verify that there is a modal opened at the start of the page

- [ ] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Additional context**  
@patrikhultgren, if you'd like to check this one out and give feedback, feel free 😄

### DIFF
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -255,7 +255,6 @@ export class TdsModal {
 
   /** Runs whenever the modal is opened and updates it. */
   private onOpen() {
-    console.log('onOpen running...');
     // Focus immediately to preserve interaction modality for :focus-visible
     this.focusFirstElement();
 
@@ -313,11 +312,9 @@ export class TdsModal {
   };
 
   handleShow = () => {
-    console.log('handleShow running...');
     const showEvent = this.tdsOpen.emit();
     if (showEvent.defaultPrevented) return;
 
-    console.log('handleShow, after if -> return...');
     this.isShown = true;
     this.onOpen();
   };
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -45,8 +45,9 @@ export class TdsModal {
   /** Element that will show the Modal (takes priority over selector) */
   @Prop() referenceEl?: HTMLElement | null;
 
-  /** Controls whether the Modal is shown or not. If this is set hiding and showing
-   * will be decided by this prop and will need to be controlled from the outside. */
+  /** Controls whether the Modal is shown or not. <br/> This prop allows the consumer of Tegel to control
+   * the open/close interaction or set the modal visibility when opening the page. If it is not set, then
+   * the modal has a fallback state for that interaction, defaulting to false. */
   @Prop({ reflect: true }) show?: boolean;
 
   /** Shows or hides the close [X] button. */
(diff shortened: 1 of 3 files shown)


packages/core/src/components/modal/modal.tsx
```diff
@@ -250,6 +255,7 @@ export class TdsModal {
 
   /** Runs whenever the modal is opened and updates it. */
   private onOpen() {
+    console.log('onOpen running...');
```
ckrook: please remove:)

packages/core/src/components/modal/modal.tsx
```diff
@@ -307,9 +313,11 @@ export class TdsModal {
   };
 
   handleShow = () => {
+    console.log('handleShow running...');
```
ckrook: please rm

packages/core/src/components/modal/modal.tsx
```diff
@@ -307,9 +313,11 @@ export class TdsModal {
...
     const showEvent = this.tdsOpen.emit();
     if (showEvent.defaultPrevented) return;
 
+    console.log('handleShow, after if -> return...');
```
ckrook: remove:)

packages/core/src/components/modal/modal.tsx
```diff
@@ -136,8 +137,12 @@ export class TdsModal {
     }
   }
 
+  connectedCallback() {
+    this.initializeWithProps();
+  }
+
   componentWillLoad() {
-    this.initializeModal();
+    this.initializeWithProps();
```
ckrook, line 145: I would say:

Do this if you understand what it means to run `this.initializeWithProps();` both in the connectedCallback and componentWillLoad.

Have a read on stencil lifecycles would be my recommendation:)
https://stenciljs.com/docs/component-lifecycle

packages/core/src/components/modal/modal.tsx
mfilipadneto: Oops, thanks Charles

packages/core/src/components/modal/modal.tsx
mfilipadneto: Thanks for the recommendation! That was precisely our reference. 
We spent some time on it, and from our understanding, the `componentWillLoad` runs once when the component is first inserted in the DOM, and the `connectedCallBack` runs every time it is reattached.

Our logic here is that these checks and initialization inside `initializeWithProps()` is very necessary in both cases. 🙂

ckrook (approved):

ritamcastro (approved):

PR #1879, opened 2026-04-30
feat(tds-icon): brand-aware icons via CSS cascade

ckrook:
## Summary
- Replaces the JS-side `closest('.scania' | '.traton')` brand lookup in `tds-icon` with a cascade-driven approach: each icon's `<path d>` is now supplied by a brand-scoped CSS variable (`--tds-icon-<name>-d`), the same way `chip.scss` & co. theme components today.
- Adds `scripts/generate-icon-paths.mjs`, which reads the existing `scaniaIconsArray.js` / `tratonIconsArray.js` and emits `tokens/scss/component/icon.scss` (252 Scania + 29 Traton paths), and wires that file into `core.scss` and `global.scss`.
- The component lost its `componentWillLoad`, both icon-array imports, the `arrayDataWatcher` JSON-parse step, and two `@State` fields; render is now a single SVG with `<path style={{ d: var(--tds-icon-<name>-d) }} />`. Title/desc/aria wiring is unchanged.

## Notes for reviewers
- `closest()` only ran once at load, so re-parenting an icon under a different brand wrapper used to be a silent no-op. With the cascade approach, brand switching just works on re-render — no JS state to reconcile.
- Browser support: this relies on `d` as a CSS property (Chrome 108+, Safari 16+, Firefox 97+). Confirm that's compatible with Tegel's supported-browsers matrix.
- `scaniaIconsArray.js` / `tratonIconsArray.js` are intentionally left in the tree but no longer imported. Deleting them is a follow-up once this lands.
- `generate-icon-paths.mjs` is currently a standalone script — it isn't yet invoked by `build-tokens.mjs`. Folding it in (and deciding whether the path data should come from Figma alongside the existing `scania-icons.scss` URL variables) is its own conversation.

## Test plan
- [ ] Storybook: default `tds-icon` story renders Scania icons (no wrapper class).
- [ ] Wrap a story in `<div class="traton">…</div>` and confirm the same `name` switches to the Traton path data.
- [ ] Inspect a `tds-icon` with `svgTitle` + `svgDescription` set; verify `<title id="title-…">` and `<desc id="desc-…">` are still emitted and referenced by `aria-labelledby`/`aria-describedby`.
- [ ] Confirm `--tds-icon-truck-d` resolves in DevTools' Computed pane (sanity check that the new SCSS file is in the bundle).
- [ ] Run the existing icon e2e/visual snapshots.

[REACT DEMO PAGE](https://pr-308.d2vh2lmnzgzrkl.amplifyapp.com/about)
[ANGULAR DEMO PAGE](https://pr-106.d24qjtb7na4hmr.amplifyapp.com/about)
[TEGEL LITE DEMO PAGE](https://pr-35.d1gpvxaycl8qno.amplifyapp.com/about)


NordbergAnna (commented):
For Tegel Lite icon component in storybook, many icons don't show at all either in Scania or Traton, it works as expected in the web component.

NordbergAnna (approved):
Nice work, LGTM! 🥳

antonstrandqvist (approved):

PR #1883, opened 2026-05-03
feat(table): add current sort state props

joakimglaving:
## **Describe pull-request**  
Adds support for displaying the current sort state in `tds-table` without requiring the user to click a sortable column first.

This adds two reflected props on `tds-table`:

- `sort-column-key`
- `sort-direction`

`tds-header-cell` now reads those table props on initial load and listens for later updates, so consumers can keep the visual sort indicator in sync with their own sorting state.

## **Issue Linking:**  
- **No issue:** Enables consumers to show the active table sort column and direction from external state before any user interaction.

## **How to test**  
1. Start Storybook and go to `Components / Table / Sorting`.
2. Use the `Current sorted column` and `Current sort direction` controls.
3. Verify the matching sortable column shows the correct arrow direction before clicking.
4. Click the sorted column and verify the existing `tdsSort` event behavior still works.

- [x] All existing tests pass


packages/core/src/components/table/table-header-cell/table-header-cell.tsx
```diff
@@ -129,17 +135,25 @@ export class TdsTableHeaderCell {
     key: string;
   }>;
 
-  @Listen('internalTdsPropChange', { target: 'body' })
+  @Listen('internalTdsTablePropChange', { target: 'body' })
```
ckrook, line 135: I think we need to update the name on `table-bode-cell.tsx Line 52` to `internalTdsTablePropChange` as well

ckrook (commented):
Heej! Left a comment :)

packages/core/src/components/table/table-header-cell/table-header-cell.tsx
joakimglaving: Ahh, I think you are right! Can you update it? I'm busy until Friday. Thank you!

joakimglaving: Fixed it now :)

joakimglaving:
@ckrook have a look when you have time

packages/core/src/components/table/table/table.tsx
```diff
@@ -81,6 +85,12 @@ export class TdsTable {
    * */
   @Prop({ reflect: true }) horizontalScrollWidth?: string | null = null;
 
+  /** Column key for the currently sorted column. */
+  @Prop({ reflect: true }) sortColumnKey?: string;
```
ritamcastro, line 89: Does it makes sense to call this `initialSortColumnKey` and `initialSortDirection` ?

From our understanding these props are used exclusively to show the initial ordering as set by the developer.

packages/core/src/components/table/table-header-cell/table-header-cell.tsx
```diff
@@ -171,6 +185,22 @@ export class TdsTableHeaderCell {
     relevantTableProps.forEach((tablePropName) => {
       this[tablePropName] = this.tableEl?.[tablePropName];
     });
+    this.syncSortingStateFromTable();
+  }
+
+  private isSortDirection(sortDirection: string | undefined): sortDirection is SortDirection {
```
ritamcastro, line 188: Why is this type a string and not a `SortDirection`? 🤔

packages/core/src/components/table/table-component-sorting.stories.tsx
```diff
@@ -77,6 +77,29 @@ export default {
         defaultValue: { summary: false },
       },
     },
+    sortColumnKey: {
+      name: 'Current sorted column',
+      description:
+        'Sets which column should display as currently sorted. The value must match a header cell key.',
```
ritamcastro, line 83: Shouldn't this be "Sets which column is sorted when the table is initially rendered." ? 
Same comment for the sort direction down below :)

ritamcastro:
hey @joakimglaving !
We have taken a look at your PR and we left some minor comments :)

Whenever you can, please take a look and let us know if you have any thoughts.
Thanks very much for your contribution 😄

packages/core/src/components/table/table/table.tsx
joakimglaving: Not if we intend these props to represent the controlled current sort state. This PR supports runtime updates, and the e2e test covers that behavior.

If we rename them to `initialSortColumnKey` / `initialSortDirection`, consumers would likely expect them to be read only once on initial render.

My intention was that consumers can keep these props in sync with their own state and also programmatically change the displayed sort order from outside the table.

packages/core/src/components/table/table-header-cell/table-header-cell.tsx
joakimglaving: Because this value can come from HTML, TypeScript cannot fully protect it at runtime.

Someone could write:

```html
<tds-table sort-direction="wrong-value">
```

So this helper checks that the value is actually asc or desc before we use it as a valid sort direction.

packages/core/src/components/table/table-component-sorting.stories.tsx
joakimglaving: I think that wording would fit if the props were only used for the first render.

Since the props are meant to represent the current controlled sort state, I think the current wording should stay broader. They can be used both for the initially rendered state and for later programmatic updates.

joakimglaving:
Hi @ritamcastro thanks for your feedback, I've replied to the comments now!

mfilipadneto (approved):

claraseverino (approved):

PR #1884, opened 2026-05-04
feat(variables): textarea textfield

ckrook:
## **Describe pull-request**  
The Textarea and Text-field components have been migrated to use the new design token variables, and their -vars.scss files updated accordingly.

## **Issue Linking:**  
[CDEP-XXX](https://jira.scania.com/browse/CDEP-XXX)

## **How to test**  
- [ ] Verify that WCAG checks passes in storybook
- [ ] Visually audit the snapshots diff
- [ ] Open the components in storybook and make sure they look good
Make a table here with

Component | Regular Web Component | Tegel Lite Component

Component | Regular Web Component | Tegel Lite Component
-- | -- | --
Textarea | [Preview Link](https://pr-1884.d3fazya28914g3.amplifyapp.com/?path=/story/components-textarea--default) | [Preview Link](https://pr-1884.d3fazya28914g3.amplifyapp.com/?path=/story/tegel-lite-beta-textarea--default)
Text-field | [Preview Link](https://pr-1884.d3fazya28914g3.amplifyapp.com/?path=/story/components-text-field--default) | [Preview Link](https://pr-1884.d3fazya28914g3.amplifyapp.com/?path=/story/tegel-lite-beta-text-field--default) 

[TEGEL REACT DEMO PAGE WITH THESE UPDATE](https://pr-315.d2vh2lmnzgzrkl.amplifyapp.com/about)

[TEGEL ANGULAR DEMO PAGE WITH THESE CHANGES](https://pr-114.d24qjtb7na4hmr.amplifyapp.com/about)

[TEGEL LITE DEMO PAGE](https://pr-53.dxqjewiof3vdp.amplifyapp.com/about)

<br class="Apple-interchange-newline">

## **Checklist before submission**
- [ ] Designer approves new design (if applicable)
- [ ] No accessibility violations in Storybook
- [ ] I have added unit tests for my changes (if applicable)
- [ ] All existing tests pass
- [ ] I have updated the documentation (if applicable)
- [ ] Not breaking production behavior
- [ ] Behavior available in Storybook with documented descriptions (if applicable)
- [ ] `npm run build:all` without errors

## **Suggested test steps**
- [ ] Browser testing (Chrome, Safari, Firefox) 
- [ ] Keyboard operability
- [ ] Interactive elements have labels.
- [ ] Storybook controls
- [ ] Design/controls/props is aligned with other components 
- [ ] Dark/light mode and variants 
- [ ] Input fields – values should be displayed properly 
- [ ] Events

PR #1919, opened 2026-05-25
test: stabilize Playwright suite after 1.60 bump

ckrook:
Stabilizes a handful of flaky Playwright tests and refreshes baselines that drifted from the current code.

## Changes
- **Sorting table (`column-headers-are-clickable`):** add `waitForTimeout(250)` between header clicks so the cell's 200 ms reset `setTimeout` settles before the next click; regenerate the four primary/secondary × light/dark baselines so Mileage shows the asc arrow deterministically.
- **Datetime error (`out-of-bounds`):** force `locale: 'en-US'` in `playwright.config.ts` and swap the locale-dependent `pressSequentially` typing for `fill('2026-04-01')`.
- **Modal open-close:** wait for `tds-modal.hydrated` before tests dispatch the open click, so the modal's selector listener is attached.
- **Text-field focus snapshots:** regenerate to match the new shared `tds-focus-state` mixin (outline + box-shadow) introduced by #1885 / #1884.
- **Playwright docker image:** keep `docker-test-runner.js` and the `update:snapshots` script aligned on `v1.60.0-jammy`, matching the declared `@playwright/test` in `package.json` and the CI workflow.


packages/core/src/components/text-field/test/focus/text-field.e2e.ts-snapshots/tds-text-field-focus-allows-text-to-be-entered-after-focus-1-linux.png
ritamcastro, line 1: why is this with camel case instead of the kebab? all the other ones are with kebab case 🤔

ckrook: It's because the describe title in this file is hard-coded as 'TdsTextField - Focus', while the other components use the kebab-case component name (e.g. 'tds-badge') as the describe title. Playwright derives the snapshot filename straight from that string

NordbergAnna (approved):
All test's pass 🌟

packages/core/src/components/datetime/test/error/datetime.e2e.ts
```diff
@@ -66,12 +65,7 @@ testConfigurations.withModeVariants.forEach((config) => {
 
       const datetime = page.getByLabel("DateTime component with type='date'");
 
-      await datetime.click();
-
-      await datetime.pressSequentially('01');
-      await datetime.pressSequentially('04');
-      await datetime.pressSequentially('2026');
-
+      await datetime.fill('2026-04-01');
```
ritamcastro: What was the reasoning to change the pressSequentially to the fill?  At the time, fill was not working as expected - maybe due to the browser specs that are now fixed :) 

Why changing it only here instead of keeping coherence and changing it everywhere?

packages/core/src/components/text-field/test/focus/text-field.e2e.ts-snapshots/tds-text-field-focus-allows-text-to-be-entered-after-focus-1-linux.png
ritamcastro: should it be fixed to keep consistency with all the other files/tests?

ritamcastro:
We have reviewed and have two open points (to summarize them here):
- change pressSequentially to fill on the datetime tests
- change naming of the TdsTextField test to match the other tests

mfilipadneto (approved):

PR #1941, opened 2026-06-16
fix(tl-datetime): set TRATON empty-state text color to prevent Safari aqua fallback

ckrook:
**Issue (Max Adolfsson, design):** In Safari the Datetime field shows aqua placeholder text in its empty/default state ("None", before a date is picked) under the TRATON theme. Scania looked OK.

**Root cause:** The tegel-lite placeholder variable pointed at `--color-foreground-text-subtle` — a token that **doesn't exist** in the published `@scania/tegel-lite` package (verified: 0 definitions in the compiled CSS). So the placeholder colour resolved to nothing and Safari fell back to its aqua UA default. (The earlier `color`-only attempt couldn't work for the same reason.)

**Fix (tegel-lite only):**
- Repoint `--datetime-placeholder` / `-disabled` / `-focus` at the real, brand-defined tokens (`--color-text-subtle` / `--color-text-disabled`) — the same ones lite text-field/textarea use.
- Also set `-webkit-text-fill-color` (not just `color`) on the empty-state `::-webkit-datetime-edit` + fields-wrapper, since Safari ignores `color` there. Scoped to `:not(:focus)` so a filled value isn't dimmed.

Core `tds-datetime` is unaffected (its token chain resolves for both brands). No `tokens/` files touched.

**How to test (in Safari):**
1. Storybook → Tegel Lite → Datetime, brand = TRATON, leave the field empty.
2. The "yyyy/mm/dd hh:mm" placeholder should be subtle grey — not aqua.
3. Switch to Scania (still correct); pick a date → the entered value shows in the normal strong colour (not dimmed).

### DIFF
--- a/packages/core/src/tegel-lite/components/tl-datetime/_tl-datetime-vars.scss
+++ b/packages/core/src/tegel-lite/components/tl-datetime/_tl-datetime-vars.scss
@@ -5,21 +5,21 @@
   --datetime-background: var(--datetime-background-primary);
   --datetime-text: var(--color-foreground-text-strong);
   --datetime-icon: var(--color-foreground-icon-strong);
-  --datetime-placeholder: var(--color-foreground-text-subtle);
+  --datetime-placeholder: var(--color-text-subtle);
 
   // Disabled
   --datetime-background-disabled-primary: var(--color-background-layer-01);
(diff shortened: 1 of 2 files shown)


NordbergAnna (approved):
LGTM! 🌟

mfilipadneto (approved):

PR #1942, opened 2026-06-16
fix(datetime): apply standard disabled opacity treatment at host level

ckrook:
**Issue (Max Adolfsson, design):** When Datetime is disabled it only shows a not-allowed cursor — it doesn't dim like other disabled components.

**Fix:** Apply the standard ~25% disabled opacity to the field + label as one unit (helper text stays readable), in both core and Tegel Lite, using the shared disabled-opacity token.

**How to test:**
1. Storybook → Datetime, set `disabled` = true.
2. The field and its label should fade together to ~25% opacity; helper text stays full opacity; cursor shows not-allowed.
3. Check both light and dark mode, and both the core and Tegel Lite stories.


NordbergAnna (commented):
In Tegel Lite Scania dark mode, everything is disabled
Screenshot 2026-06-16 at 15 04 52

But in the web component the helper text is not disabled:
Screenshot 2026-06-16 at 15 05 13

mfilipadneto:
> helper text stays full opacity


This does not verify in the core components from what I can see, the color is slightly different: 

### Example: Scania light mode
Not disabled:
image

Disabled:
image

### Example: TRATON dark mode
Not disabled: 
image

Disabled:
image


Helper text becomes a bit less opaque when disabled both in light mode and dark mode. 




------

mfilipadneto:
For Tegel Lite I subscribe to what Anna pointed out. Helper text also seems to have less opacity (same behavior as Label text). So it does not go according to the "How to test" description

NordbergAnna (commented):
There's a slight discrepancy between the Web Component and Tegel Lite in lightmode. In the Web Component, all text has opacity: 0.25 when disabled, Tegel Lite is slightly darker/stronger.

Also, looking at your PR description, it mentions that the helper text should maintain full opacity. Right now, it's at 25%, so I'm a bit unsure what the correct spec is here (I know I previously commented that it shouldn't be 25%, but maybe that was intended after all?).

PR #1943, opened 2026-06-16
fix(modal): override Scania dark-mode background to raised surface

ckrook:
**Issue (Max Adolfsson, design):** In Scania dark mode the modal background is wrong — it blends into the page instead of sitting on a raised surface.

**Fix:** Override the modal background to the raised surface (`layer-02`) for Scania dark, at the component level (core + Tegel Lite).

**Note:** Root cause is the Figma token export mapping Scania-dark modal background to the page base (`layer-01`) instead of `layer-02` like every other theme/mode. Per our rule we don't hand-edit exported tokens, so this overrides at the component level — once the token is corrected in Figma, this override can be removed.

**How to test:**
1. Storybook → Modal, set theme to Scania + dark mode.
2. Open the modal — its surface should be a raised grey, clearly lighter than the page behind it.
3. Sanity-check Scania light, TRATON light, TRATON dark — unchanged. Check the Tegel Lite modal too.

### DIFF
--- a/packages/core/src/components/modal/modal-vars.scss
+++ b/packages/core/src/components/modal/modal-vars.scss
@@ -9,6 +9,19 @@ tds-modal {
   --tds-modal-text-color: var(--component-modal-text-title-default);
   --tds-modal-body-text-color: var(--component-modal-text-body-default);
   --tds-modal-icon-color: var(--component-modal-icon-dismiss-default);
-  --tds-modal-box-shadow: 0 3px 3px 0 var(--component-shadow-default, rgba(0, 0, 0, 0.15)),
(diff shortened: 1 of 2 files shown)


ckrook:
Closing per review — the component-level override isn't the right place to fix this.

Root cause is in the Figma-exported token: for **Scania dark**, `modal.background.default` (→ `--component-modal-background-default`) maps to `color.background.layer-01` (page base) instead of `color.background.layer-02` (the raised surface used by Scania light, TRATON light and TRATON dark).

**Fix in Figma:** set the Scania dark modal background to **layer-02**. Once re-exported, no code change is needed.

PR #1944, opened 2026-06-16
fix(toast): render status icon at consistent size across variants

ckrook:
**Issue (Max Adolfsson, design):** The status icon appears to change size between toast variants (most visible on Warning), which shifts the component width.

**Fix:** Pin the leading status icon to a fixed 20×20 box so every variant reserves identical space (no width shift). Tegel Lite was already pinned.

**How to test:**
1. Storybook → Toast, cycle Information / Success / Warning / Error with the same text.
2. The icon size and total toast width should be identical across variants (Warning should now match the others).

### DIFF
--- a/packages/core/src/components/toast/toast.scss
+++ b/packages/core/src/components/toast/toast.scss
@@ -76,6 +76,14 @@
   }
 
   tds-icon {
+    /* Pin the leading status icon to a fixed box so every variant (info /
+       success / warning / error) reserves the same width and height. Without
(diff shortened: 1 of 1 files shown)


ckrook:
Closing — couldn't reproduce a real icon-size/width difference between toast variants in Storybook. The change was only defensive (pinning the status icon to a fixed 20×20 box) and isn't needed if there's no visible issue. Happy to revisit if the Warning-variant shift reappears.

PR #1946, opened 2026-06-16
fix(icon): render TRATON placeholder fallback in Safari

ckrook:
**Issue (Max Adolfsson, design):** In TRATON, when an icon name isn't found it should fall back to a placeholder — but the fallback doesn't show in Safari.

**Fix:** Safari reports support for the CSS `d` property but doesn't substitute `var()` inside it (WebKit bug 219448), so the icon path rendered empty. Replaced the static feature check with a runtime probe that detects whether `var()` actually resolves in `d`; Safari then uses the existing JS fallback that writes the path attribute. Chrome/Firefox behaviour is unchanged.

**How to test:**
1. In Safari, set theme to TRATON.
2. Render `<tds-icon name="truck_inactive">` (a name not in the TRATON set) — the dashed-square placeholder should appear (previously blank).
3. Confirm Chrome and Firefox still render icons identically (no regression).

### DIFF
--- a/packages/core/src/components/icon/icon.tsx
+++ b/packages/core/src/components/icon/icon.tsx
@@ -5,12 +5,48 @@ import { IconNames } from '../../types/Icons';
 // supported (Chromium, Firefox, Safari 18+) the path renders reactively from
 // the brand-scoped --tds-icon-<name>-d variable via the CSS `d` property, so
 // none of the imperative getComputedStyle / MutationObserver fallback runs —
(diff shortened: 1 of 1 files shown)


ckrook:
Closing — the original TRATON placeholder-fallback issue isn't reproducible on review. The change was a guard for a real WebKit bug (var() not resolving in the CSS `d` property) but isn't worth shipping if the problem can't be confirmed. Easy to revisit if a blank placeholder shows up in real Safari (16.4+, TRATON).

PR #1947, opened 2026-06-16
fix(link): order Storybook stories consistently with other components

ckrook:
**Issue (Max Adolfsson, design):** In Storybook the Link sidebar entries are ordered oddly (Docs/Notes vs the stories) compared to other components.

**Fix:** Renamed the primary Link story to `Default` (the convention every other component uses) so the global alphabetical storySort lists it first → Default, Link Within Text, Docs, Notes.

**How to test:**
1. Run Storybook → Components → Link.
2. The sidebar order should read: Default, Link Within Text, Docs, Notes — matching Button/Checkbox/Toggle.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/charles.krook/Umain/Scania/tegel/node_modules
--- a/packages/core/node_modules
+++ b/packages/core/node_modules
@@ -1,0 +1,1 @@
+/Users/charles.krook/Umain/Scania/tegel/packages/core/node_modules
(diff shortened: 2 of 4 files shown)


NordbergAnna (commented):
Looks correct in the web component but not following the same order in Tegel Lite
Image

packages/core/node_modules
```diff
@@ -0,0 +1 @@
+/Users/charles.krook/Umain/Scania/tegel/packages/core/node_modules
```
mfilipadneto, line 1: Not sure what this is!! Please check Charles 😅

node_modules
```diff
@@ -0,0 +1 @@
+/Users/charles.krook/Umain/Scania/tegel/node_modules
```
mfilipadneto, line 1: Same here!! 
Not sure what this is!! Please check Charles 😅

mfilipadneto:
The Amplify check failed! Maybe re-run? @ckrook

PR #1949, opened 2026-06-16
fix(tl-footer): links should not be bold

ckrook:
Aligns tegel-lite footer links with the regular tds-footer by using the detail-02 type style (normal weight) instead of headline-07 (bold). Both are 14px, so only the weight changes.


NordbergAnna (approved):

mfilipadneto:
From what I can see, seems like the font family isn't being applied, because it looks like a different font-family from your PR's TL and the main Storybook's core component Footer. 

Your PR TL: 
image


Footer in Storybook: 
image


If you can see from the screenshots, the Scania font-family doesn't seem to be applied in TL. This happened also when we were migrating the Popover Menu, in each menu item. It's probably falling back to one of the fallback fonts.

PR #1953, opened 2026-06-16
feat(radio-button): demo single disabled option in a group

ckrook:
Adds a Storybook story to both the regular tds-radio-button and tegel-lite tl-radio-button showing a group where only one Radio Button is disabled while the others stay interactive.

### DIFF
--- a/packages/core/src/components/radio-button/radio-button.stories.tsx
+++ b/packages/core/src/components/radio-button/radio-button.stories.tsx
@@ -25,6 +25,17 @@ export default {
         type: 'text',
       },
     },
+    checkedIndex: {
+      name: 'Checked index',
+      description: 'Sets which Radio Button in the group is checked by index, or "none".',
+      control: {
+        type: 'radio',
+      },
+      options: [0, 1, 2, 'none'],
+      table: {
+        defaultValue: { summary: 0 },
+      },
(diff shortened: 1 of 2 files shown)


mfilipadneto (approved):
Looks good!!

packages/core/src/components/radio-button/radio-button.stories.tsx
```diff
@@ -25,42 +25,44 @@ export default {
...
-  <tds-radio-button 
+  <tds-radio-button
     name="rb-example"
     value="option1"
     radio-id="option-1"
     required=false
-    ${disabled ? 'disabled' : ''}
-    checked="true" 
+    ${disabledIndex === 0 || disabledIndex === 'all' ? 'disabled' : ''}
+    checked="true"
```
claraseverino: Could we make the checked state dynamic? So that the selected option persists when we select a different disabled value. I see that this behavior was already here, but it is a bit confusing

claraseverino:
The tegel lite story has a checkedIndex, but tegel does not. Does it make sense to add it to tegel as well? To make it more consistent

NordbergAnna (approved):
LGTM!

claraseverino (approved):

PR #1954, opened 2026-06-16
feat(divider): all six Figma variants + rename dark-blue to expressive

ckrook:
Aligns both tds-divider and tl-divider with the Figma variant set: Discrete, Subtle, Soft, Defined, Strong, Expressive (shown in Storybook and used in code). 'Dark-Blue' is renamed to 'Expressive'; the old `dark-blue` prop value is kept as a silent deprecated alias so existing usage doesn't break. tl-divider gains the four missing variants (subtle, soft, defined, strong) to match the regular component, and its `expressive` now correctly uses the expressive token.


mfilipadneto (approved):
Looks great! Maybe we should leave a note in the release notes when this is released to let the users know the `dark-blue` will be deprecated?

NordbergAnna (approved):

PR #1989, opened 2026-07-28
fix(stencil): disable adding global styles to every component

theJohnnyMe:
## **Describe pull-request**
This PR fixes a memory issue seen in Angular unit tests when using Tegel.

After the Stencil upgrade, Tegel started adding the global stylesheet into every component shadow root. In jsdom, this creates many large `<style>` elements when tests create and remove Tegel components. This causes very high Node.js heap usage and can end with:

`JavaScript heap out of memory`

The proposed fix disables this Stencil behavior by setting:

`addGlobalStyleToComponents: false`

## **Issue Linking:**
- **No issue:** Angular unit tests using Tegel started running out of memory after upgrading from `@scania/tegel` `1.47.2` to `1.48.0` and later versions.

## **How to test**
1. Run the core build:
   `npm run build:core`

2. Check the generated Stencil runtime in `packages/core/dist/esm/index-*.js`.

3. Confirm that the generated runtime no longer includes the large embedded `globalStyles` string.

4. Run a jsdom memory test that creates and removes Tegel custom elements.

Testing done locally:
- `npm run build:core` passed.
- Generated runtime no longer contains `globalStyles`.
- Generated runtime no longer calls `createStyleSheetIfNeededAndSupported(globalStyles)`.
- Local jsdom repro with 83 Tegel tags, 10 iterations, 2 elements per tag:
  - Before fix / `1.48.0`: around `1900 MB` heap usage.
  - With this fix: around `60 MB` heap usage.

- [x] All existing tests pass

## **Suggested test steps**
- [ ] Keyboard operability
- [ ] Storybook controls
- [ ] Events

## **Screenshots**
No UI changes.

## **Additional context**
This issue seems to be related to how jsdom handles styles in shadow roots. Browsers can use constructable stylesheets, but jsdom falls back to normal `<style>` elements. Because Tegel has a large global stylesheet, adding it to every component shadow root causes memory usage to grow very quickly in unit tests.

This change keeps the global stylesheet output, but prevents Stencil from copying it into every component shadow root.


theJohnnyMe:
I spotted that in screenshot comparison of unit test that had to be updated problem is missing styling for checkbox component. If this PR proves to solve the problem for memory heap, I could take a look in dependent components - that use other component not by slotting but inside a shadowDOM and make sure stylings of those components are also consumed by "consuming" component. Lets have this in mind. 
Now best would we to have a beta release made with this branch just to test hypothesis of injecting global styles into all component is source of memory heap issue.

theJohnnyMe:
Tested `1.60.0-mem-issues-beta.0` in `chs-chargingservices-ui`.

I reverted our local unit-test workaround that had been added because Tegel exhausted jsdom/Vitest worker memory. The workaround replaced the real `AddEditDeparturesComponent` template with an empty one via `overrideComponent(... template: '')`.

With the real `AddEditDeparturesComponent` template restored, `@scania/tegel@1.59.1` reproduced the issue: the Vitest worker hit `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory`, and only `13 / 68` tests completed before the worker crashed.

Then I tested:

@scania/tegel: 1.60.0-mem-issues-beta.0
@scania/tegel-angular-17: 1.60.0-mem-issues-beta.0

Results:

Targeted restored-template spec:
PASS - 68 / 68 tests
Peak tree RSS: 1758.1 MB
Peak process RSS: 1185.1 MB

Targeted restored-template spec with NODE_OPTIONS=--max-old-space-size=2048:
PASS - 68 / 68 tests

Full unit suite:
No heap OOM
63 test files ran
62 passed, 1 failed
519 passed, 4 failed
Peak tree RSS: 4915.7 MB
Peak process RSS: 1289.9 MB

Full unit suite with NODE_OPTIONS=--max-old-space-size=2048:
No heap OOM
Same 4 test assertion failures

The remaining 4 failures are in valid-from-offset.component.spec.ts and appear unrelated to Tegel memory/rendering.

Conclusion: this beta fixes the jsdom/Vitest heap issue for our previously failing restored-template repro.

I will find time to fix components affected by this change and that might be table and dropdown that use checkbox component int hem but in shadowDOM. For those we need to provide stylings of checkbox inside of consuming component.

mfilipadneto (approved):

claraseverino (approved):

Issue #1806, opened 2026-03-18
[Feature Request]: Option to hide min/max character indicator on text field and text area


patrikhultgren:
### Requirements before reporting

- [x] No duplicated [issue](https://github.com/scania-digital-design-system/tegel/issues) reported
- [x] I have checked the latest version if the feature exist there

### Type of feature request

Functionality

### What is your feature request?

I would like an option to hide the character counter (e.g. “0 / 200”) below text fields. For simple inputs like “First name,” it doesn’t add much value and can make the UI feel cluttered. It would be useful to control this per field or only show it when needed (for example after typing or near the limit).

### Background

Right now, the character counter is always shown below text fields, no matter the type of input. It can be helpful for longer text, but not for short fields like names or titles. In those cases, it doesn’t add much value and can make the UI feel a bit cluttered. More control over when it’s shown would help keep the interface cleaner and more relevant.

<img width="218" height="115" alt="Image" src="https://github.com/user-attachments/assets/edebcaab-0ab6-4efe-9c91-d4349deafade" />

### Idea

Add an option to show or hide the character counter for each field. It could be hidden for simple inputs, or only shown when needed (for example after typing or near the limit). This would make forms more flexible and help keep the design clean.

### Goals and objectives

The goal is to make forms cleaner and easier to use by removing unnecessary elements. By making the character counter optional, we can reduce visual clutter for simple fields while still showing it where it’s useful. This should make the overall experience more clear and intuitive.

### Contact information

patrik.hultgren@scania.com

antonstrandqvist:
Hi @patrikhultgren !

Do you need to have the maxLenght on those fields at all then? you could just disable it and the counter disappears

patrikhultgren:
I still want to keep maxLength for validation though, it’s native to the input and prevents invalid data. The need is mainly to hide the counter visually, not remove the limit.

So it would be useful to decouple those, and control whether the counter is shown or not.

antonstrandqvist:
Ahh i see. Will create a ticket for this!

mfilipadneto:
Hello @patrikhultgren 🙂 
Better late than ever, right?
I have opened a PR for this Feature Request. If you want, you can check it out here (#2028 ) and let me know if this is what you requested

mfilipadneto:
Hello Patrik 😄 

This has been fixed and released in version 1.64.0 of Tegel! 
Will close this ticket.

PR #2028, opened 2026-09-01
Feat(tds-text-field, tds-textarea): add Prop to control character count indicator visibility

mfilipadneto:
## **Describe pull-request**  
This PR implements a Feature Request (linked below) where the users will be able to control whether to hide the character indicator or not, without losing the validation/limitation of the maximum characters imposed. 
Also added new snapshot tests to verify this new feature.

## **Issue Linking:**  
- **GitHub:** #1806 by @patrikhultgren 

## **How to test**  
1. Go to this PR's Storybook
2. Go to [Text Field](https://pr-2028.d3fazya28914g3.amplifyapp.com/?path=/story/components-text-field--default) / [Textarea](https://pr-2028.d3fazya28914g3.amplifyapp.com/?path=/story/components-textarea--default) Tegel component
3. Define a small maximum length for the field/area just to test (eg. 5)
4. Set control to hide max length indicator as True
5. Try to type more characters than the maximum length defined above
6. Verify that even though the indicator is hidden, it still limits the number of characters written


packages/core/src/components/text-field/text-field.tsx
```diff
@@ -75,6 +75,9 @@ export class TdsTextField {
   /** Max length of input */
   @Prop({ reflect: true }) maxLength?: number;
 
+  /** Lets user control whether to hide max length indicator of input */
+  @Prop({ reflect: true }) hideMaxLength?: boolean = false;
```
joao-goncalves-vwgds: Seems like `hideMaxLength` property doesn't need to be `undefined`, similar to `hideReadOnlyIcon` and `noMinWidth` props.

packages/core/src/components/textarea/textarea.tsx
```diff
@@ -55,6 +55,9 @@ export class TdsTextarea {
   /** Max length of input */
   @Prop({ reflect: true }) maxLength?: number;
 
+  /** Lets user control whether to hide max length indicator of textarea */
+  @Prop({ reflect: true }) hideMaxLength?: boolean = false;
```
joao-goncalves-vwgds: Same for this prop.

joao-goncalves-vwgds (approved):

PR #2060, opened 2026-09-11
chore(pnpm): migrate package manager from npm to pnpm

joao-goncalves-vwgds:
## **Describe pull-request**  

The goal of this PR is to replace the current package manager (`npm`) with `pnpm`, which is a faster and less disk consuming package manager. With this PR, `pnpm` is introduced, and the whole repository is now a `pnpm workspace`, which largely facilitates the development of the monorepo, with packages being automatically linked together locally, without the need of linking or packing them.

What was done:

- GitHub Actions Workflows were updated, so that CI now uses pnpm
    - All workflows were tested with the dry run versions
- Dependabot configuration was updated to only consider the root directory (it will automatically track the workspace packages, based on the workspace configuration file on the root)
- All documentation was updated, to now reference pnpm
- Git hooks were updated to use pnpm, along with all scripts in the `package.json` files
- Angular value accessor directives were fixed, with a `forwardRef` call, due to a TypeScript self referencing issue
- `@angular/forms` was added as a peer dependency to the `@scania/tegel-angular-17`, due to it being used on the distribution folder, on the value accessor directives
- `axe-core` and `glob` packages were explicitly added as dependencies, due to previously being used on our source files, but only relying on them being transitive dependencies
- `@playwright/test` was added to the default pnpm catalog, to demonstrate the value of having packages whose version should be shared across all packages, in a single location
- Added a Dockerfile to run the Playwright tests for both core and tegel-lite, instead of using the previous docker-runner script
- Added a script to create an exportable manifest, which basically turns a `package.json` that has `workspace:` and `catalog:` notations, with their actual versions (this is needed because the `package.json` file on the `dist/components` is outside of the workspace, and publishing does not work there, out of the box, yet)
- Added `pnpm` and the corresponding version as a `packageManager` entry in the root `package.json`
- Create a `pnpm-workspace.yaml`, with the configuration of all the packages, and the default catalog
- Improved the `Setting up the development environment` section on the root `README.md`, with steps to ensure a proper environment with `pnpm`

## **Issue Linking:**  

- **No issue:** package manager replaced, from npm to pnpm.  

## **How to test**  

1. Go to these pages and check that everything works, as before:
    - [Storybook](TBD)
    - [Tegel React Demo](https://pr-345.d2vh2lmnzgzrkl.amplifyapp.com)
    - [Tegel Angular Demo](https://pr-140.d24qjtb7na4hmr.amplifyapp.com)
2. Checkout to this branch and:
    - Follow the steps in the `Setting up the development environment` section, on the root `README.md`, and ensure there are no local issues (if your workspace still has `node_modules` that referred to an `npm install`, perhaps it can be easier to run `pnpm reinstall`)
    - Run the `pnpm test` and `pnpm test:lte` scripts, and ensure all tests work, same as before
    - Run `pnpm start`, and ensure the server starts and storybook shows everything, same as before

## **Additional context**  

This PR is another step towards improving the repository, and the developer experience.


joao-goncalves-vwgds:
With the usage of `pnpm`, the build settings for Amplify need to be updated. Since updating it might break the other branches not using `pnpm` yet, right before merging this PR, let's update `amplify.yml`, to ensure Amplify works, and Storybook is correctly deployed:

```yml
version: 1
frontend:
  phases:
    preBuild:
      commands:
        - corepack enable
        - pnpm config set store-dir .pnpm-store
        - if [ "${AWS_BRANCH}" != "main" ]; then export VITE_STORYBOOK_ENV=dev; fi
        - pnpm install --frozen-lockfile
    build:
      commands:
        - cd packages/core
        - if [ "${AWS_BRANCH}" = "main" ]; then pnpm add @scania/tegel-lite@latest; fi
        - pnpm build-stencil:storybook
  artifacts:
    baseDirectory: packages/core/storybook-static
    files:
      - '**/*'
  cache:
    paths:
      - node_modules/**/*
      - .pnpm-store/**/*
```

mfilipadneto (approved):

crseverino (approved):

